PR #34, opened 2019-10-18
Add the option of specifiying the use of threading, OpenCL or any custom compile option

rok-cesnovar:
This PR adds the option to compile the models with: 
- stan threads flag
- OpenCL support, you can also specify the platform and device ids
- any custom compilation flag

It also adds two functions for setting and reading the STAN_NUM_THREADS. For now these are named set_num_threads() and get_num_threads(), though those are probably too generic names?

It also adds a convenience for Cmdstan by recompiling main.o and the model executable, .hpp and .o files if you change the compile flags. This is a convenience because cmdstan would not recompile the model if make/local was changed.  It also circumvents [an issue](https://github.com/stan-dev/cmdstan/issues/757) of Cmdstan 2.20+ where you need to recompile main.o manually if you change specific compilation flags. In cmdstanr we would delete main.o if the contents of make/local were modified. If there were no changes, this compiles normally as before.

I added test for the get/set_num_threads. Not sure what tests can I write for the compilation stuff.


jgabry:
Great, thanks for this! I’ll review both of the PRs this weekend or early next week.

rok-cesnovar:
No worries & no rush. Get to them when you can. I know you are a pretty busy.

R/utils.R
```diff
@@ -193,3 +193,84 @@ write_rdump <- function(data) {
...
+  old_make_local_content <- ""
+  if(file.exists(make_local_path)) {
+    # read the contents of make/local to compare with the new contents
+    old_make_local_content <- paste(readLines(make_local_path), collapse = "\n")
+  }
+  if(opencl) {
+    stan_opencl <- "STAN_OPENCL = true"
+    platform_id <- paste("OPENCL_PLATFORM_ID", opencl_platform_id, sep = " = ")
+    device_id <- paste("OPENCL_DEVICE_ID", opencl_device_id, sep = " = ")
+    compiler_flags <- c(compiler_flags, stan_opencl, platform_id, device)
```
jgabry: Does this work for you? I think the object `device` that is included as the last element of `compiler_flags` is not defined  (just `device_id`), so I would think this errors. Or am I missing something?

R/utils.R
```diff
@@ -193,3 +193,84 @@ write_rdump <- function(data) {
...
+#'
+#' @param num_threads (non-zero positive integer) the number of threads to set
+#' @noRd
+set_num_threads <- function(num_threads) {
+  if(is.numeric(num_threads) && num_threads%%1==0 && num_threads > 0) {
+    Sys.setenv("STAN_NUM_THREADS" = num_threads)
+  } else {
+    stop("Please set a valid number of threads. Valid values are integers > 0.", call. = FALSE)
+  }
+}
```
jgabry: Are these functions intended to be exposed to the user or for internal use? Currently they aren't getting exposed but I also don't them getting used internally anywhere.

R/model.R
```diff
@@ -140,6 +139,20 @@ CmdStanModel <- R6::R6Class(
...
 #'
+#' @template model-common-args
+#' @section Arguments unique to the `compile` method:
+#'   These arguments are described briefly here and in greater detail in the
+#'   CmdStan manual. Arguments left at `NULL` default to the default used by the
+#'   installed version of CmdStan.
+#'   * `opencl`: (logical) Should the model be compiled with OpenCL support enabled?
+#'   * `opencl_platform_id`: (positive integer) The ID of the OpenCL platform on which
+#'     to run the compiled model.
+#'   * `opencl_device_id`: (positive integer) The ID of the OpenCL device on the selected
```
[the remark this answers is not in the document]
rok-cesnovar: Its " >= 0", I will change to non-negative integers. Would that work? 0 is the default and works.

R/utils.R
rok-cesnovar: Yeah, I switched it up at the last second. Will fix.

R/model.R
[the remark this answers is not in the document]
jgabry: Yeah that's fine!

R/utils.R
rok-cesnovar: I wanted to expose them but failed :/ Learning all the ins and outs of making a proper R package on the job here. Sorry about that.

These should be function on the same level as set_cmdstan_path() etc. 

Something like this:

```
# compile
mod$compile(threads = TRUE)

set_num_threads(6)
# try sampling
standata <- list(N = 10, y =c(0,1,0,0,0,0,0,0,0,1))
fit <- mod$sample(data = standata)
```

That is why I am not sure if we should add any prefix to the name or are R users accustomed to these cases and then use cmdstanr::set_num_threads()?

R/utils.R
```diff
@@ -193,3 +193,84 @@ write_rdump <- function(data) {
...
+    file.remove(model_hpp_file)
+  }
+  if(file.exists(model_o_file)) {
+    file.remove(model_o_file)
+  }
+  if(file.exists(model_path)) {
+    file.remove(model_path)
+  }
+  if(remove_main) {
+    main_o_file <- paste(cmdstan_path(), "src", "cmdstan", "main.o", sep = "/")
```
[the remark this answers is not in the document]
rok-cesnovar: I wasnt aware of the standard use :) Not really fluent in "textbook R", up to now I used it more or less for personal use so as long as it worked for me I was fine with it :)

R/utils.R
rok-cesnovar: Added @export to both. The names are up for debate.

jgabry: Ok, I’ll think about the names. In terms of exporting these functions, can you either give me push access to your fork or put this on a branch in the main repo? Then I can make a quick commit showing how to export these properly and you can see that and use the technique going forward? What do you think?

R/utils.R
rok-cesnovar: Fixed.

R/model.R
```diff
@@ -140,6 +139,20 @@ CmdStanModel <- R6::R6Class(
...
+#'     OpenCL platform on which to run the compiled model.
+#'   * `compiler_flags`: (vector of strings) Vector of strings representing custom compiler
```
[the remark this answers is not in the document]
rok-cesnovar: Fixed.

R/utils.R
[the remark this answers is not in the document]
rok-cesnovar: Fixed both uses of paste where file.path could be used.

R/model.R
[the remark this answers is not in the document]
rok-cesnovar: Fixed.

R/model.R
```diff
@@ -140,6 +139,20 @@ CmdStanModel <- R6::R6Class(
 #'   $compile()
 #'   ```
 #'
+#' @template model-common-args
```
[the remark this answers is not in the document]
rok-cesnovar: Thank you for the explanation! Removed.

R/utils.R
jgabry: Didn’t see your comment before posting my previous reply. @export is good. In order to put the two functions in the same doc page you also need to use @rdname when documenting the second function. For example to document `set_cmdstan_path` together with `cmdstan_path` I did this: 

https://github.com/stan-dev/cmdstanr/blob/b7916768ab81d4fe5101ee081e666f12ff9fe691/R/path.R#L43

rok-cesnovar: I added export and ran build source package. Do I need to do anything else? 

You should have push access to the branch as the "Allow edits from maintainers" is ticked on the PR. Also feel free to push to any of my PRs at any time. The next one will be on the main repo so that should make things easier. I like learning by example :) Thank you!

rok-cesnovar: Yeah, Github doesnt refresh replies. will add `@rdname`.

jgabry: Great. Just `@rdname` if you haven’t already (see previous comment).

jgabry: Haha yeah I’m having that issue too!

rok-cesnovar:
@jgabry added rdname so I think these are now all review comments addressed. Do you have any good / quick-to-the-point resources for roxygen so I get a better overview of these @ keywords so I am not as lost :)

jgabry:
They have some vignettes for roxygen here 

https://roxygen2.r-lib.org/articles/

One particular thing that’s a bit confusing about using roxygen doc for cmdstanr is that we can only use `@param` for regular functions and not for methods associated with the R6 classes. That’s why I resorted to itemized lists for the arguments to methods like $sample(), but use `@param` to document arguments to regular functions like cmdstan_path(). I think eventually roxygen will have a better solution for R6 methods as I’ve seen a lot of people requesting it.

jgabry:
@rok-cesnovar For some reason it's only letting me push to your branch if I do it online directly from the github website and not if I clone and modify locally and then push (it says authentication failed). I did push a few changes but I couldn't do everything I wanted because of that limitation. Can you pull and then regenerate the documentation and namespace by running `devtools::document()` or `roxygen2::roxygenize()` and then push? That will update the manual pages from the roxygen tags and update the namespace file and then I think it should pass on travis.

jgabry:
Also, one more question: are we sure that CmdStan is can the STAN_NUM_THREADS environment variable? I'm not saying it doesn't, but were you able to verify that? I don't know if setting it in the R session is enough or if we need to set it directly via a system call.

rok-cesnovar:
Thanks for cleaning up the docs. Thats where I lack the knowledge when it comes to R. The tests pass now. Will do the next PRs on the main repo to avoid all these permission issues. I also gave you push permission on the fork on the off chance this issue comes up again.

> are we sure that CmdStan can find the STAN_NUM_THREADS environment variable?

I had the exact same doubts and tested this on linux when I implemented it. I had to add a print statement to the initalize threads function in Cmdstan (Cmdstan doesnt give any info on how many threads are running, which it probably should, will make an issue but that doesnt help us here). It works a expected. 

Also the Sys.setenv doc states `Sys.setenv sets environment variables (for other processes called from within R or future calls to Sys.getenv from this R process).` So it should work normally. Will double check on Windows since Windows is the most tricky in terms of environment variables and processes etc. Other than that I think we are good.

Cant comment on the commit but thank you very much for adding me as an author. Very generous of you! Will certainly try my best to justify your trust in me. You can add my e-mail if you think it helps in any way. Not sure you have it or if its seen on github (name.lastname at fri.uni-lj.si).

Issue #42, opened 2019-10-24
don't use symlinks on windows? 


jgabry:
Currently `install_cmdstan` installs the latest CmdStan and creates a symlink for it so it can just be referred to by `cmdstan` instead of the including the particular version number. This is convenient, but I've heard symlinks could be problematic on Windows. @rok-cesnovar do you know anything about this?

rok-cesnovar:
Havent tried installing cmdstan via install_cmdstan() on Windows yet, but in general these simple symlinks should work. 

The symlinks that dont usually work is when you are tryign to dynamically link a .dll file via symlink. Will investigate. Assigning myself.

jgabry:
Thanks!

rok-cesnovar:
ln -s ${CS} essentially does a copy on Windows which I guess is fine. I will open a PR to cleanup some of the minor Windows issues. I am also wondering why is there a need for the symlink at all.

jgabry:
Thanks, and yeah i think you’re probably right to question using symlinks at all. if you want and have time feel free to remove the symlink usage in the package entirely.

rok-cesnovar:
We can close this one as we no longer use symlinks.

PR #44, opened 2019-10-24
install_cmdstan windows fixes & cleanup, set TBB path

### LINKED ISSUE #37: Handle TBB lib on Windows
Cmdstan 2.21 now includes the Intel TBB library. This changes nothing for non-Windows users in terms of how they build models.

Windows users of cmdstan have the option of adding the path to the build library (<cmdstan-path>/stan/lib/stan_math/lib/tbb) to their user of system PATH variable. If they dont do that, the tbb.dll is copied to the executable folder for each model. 

So techincally they also dont have to do anything, but to avoid copying of dlls or requiring the users to set the path variable cmdstanpy and cmdstanr could provide convenience features that Windows users dont have to add that. 

The easiest solution would be to add the above path to the PATH variable for the session or simply for every run.

I added a windows label, hope that is ok with you @jgabry.

rok-cesnovar:
Fixes #37 and #42 

This PR does the following: 
- on windows calls make_cmdstan.sh with `-w` so we can handle window specific stuff
- uses mingw32-make on Windows to build (mandatory in 2.21 due to TBB, its also fine for all previous versions, comes with RTools just like make)
- removes the need for smylinking of cmdstan on install; just extracts to cmdstan/ (was there a reason this wasnt done this way? I might be missing something);
- adds removal of the tar.gz file to make_cmdstan.sh
- on Windows with 2.21+ adds path to TBB lib to the environment variable PATH so we dont need to copy tbb.dll every time in the model exe folders

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -28,6 +28,7 @@ install_cmdstan <- function(dir = NULL, cores = 2, quiet = FALSE) {
   if (os_is_windows()) {
     make_cmdstan <- c(make_cmdstan, "-w")
   }
+
   install_log <- processx::run(
     command = "bash",
     args = make_cmdstan,
--- a/R/install.R
+++ b/R/install.R
@@ -25,6 +25,9 @@ install_cmdstan <- function(dir = NULL, cores = 2, quiet = FALSE) {
     make_cmdstan <- c(make_cmdstan, paste0("-d ", dir))
   }
   make_cmdstan <- c(make_cmdstan, paste0("-j", cores))
+  if (os_is_windows()) {
+    make_cmdstan <- c(make_cmdstan, "-w")
+  }
   install_log <- processx::run(
     command = "bash",
     args = make_cmdstan,
--- a/R/install.R
+++ b/R/install.R
@@ -25,7 +25,7 @@ install_cmdstan <- function(dir = NULL, cores = 2, quiet = FALSE) {
     make_cmdstan <- c(make_cmdstan, paste0("-d ", dir))
   }
   make_cmdstan <- c(make_cmdstan, paste0("-j", cores))
-  if ((.cmdstanr$VERSION >= "2.21") && os_is_windows()) {
+  if (os_is_windows()) {
     make_cmdstan <- c(make_cmdstan, "-w")
   }
   install_log <- processx::run(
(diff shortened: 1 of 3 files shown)


R/install.R
```diff
@@ -25,6 +25,9 @@ install_cmdstan <- function(dir = NULL, cores = 2, quiet = FALSE) {
     make_cmdstan <- c(make_cmdstan, paste0("-d ", dir))
   }
   make_cmdstan <- c(make_cmdstan, paste0("-j", cores))
+  if ((.cmdstanr$VERSION >= "2.21") && os_is_windows()) {
```
jgabry: I think (for now at least) we can just assume this function always installs the latest cmdstan release so just checking for windows is fine. Anyway, checking the version number would only tell us the version number they _already_ have installed (if any), right?

R/model.R
```diff
@@ -192,6 +192,11 @@ compile_method <- function(threads = FALSE,
     build_cleanup(exe,
                   remove_main = TRUE)
   }
+  # add path to the build tbb library to the PATH variable to avoid copying the dll file
+  if (os_is_windows()) {
+    path_to_TBB <- file.path(cmdstan_path(), "stan", "lib", "stan_math", "lib", "tbb")
+    Sys.setenv(PATH = paste0(path_to_TBB, ";", Sys.getenv("PATH"),sep=""))
```
jgabry: This looks right, but just checking, were you able to confirm on windows that this works?

inst/make_cmdstan.sh
```diff
@@ -4,14 +4,17 @@
 #  - build binaries, compile example model to build model header
 #  - symlink downloaded version as "cmdstan"
 
-while getopts ":d:v:j:" opt; do
+WIN=0
+while getopts ":d:v:j:w" opt; do
```
jgabry: The changes to this script look good. I assume it does, but if you haven’t already can you check that it still works when _not_ on windows?

jgabry: Actually never mind, don’t worry about that. Travis will do that for us.

R/install.R
rok-cesnovar: Oh right, that is correct.

R/model.R
rok-cesnovar: Yes, I double checked that the .dll file is not copied.

rok-cesnovar: This is where I wanted to put that version check in with the last commit. This doesnt need to be done for 2.20 and earlier. It doesnt really hurt, but still unnecessary. Fixed now.

inst/make_cmdstan.sh
rok-cesnovar: I was counting on Travis yes, though I did go in and check now and its all good.

PR #46, opened 2019-10-25
Write pretty json

jgabry:
This PR add the argument `pretty=TRUE` internally when `write_stan_json()` calls `jsonline::write_json()`. The `pretty` argument gets passed to `jsonlite::toJSON()` and results in writing a JSON file that's easier for a human to read. 

@rok-cesnovar Presumably there shouldn't be any speed difference for CmdStan when reading it in, right? 

I had to fix the tests after changing this so I also went ahead and changed to using `expect_known_output` for the JSON tests so we can just compare to JSON files that are now stored in `tests/testthat/answers`. 

Also, it occurred to me that since these tests don't actually require a CmdStan installation we can let them run on CRAN (when we eventually release) so I removed the `skip_on_cran()` usage from these tests.


rok-cesnovar:
Also thanks for doing this. Didnt occur to me about pretty... 

The known output is also much better than my approach.

jgabry:
It didn’t occur to me to me before either. I didn’t even notice the pretty argument before

jgabry:
Thanks for taking a look so quickly.

PR #48, opened 2019-10-26
Automatically compile

### LINKED ISSUE #45: Should cmdstan_model default to automatically compiling?
There doesn't really seem to be a use case for a CmdStanModel object before compilation, so maybe it should just happen by default? E.g. `cmdstan_model()` could have a `compile` argument defaulting to `TRUE`.  

We can decide this in conjunction with @mitzimorris and other CmdStanPy devs. Corresponding isssue is https://github.com/stan-dev/cmdstanpy/issues/153

jgabry:
Closes #45

This PR makes it so compilation happens by default when calling `cmdstan_model()`, which now has signature: 

```
cmdstan_model(stan_file, compile = TRUE, ...)
```

where `...` can be used to pass optional arguments to the `$compile()` method (e.g. `opencl=TRUE`).


jgabry:
Thanks! Just fixed the conflicts (that was my fault). I'll wait for travis to run and then merge if there are no issues.

PR #49, opened 2019-10-26
Dont overwrite users make local flags

### LINKED ISSUE #47: Not sure if overwriting make/local is the correct behavior
My custom make/local settings keep getting overwritten by cmdstanr when I go to compile. I"m not sure that's good behavior.

rok-cesnovar:
Fixes #47 
This issue was raised by @bbbales2, thank you for that. Sorry for any inconvenience.

The way I envisioned things before was that the users would set all of their compile time flags through cmdstanr which now that I think of it was probably not the brightest idea. It would cause problems for users coming from cmdstan because it would demand they copy their flags to R.

With this PR users can still handle all their compile time flags directly in make/local. cmdstanr will append any additional flags after the directly set flags. The convenience cmdstanr features (force recompiling of model.o and main.o on changes to the flags) will only be available for flags set through cmdstanr.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -195,7 +195,7 @@ compile_method <- function(threads = FALSE,
   # add path to the build tbb library to the PATH variable to avoid copying the dll file
   if ((.cmdstanr$VERSION >= "2.21") && os_is_windows()) {
     path_to_TBB <- file.path(cmdstan_path(), "stan", "lib", "stan_math", "lib", "tbb")
-    Sys.setenv(PATH = paste0(path_to_TBB, ";", Sys.getenv("PATH"),sep=""))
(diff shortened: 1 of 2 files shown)


rok-cesnovar:
En example of a make/local with both directly set and cmdstanr flag

```
CXXFLAGS += -DDIRECT_FLAG
CXXFLAGS += -DDIRECT_FLAG2
# cmdstanr generated make/local flags (add user flags above this line)
CXXFLAGS += -DCMDSTANR_TEST1
CXXFLAGS += -DCMDSTANR_TEST234
```

rok-cesnovar:
Thank you for cleaning up my R code. Hopefully I will learn :)

PR #51, opened 2019-10-28
Suppress CmdStan compilation output by default

### LINKED ISSUE #50: Turn off compilation output by default?
Similar to what already exists for `install_cmdstan()`, I think we should have an argument `quiet` for the `$compile()` method, but unlike for `install_cmdstan()` I think for `$compile()` it should default to `TRUE.`

jgabry:
Closes #50

This PR adds an argument `quiet` (defaulting to `TRUE`) to the `$compile()` method. This means that the standard output and error from CmdStan is suppressed during compilation. The downside to this is that any errors are just generic errors (instead of the content of stderr), so I put in the doc a suggestion to set `quiet=FALSE` if the user encounters an error during compilation and wants to see more of the output. Another option I guess would be to store the stderr output in case there's an error and then return that to the user if necessary?

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -115,6 +115,7 @@ CmdStanModel <- R6::R6Class(
       checkmate::assert_flag(compile)
       private$stan_file_ <- absolute_path(stan_file)
       if (compile) {
+        message("Compiling Stan program...")
         self$compile(...)
       }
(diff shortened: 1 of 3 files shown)


jgabry:
Sounds good, thanks!

PR #52, opened 2019-10-29
Optionally save and access diagnostic files

### LINKED ISSUE #39: Allow specification of cmdstan diagonostic_file and parse the outputs
This file has the unconstrained parameters, momentums, and gradients

jgabry:
Closes #39 

This PR adds an argument `save_diagnostics` (defaulting to `FALSE`) to the fitting methods instead of making the user specify the argument `diagnostic_file` as a path. If `save_diagnostics=TRUE` then CmdStanR will create the diagnostic csv files and specify the `diagnostic_file` argument for each chain when calling CmdStan. 

Just like the output csv files and the data file there are methods now provided for getting the path to the diagnostic csv tempfiles (`fit$diagnostic_files()`) and for saving the tempfiles to a different directory (`fit$save_diagnostic_files()`). 

For now it is up to the user what to do with the csv files, but we may add methods for working with the diagnostic output at some point in the future.


jgabry:
Thanks @rok-cesnovar. And yeah I'll wait to see if @bbbales2 is happy with this before merging.

bbbales2:
Here are some observations/comments. I'm happy with this as is. I think it'd be good if number 2 got addressed though.

1. It looks like diagnostic_file for optimization doesn't contain anything other than the comments. I guess it's still fine to take it as an argument cause that is the cmdstan behavior.

2. I think add the error "Error: No diagnostic files found. Set 'save_diagnostics=TRUE' when fitting the model." when someone calls ```diagnostic_files()``` without doing save_diagnostics. (it's already there for save_diagnostic_files apparently -- that's where this error came from).

3. It looks like ?fit-method-diagnostic_files and ?fit-method-output_files don't exist. Is this intentional?

jgabry:
Thanks @bbbales2! And yeah all of those are accurate observations. Some responses to your numbered points: 

(1) yeah this is useless right now but is valid for cmdstan and I guess could someday include something? Now I’m thinking that we should disabling it for optimization until it becomes useful. If it’s allowed but doesn’t include anything but comments users could mistake that for a bug or a mistake on their part. 

(2) Good idea. I should have done that and will do. 

(3) Not intentional, those should point somewhere, thanks! Will fix

jgabry:
Alternatively for (1) I suppose we could leave it working (so we don’t need to remember to enable it later) but clearly document that it just returns whatever cmdstan writes, which as of 2.21 is just comments for optimization.

jgabry:
Ok I took care of what @bbbales2 mentioned (and also touched up the doc in a bunch of other places along the way). 
Will let travis finish before merging.

Issue #63, opened 2019-11-04
Passing metric as a file or R object 


bbbales2:
In Cmdstan the metric argument determines the type of metric and then the metric_file argument provides the path.

Currently in cmdstanr metric can take a file path, but it doesn't then break the arguments out correctly to Cmdstan (it just sets cmdstan metric to a path and that gives an error).

I think we either break this out into the two arguments cmdstan uses (metric and metric_file) or something that makes sense from R (metric and metric_value), where metric_value is a vector/matrix that cmdstanr then writes to a file.

I'm gonna try to do the metric, metric_file thing.

jgabry:
Thanks! I think it would be convenient for users if we also allow metric to be an R object, eg numeric vector. cmdstanr can write it to json and pass it to metric_file. What do you think?

bbbales2:
> cmdstanr can write it to json and pass it to metric_file. What do you think?

It would be more convenient. If you want to implement this, feel free to take over. Otherwise I'm just gonna do metric_file for now.

I think specifying metric_value is better than metric_file totally.

bbbales2:
No wait. I don't know why I wouldn't just do metric_value right off the bat. It seems just as easy.

All I need to do is add some arguments to the model.R and args.R files.

My question is where would I create the temporary metric files and save references to them?

RunSet seems like the natural place cause that's where the diagnostic and output files are saved, but that is shared across MCMC and variational inference and optimization. This is specific to MCMC.

I guess I would do this in args?

jgabry:
Thanks! Yeah I guess the file could just be created immediately when the SampleArgs object is `initialize`d. And I think you can just use the `write_stan_json` function to write the json file.

jgabry:
Also, I guess we should also allow `metric_file` since it's in CmdStan. Is that what you were thinking? Or do you think we should _only_ allow `metric_value` (in addition to `metric`)?

bbbales2:
I think only metric_value. Or I guess the natural argument would be inv_metric since that's what you'd regularly specify in metric_file.

I think we skip the metric_file option. In the same way that in cmdstanr we pass in data as lists (instead of filenames), it makes sense to encapsulate this.

bbbales2:
And by encapsulate this I mean actually hide this interface to cmdstan.

jgabry:
I'm fine with that (although technically we still allow passing data as a file)

jgabry:
[...]

PR #64, opened 2019-11-05
Make it possible to provide metric as vector/matrix (also fix metric_file)

bbbales2:
This pull request picks up from over here: https://github.com/stan-dev/cmdstanr/issues/63#issuecomment-550053480


bbbales2:
I could use some advice on the test framework here. I think I was doing it wrong.

I did my testing outside the test framework which is annoying cause then I write all tests twice.

R/args.R
```diff
@@ -560,68 +587,70 @@ maybe_generate_seed <- function(seed, num_runs) {
...
+  }
+
+  checkmate::assert_character(metric_file, any.missing = FALSE, min.len = 1)
+
+  if(is.character(metric_file)) {
+    files_exist = sapply(metric_file, function(file) checkmate::test_file_exists(file, access = "r"))
+    # TODO: need to check if in the right format (see CmdStanPy implementation)
+    if (any(!files_exist)) {
+      stop(paste0("Could not read metric file(s): ", paste(metric_file[!files_exist], collapse = ", ")), call. = FALSE)
+    }
```
jgabry: If you `assert_character(metric_file)` then you don't need to check `if(is.character(metric_file))`. 

Edit: But actually, what about just replacing all of the highlighted lines above with just the single line below? 

```r
checkmate::assert_file_exists(metric_file, access = "r") # vectorized
```

R/args.R
```diff
@@ -147,15 +149,38 @@ SampleArgs <- R6::R6Class(
       self$thin <- thin
       self$max_depth <- max_depth
       self$metric <- metric
-      # self$metric_file <- character()
+      self$inv_metric <- inv_metric
+      if(!is.null(inv_metric)) {
+        if(!is.null(metric_file)) {
+          stop("Only one of inv_metric and metric_file can be non-NULL")
```
jgabry:
```r
stop("Only one of inv_metric and metric_file can be specified.", call. = FALSE)
```
Are you ok with that?

R/args.R
```diff
@@ -147,15 +149,38 @@ SampleArgs <- R6::R6Class(
...
+        }
+
+        # write all inv_metrics to disk
+        inv_metric_paths <- c()
+        for(metric_i in 1:length(inv_metric)) {
+          inv_metric_path <-
+            file.path(
+              cmdstan_tempdir(),
+              paste0("inv_metric-", metric_i, ".json")
+            )
```
jgabry: I just fixed a bug (that was my fault) on master in https://github.com/stan-dev/cmdstanr/commit/c2badbe8d18ddf3278dd0ac20b956e3b699b3c88 that's related to this. We need to call `tempfile` in situations like this to get some randomness in the name to avoid overwriting in the case that someone uses the same model object with different (inv_)metrics to create separate fit objects. The corresponding fix here would be: 

```r
inv_metric_path <-
  tempfile(
   pattern = paste0("inv_metric-", metric_i, "-"),
   tmpdir = cmdstan_tempdir(),
   fileext = ".json"
  )
```

R/args.R
```diff
@@ -147,15 +149,38 @@ SampleArgs <- R6::R6Class(
...
+          write_stan_json(list(inv_metric = inv_metric[[metric_i]]), inv_metric_path)
+          inv_metric_paths <- c(inv_metric_paths, inv_metric_path)
+        }
```
jgabry: This is nitpicking, but I think it's a bit cleaner if you create all the files together and then use a shorter loop instead of accumulating up the vector of paths. For example:

```r
# also incorporating the addition of `tempfile` mentioned above
inv_metric_paths <-
  tempfile(
    pattern = paste0("inv_metric-", 1:length(inv_metric), "-"),
    tmpdir = cmdstan_tempdir(),
    fileext = ".json"
  )
for (i in seq_along(inv_metric_paths)) {
  write_stan_json(list(inv_metric = inv_metric[[i]]), inv_metric_paths[i])
}
```

It's nitpicky so if you disagree it's fine to leave as is!

R/args.R
```diff
@@ -147,15 +149,38 @@ SampleArgs <- R6::R6Class(
...
+
+        self$metric_file <- inv_metric_paths
+      } else {
+        self$metric_file <- sapply(metric_file, repair_path)
+        self$metric_file <- sapply(self$metric_file, absolute_path)
```
jgabry: I think `absolute_path` calls `repair_path` internally, so you just need `absolute_path`. Di you do this because I use them both back to back in this order somewhere else? If so maybe I did that for a reason but I'd have to take a look.

R/args.R
```diff
@@ -560,68 +587,70 @@ maybe_generate_seed <- function(seed, num_runs) {
...
+    files_exist = sapply(metric_file, function(file) checkmate::test_file_exists(file, access = "r"))
+    # TODO: need to check if in the right format (see CmdStanPy implementation)
+    if (any(!files_exist)) {
+      stop(paste0("Could not read metric file(s): ", paste(metric_file[!files_exist], collapse = ", ")), call. = FALSE)
+    }
+
+    if(length(metric_file) != 1 && length(metric_file) != num_runs) {
+      stop(paste0(length(metric_file), " metric(s) provided. Must provide ",
+                  if(num_runs > 1) "1 or " else "", num_runs, " metric(s) for ",
+                  num_runs, " chain(s)"))
```
jgabry: `stop()` lets you concatenate without needing to also call paste, e.g.: 

```r
stop(
  length(metric_file), " metric(s) provided but must provide ",
  if (num_runs > 1) "either 1 or ",  # you also don't need the else ""
  num_runs, " metric(s) for ", num_runs, " chain(s).", 
  call. = FALSE
)
```

R/model.R
```diff
@@ -263,6 +263,8 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
 #'     thin = NULL,
 #'     max_depth = NULL,
 #'     metric = NULL,
+#'     inv_metric = NULL,
+#'     metric_file = NULL,
```
jgabry: is the order of these three arguments intentional? If not then I guess I slightly prefer keeping the two actual CmdStan arguments together in the signature, so that would mean either `inv_metric` before or after the other two. Or is the idea to encourage using `inv_metric` instead of `metric_file`? (that could be a good idea i guess)

R/model.R
```diff
@@ -304,13 +306,22 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
...
+#'   * `inv_metric`: (vector, matrix) A vector (if metric == 'diag_e') or a
+#'     matrix (if metric == 'dense_e') for initializing the inverse metric. The
+#'     vector is interpreted as a diagonal metric. The inverse metric is
+#'     usually set to an estimate of the posterior covariance. If
 #'     `adapt_engaged=TRUE`, Stan will use the provided metric just as an
-#'     initial guess during adaptation. To turn off adaptation when using a
-#'     precomputed metric set `adapt_engaged=FALSE`.
+#'     initial guess during adaptation. To turn off adaptation when using
+#'     a precomputed metric set `adapt_engaged=FALSE`. Cannot be used in
+#'     conjunction with metric_file.
```
jgabry: Can you add the single back ticks "`" for markdown code formatting around "metric_file"? That way it gets rendered properly in the help pages.

R/model.R
```diff
@@ -304,13 +306,22 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
...
+#'   * `metric_file`: (character) A character vector containing paths to files
+#'     (one per chain) compatible with CmdStan that contain precomputed metrics.
+#'     Each path must be to a JSON or Rdump file that contains an entry
+#'     `inv_metric` whose value is either the diagonal vector or the full
+#'     covariance matrix. If `adapt_engaged=TRUE`, Stan will use the provided
+#'     metric just as an initial guess during adaptation. To turn off adaptation
+#'     when using a precomputed metric set `adapt_engaged=FALSE`. Cannot be
+#'     used in conjunction with inv_metric.
```
jgabry: Can you also add the single back ticks for markdown code around "inv_metric" here?

R/model.R
```diff
@@ -336,6 +347,8 @@ sample_method <- function(data = NULL,
                           adapt_engaged = TRUE,
                           adapt_delta = NULL,
                           metric = NULL,
+                          inv_metric = NULL,
+                          metric_file = NULL,
```
jgabry: Same question as above about the argument order, but can leave as is if we want to promote using `inv_metric` over `metric_file`

R/args.R
jgabry: If you `assert_character(metric_file)` then you don't need to check `if(is.character(metric_file))`. 

Edit: But actually, what about just replacing all of the highlighted lines above with just the single line below? 

```r
checkmate::assert_file_exists(metric_file, access = "r") # vectorized
```

R/args.R
jgabry: If you `assert_character(metric_file)` then you don't need to check `if(is.character(metric_file))`. You could also just use

```r
checkmate::assert_file_exists(metric_file, access = "r") # vectorized
```

to replace all of the highlighted lines above. Or do you strongly prefer this error message? 


Also no need to call `paste` inside `stop`, e.g.: 

```r
stop(
  length(metric_file), " metric(s) provided but must provide ",
  if (num_runs > 1) "either 1 or ",  # you also don't need the else ""
  num_runs, " metric(s) for ", num_runs, " chain(s).", 
  call. = FALSE
)
```

R/args.R
jgabry: No need to call `paste` inside `stop`, e.g.: 

```r
stop(
  length(metric_file), " metric(s) provided but must provide ",
  if (num_runs > 1) "either 1 or ",  # you also don't need the else ""
  num_runs, " metric(s) for ", num_runs, " chain(s).", 
  call. = FALSE
)
```

R/args.R
jgabry: If you `assert_character(metric_file)` then you don't need to check `if(is.character(metric_file))`. You could also just use

```r
checkmate::assert_file_exists(metric_file, access = "r") # vectorized
```

to replace all of the highlighted lines above. Or do you strongly prefer this particular error message?

R/args.R
```diff
@@ -560,68 +587,70 @@ maybe_generate_seed <- function(seed, num_runs) {
...
+#' @param metric_file User's `metric_file` argument.
+#' @param num_runs Number of CmdStan runs (number of MCMC chains).
+#' @return Either throws an error or returns `invisible(TRUE)`.
+#'
+validate_metric_file <- function(metric_file, num_runs) {
+  if (is.null(metric_file)) {
+    return(invisible(TRUE))
+  }
+
+  checkmate::assert_character(metric_file, any.missing = FALSE, min.len = 1)
```
jgabry: I think `min.len=1` will cause this to fail every time `metric_file` isn't specified! (`NULL is length 0)

jgabry: I think this is one reason why tests are failing

tests/testthat/test-model.R
```diff
@@ -183,6 +183,68 @@ test_that("sample() method works with init file", {
...
+
+  inv_metric_vector <- array(1, dim = c(1))
+  inv_metric_vector_json <- test_path("resources", "data", "bernoulli.inv_metric.diag_e.json")
+
+  expect_error(mod$sample(data = data_file_r,
+                          num_chains = 1,
+                          metric = "diag_e",
+                          inv_metric = inv_metric_vector,
+                          metric_file = inv_metric_vector_json))
+})
```
jgabry: Can you add the second argument to `expect_error`, which is a regular expression for matching to the error message. So here you could put part of the expected error message like `"Only one of inv_metric and metric_file"`

tests/testthat/test-model.R
```diff
@@ -183,6 +183,68 @@ test_that("sample() method works with init file", {
...
+  skip_on_cran()
+
+  inv_metric_vector <- array(1, dim = c(1))
+  inv_metric_matrix <- matrix(1, nrow = 1, ncol = 1)
+
+  inv_metric_vector_json <- test_path("resources", "data", "bernoulli.inv_metric.diag_e.json")
+  inv_metric_matrix_json <- test_path("resources", "data", "bernoulli.inv_metric.dense_e.json")
+
+  inv_metric_vector_r <- test_path("resources", "data", "bernoulli.inv_metric.diag_e.data.R")
+  inv_metric_matrix_r <- test_path("resources", "data", "bernoulli.inv_metric.dense_e.data.R")
```
jgabry: I don't actually see these files in `tests/resources/data` so using them in tests below isn't going to work. Is the intention to save some inverse metrics in those files and then make sure they can be used successfully as the `metric_file` argument?

jgabry:
Also, what kinds of tests did you implement outside the testing framework?

bbbales2:
> Also, what kinds of tests did you implement outside the testing framework?

Eh just had a .R script with some stuff and calling the functions with all sorts of different combos of arguments.

Just seemed like the wrong way of doing the dev cause then the actual tests are just another annoying thing to do afterwards. This seems like the type of code that could be written to the tests.

Thanks for all the feedback, hoooray!

jgabry:
> Just seemed like the wrong way of doing the dev cause then the actual tests are just another annoying thing to do afterwards. This seems like the type of code that could be written to the tests.

Ideally, yeah. If you want help doing that let me know!

R/args.R
bbbales2: Well it kinda sucks from an interface perspective. We could try to do a combined argument, but I just didn't want to call an argument where you're passing in the inverse metric by value metric_file.

R/args.R
bbbales2: Also should I sapply here or should we just wrap the utilities with vectorize?

R/model.R
bbbales2: Switched

R/args.R
bbbales2: Also absolute_path does call repair path, good call, fixed

R/model.R
bbbales2: I'm doing cntrl+shift+B to rebuild and reinstall cmdstanr, but when I do '?`model-method-sample`' to check the docs, I don't see the updated docs. Is there something else I should be doing? Or am I looking at the wrong docs?

bbbales2:
Okeedokee. I made changes. My questions now are:

1. How do I test the doc? I'm not seeing the changes I made end up in '?`model-method-sample`'

2. I moved the tests to test-metric.R. I like the broken out tests so I can run them really fast. Do you mind if we break out the tests like this? Is there some strategy you'd like for organizing them?

3. I duplicated the definition of ```expect_sample_output``` in my test file. That doesn't seem appropriate. How should the common test helpers should be organized?

R/args.R
jgabry: I was actually just suggesting changing the error message to 

```r
stop("Only one of inv_metric and metric_file can be specified.", call. = FALSE)
```

but I agree that it's not ideal from an interface perspective. (super weird that CmdStan uses `metric_file` for specifying `inv_metric`)

bbbales2: Oh lol whoops. Yeah I'll change that.

R/args.R
jgabry: We should probably vectorize `absolute_path` but not a problem if you want to leave it with `sapply` for now.

R/model.R
jgabry: Ah, good question. Two options:

1) Run `devtools::document()` before rebuilding with `cntrl+shift+B`

2) If you want it to automatically do it when rebuilding (which I do, but I think has to be set individually): 
  - open the RStudio project options (for me on my mac if I click on the rstudio projects list in the top right corner then "Project Options" is the last thing in the list after all the project names but might be different for you)
  - Once in the project options interface go to the "Build Tools" panel
  - Make sure "Generate documentation with roxygen" has a check mark and then click the "Configure..." button. 
  - In the window opened by "Configure..." make sure everything listed under "Automatically roxygenize when running:" has a check mark

jgabry:
Thanks! 

1) Regarding the doc, I just responded to your original question about that above, but quoting here: 

> Ah, good question. Two options:
> 
> 1. Run `devtools::document()` before rebuilding with `cntrl+shift+B`
> 2. If you want it to automatically do it when rebuilding (which I do, but I think has to be set individually):
> 
> * open the RStudio project options (for me on my mac if I click on the rstudio projects list in the top right corner then "Project Options" is the last thing in the list after all the project names but might be different for you)
> * Once in the project options interface go to the "Build Tools" panel
> * Make sure "Generate documentation with roxygen" has a check mark and then click the "Configure..." button.
> * In the window opened by "Configure..." make sure everything listed under "Automatically roxygenize when running:" has a check mark

2) Fine to split out the metric tests from the others. And yeah we can split the others out too! Let's open a separate issue for that so we can discuss the best option for organizing that. We don't need to wait for that for this PR

3) Yeah there are a few options. I've handled this in several different ways in other packages before. Among the options are: 
    - (a) define these functions as internal (not exposed to user) functions in the cmdstanr package (then they're available for use in the test files. 
   - (b) create an R file in `tests/testthat` that has the definitions of these helper functions and then `source` that file as needed
   - (c) use the testthat functionality described here: https://www.tidyverse.org/blog/2017/12/testthat-2-0-0/#setup-and-teardown

But this is also something I'd be fine with addressing in a separate issue/PR since it's more of a general improvement than specific to the metric issue. Up to you though.

R/args.R
```diff
@@ -147,15 +149,36 @@ SampleArgs <- R6::R6Class(
...
+            fileext = ".json"
+          )
+        for(i in seq_along(inv_metric_paths)) {
+          write_stan_json(list(inv_metric = inv_metric[[i]]), inv_metric_paths[i])
+        }
+
+        self$metric_file <- inv_metric_paths
+      } else {
+        self$metric_file <- sapply(metric_file, absolute_path)
+      }
```
jgabry: It seems that tests are still failing (not your metric tests) because now _not_ including `inv_metric` or `metric_file` results in an error (that's the hazard of just running the metric tests!). So allowing for `NULL` in `validate_metric_file` wasn't actually enough. I think what will do the trick is changing the `else` clause above to this:

```r
      } else if (!is.null(metric_file)) {
        self$metric_file <- sapply(metric_file, absolute_path)
      }
```

bbbales2:
```R
} else if (!is.null(metric_file)) {
        self$metric_file <- sapply(metric_file, absolute_path)
}
```

This came up here as well: https://github.com/stan-dev/cmdstanr/pull/67

> Fine to split out the metric tests from the others. And yeah we can split the others out too! Let's open a separate issue for that so we can discuss the best option for organizing that. We don't need to wait for that for this PR

Did it. We'll see what folks think and make a decision.

I don't think there's a huge reason to break out tests that already exist or keep them organized in a super consistent way.

> use the testthat functionality described here:

That sounds cool. I think coming from the outside it'd be easier to understand what source was doing, but, if setup files are how people do things then we can just roll with that.

bbbales2:
How should I interpret this failure: https://travis-ci.org/stan-dev/cmdstanr/jobs/608400371 ? Is that just spurious or is there something happening?

jgabry:
I restarted the travis job. Sometimes the download of the .tar.gz from https://github.com/stan-dev/cmdstan/releases gets messed up for some reason.

jgabry:
@bbbales2 Aside from the travis issue are you good with this PR? If there's nothing else on your end then I'll just go ahead and merge this whenever I get travis working.

bbbales2:
Yeah I'm good with this. Ready to start using it.

PR #65, opened 2019-11-06
Add include_paths argument to $compile() method

### LINKED ISSUE #59: Allow specifying include paths
Thanks to @mitzimorris for reminding us to check this. Right now it doesn't seem to be possible to compile via CmdStanR if using `#include` statements in the Stan program. 

@rok-cesnovar @bbbales2 Here's some code that produces the error (using some Stan programs I just added to the tests directory on the master branch)

```r
stan_file <- testthat::test_path("resources", "stan", "bernoulli_include.stan")
mod <- cmdstan_model(stan_file, quiet = FALSE) # will error
```

apparently because CmdStan can't find the Stan program specified by the `#include` statement (`divide_real_by_two.stan`) even though it's in the same directory as the main Stan program (`bernoulli_include.stan`).

jgabry:
closes #59

This PR adds an `include_paths` argument to the `$compile()` method and uses it with `make` directly by adding `STANCFLAGS += --include_paths`. It didn’t seem necessary to use `make/local`, but @rok-cesnovar do you see any problems with handling the `include_paths` in this way? Would it be better to do it via `make/local` (or some other way)? 

@bbbales2 feel free to chime in as always, but no obligation!

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -154,6 +154,7 @@ CmdStanModel <- R6::R6Class(
 #'   ```
 #'   $compile(
 #'     quiet = TRUE,
+#'     include_paths = NULL,
 #'     threads = FALSE,
 #'     opencl = FALSE,
 #'     opencl_platform_id = 0,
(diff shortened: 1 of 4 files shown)


jgabry:
Thanks!

Issue #66, opened 2019-11-06
Add ability to specify init_buffer, term_buffer, and warmup arguments


bbbales2:
I want these for my experimentation.

They're important in the situation that you want to set the metric but allow Stan to still do timestep adaptation.

jgabry:
Sure!

PR #67, opened 2019-11-06
Added init_buffer, term_buffer, window arguments. (Issue #66)

bbbales2:
Addresses the issue here: https://github.com/stan-dev/cmdstanr/issues/66


bbbales2:
These bad values weren't being tested previously so I added them in: https://github.com/stan-dev/cmdstanr/blob/e71ee47a07b8f97b991c9232b602d3300e5f08f7/tests/testthat/test-model.R#L219

To avoid warnings I had to add some extra stuff here and a couple lines below: https://github.com/stan-dev/cmdstanr/blob/e71ee47a07b8f97b991c9232b602d3300e5f08f7/R/args.R#L42

tests/testthat/test-model.R
```diff
@@ -165,17 +168,24 @@ if (NOT_CRAN) {
...
     stepsize = 1:10,
-    metric = c("AA", "BB")
+    metric = c("AA", "BB"),
+    init_buffer = -5,
+    term_buffer = -6,
+    window = -7
   )
 
   bad_arg_values_3 <- list(
+    data_file = rep("NOT_A_FILE", 5),
```
jgabry: There actually isn’t a `data_file` argument to the sample method. All options are handled via `data` and the name`data_file` is just used internally. So this would error but because the argument doesn’t exist. 

I think the best option might be to change `data_file` to `data` here and then assert that the internal `self$data_file` is length 1 when validating the CmdStanArgs object. I’m presuming we want it to error if a user tries to specify multiple data files, right?

R/model.R
```diff
@@ -366,7 +375,10 @@ sample_method <- function(data = NULL,
                           metric_file = NULL,
                           inv_metric = NULL,
                           stepsize = NULL,
-                          max_depth = NULL) {
+                          max_depth = NULL,
+                          init_buffer = NULL,
+                          term_buffer = NULL,
+                          window = NULL) {
```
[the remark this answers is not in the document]
bbbales2: Shuffled the arguments to look like:

```R
thin = NULL,
max_depth = NULL,
adapt_engaged = TRUE,
adapt_delta = NULL,
stepsize = NULL,
metric = NULL,
metric_file = NULL,
inv_metric = NULL,
init_buffer = NULL,
term_buffer = NULL,
window = NULL
```

tests/testthat/test-model.R
bbbales2: I just removed this test. I guess we'd want to error if we had a list of lists or something?

But there wasn't a bug I was fixing with this -- just saw the tests were unused and needed em' so I put them back. So I just removed the leftover test.

R/model.R
```diff
@@ -275,12 +275,15 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
 #'     save_warmup = FALSE,
 #'     thin = NULL,
 #'     max_depth = NULL,
+#'     adapt_engaged = TRUE,
+#'     adapt_delta = NULL,
+#'     stepsize = NULL
```
jgabry: missing comma after NULL

R/model.R
```diff
@@ -275,12 +275,15 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
...
 #'     metric = NULL,
 #'     metric_file = NULL,
-#'     inv_metric = NULL,
-#'     stepsize = NULL,
-#'     adapt_engaged = TRUE,
-#'     adapt_delta = NULL
+#'     inv_metric = NULL,,
```
jgabry: too many commas

PR #77, opened 2019-11-12
Improve `$summary()` method

### LINKED ISSUE #55: Improve summary method
Right now `fit$summary()` just prints CmdStan's `bin/stansummary`, but this is suboptimal. I'm assigning myself because I'm already working with @paul-buerkner on a nicer version of summarizing draws in general that we can use in CmdStanR. Should be ready soon.

jgabry:
Resolves #55 

This PR uses our still unreleased (but now available via GitHub) `posterior` package to provide summaries of the MCMC draws (and approximate draws from VB). 

`posterior` (https://github.com/jgabry/posterior, will eventually move to `stan-dev`) is a new package that will be used behind the scenes by various R packages we develop (e.g., rstanarm, brms, etc.) to deal with summarizing and transforming between different formats for posterior draws (instead of reimplementing stuff in many packages). 

@paul-buerkner since cmdstanr is still unreleased I think this makes it a good choice as a test case for putting posterior in a package. I added you as a reviewer in case you have any thoughts on how I’m using it here, but it’s pretty straightforward, which is great (since that’s why we made it!)

Changes:

* Until it is released, the `posterior` package is automatically installed from GitHub (instead of CRAN) when installing CmdStanR from GitHub with `devtools::install_github` (via `Remotes` field in `DESCRIPTION`). This seems ok as long as posterior eventually makes it to CRAN before cmdstanr. 
* The `$summary()` method now calls `posterior::summarise_draws()`. This is just for CmdStanMCMC and CmdStanVB objects  since there are no draws for CmdStanMLE objects. 
* The CmdStan summary from `bin/stansummary` is still available but now only via the `$cmdstan_summary()` method. `$diagnose()` has also been renamed `$cmdstan_diagnose()` for consistency. Both of these now call `runset$run_cmdstan_tool()` internally.
* The `$draws()` methods for CmdStanMCMC and CmdStanVB objects convert the draws to `posterior::draws_array` and `posterior::draws_matrix` formats, respectively. (@rok-cesnovar This shouldn't interfere with what you're working on to read in the CSV output. This conversion happens after `private$draws_` is already created.)

#### Example

```r
mod <- cmdstan_model(stan_file = test_path("resources/stan/logistic.stan"))
fit_mcmc <- mod$sample(data = test_path("resources/data/logistic.data.json"))
```
```r
fit_mcmc$summary()
```
```
# A tibble: 5 x 10
  variable    mean  median    sd   mad       q5      q95  rhat ess_bulk ess_tail
  <chr>      <dbl>   <dbl> <dbl> <dbl>    <dbl>    <dbl> <dbl>    <dbl>    <dbl>
1 alpha      0.376   0.371 0.222 0.220   0.0136   0.750   1.00    4252.    3083.
2 beta[1]   -0.672  -0.663 0.247 0.245  -1.10    -0.284   1.00    4491.    3103.
3 beta[2]   -0.274  -0.270 0.220 0.217  -0.640    0.0773  1.00    4277.    3049.
4 beta[3]    0.680   0.682 0.262 0.262   0.257    1.11    1.00    4059.    3055.
5 lp__     -65.9   -65.6   1.46  1.21  -68.7    -64.3     1.00    2002.    2564.
```

```r
fit_mcmc$summary(measures = "quantile", probs = c(0.25, 0.75))
```
```
# A tibble: 5 x 3
  variable     q25     q75
  <chr>      <dbl>   <dbl>
1 alpha      0.229   0.524
2 beta[1]   -0.835  -0.503
3 beta[2]   -0.420  -0.127
4 beta[3]    0.502   0.855
5 lp__     -66.6   -64.9  
```


jgabry:
update: i think travis will fail until https://github.com/jgabry/posterior/issues/49 is resolved

jgabry:
> I am not completely sure that we actually still need `cmdstan_summary_method`. As far as I understand this new summary has all the features that `stansummary` has plus the option of actually retrieving the numbers printed and more. Its not a big effort to have and maintain both, so I guess we can leave it.

I agree we don't necessarily need to keep `cmdstan_summary()`. I'd say lets keep it temporarily at least. Once we have divergence warnings etc. then we could get rid of both `cmdstan_diagnose()` and `cmdstan_summary()` at the same time if we want. We can decide then.

SteveBronder:
Dunno if this is the right place to ask, but how hard would it be to get back the 'pretty names' like rstanarm and brms have for the parameter names?

Issue #76, opened 2019-11-11
Output handling with failed chains


rok-cesnovar:
From @bbbales2 in #61 :

- output_files gives us all the output files for things that work and I like that. But it would still be good to access files of chains that failed for debugging. Can we add an included_failed = FALSE argument to output_files and diagnostic_files? Probably should add that flag with save_output_files and save_diagnostic_files as well -- though in this cause I kinda feel like the default should be the opposite.

- diagnostic_files returns files from all chains, whether they succeeded or failed.

By default the output files and draws, etc. should be returned only for chains that succeeded. With include_failed = TRUE we would also include available info for failed chains. Take special care for cases where chains finish extremely early and there might be no info to show.

jgabry:
Sounds good to me!

rok-cesnovar:
@bbbales2 do you maybe have a model that runs for some time and then fails? I would need such a model to test this and am failing to come up with a tests model. The one we currently have for failed chains either finishes or fails immediately. Thanks!

bbbales2:
Check this one: https://github.com/stan-dev/cmdstanr/blob/master/tests/testthat/resources/stan/chain_fails.stan

It takes one argument which is the probability of failure (and it'll only fail at runtime).

rok-cesnovar:
I checked that one yeah and when it failed I found no samples in the csv file.

rok-cesnovar:
I am actually not sure now if cmdstan actually outputs samples to the csv if chains fail?

bbbales2:
If you get one that blows up halfway through, you'll get half the output. But I think this issue doesn't need that. The empty files can be annoying if you have scripts that expect cmdstan output.

jgabry:
@rok-cesnovar @bbbales2 Is this still an issue or was this addressed by one of the recent or semi-recent PRs?

rok-cesnovar:
Still an issue. I plan on tackling this after the fix to #131 and the new compile flags system.

jgabry:
Ok thanks!

rok-cesnovar:
@bbbales2 Do you have a model that fails half way through or have a trick up you sleeve for forcing it? It would really help when adding this. The model above either fails immediatelly or runs fine.
I am going to tackle this tomorrow or on Thursday.

bbbales2:
Oof, nah. I think those failures come from somebody killing a process while it's running.

[...]

PR #79, opened 2019-11-12
change internal paths after `$save_*_files()`

### LINKED ISSUE #73: `save_output_files` should change internal output_files too
I think methods like `fit$save_output_files()` should do two more things in addition to what they currently do: 

(1) remove the temporary files after copying them to the new location
(2) change the value of `fit$runset$private$output_files_` to point to the new location where the user saved them

jgabry:
Fixes #73 

This PR adds two new behaviors when the user calls `fit$save_output_files()`, `fit$save_diagnostic_files()`, and `$save_data_file()`:

(1) remove the temporary files after copying them to the new non-temporary location
(2) change internally stored paths to point to the new location where the user saved them (for this we probably want to use `include_failed=TRUE` after we have #76). 

#### Copyright and Licensing

**Columbia University**


jgabry:
> (1) remove the temporary files after copying them to the new non-temporary location specified by the user
> (2) change internally stored paths to point to the new location where the user saved them (for this we probably want to use `include_failed=TRUE` after we have #76).

@bbbales2 @rok-cesnovar The code changes here are minimal, but just want to double check if you guys are ok with this new behavior?

bbbales2:
Can test this tomorrow. The only question I have reading the docs is why put the timestamp in the name? Why not lean on some sort of tempfile mechanism for the filenames?

jgabry:
We could use something other than a timestamp. It seemed nicer to add time to the user's file basename than a bunch of random letters (a la tempfiles) since these functions are for saving in a non-temporary fashion. But I'm not attached to the timestamp.

bbbales2:
I ran this with a model that fails randomly:

```
> fit$save_output_files()
Moved 3 output files and set internal paths to new locations:
- ./error-1_20191114-0944.csv
- ./error-2_20191114-0944.csv
- ./error-3_20191114-0944.csv
- ./error-4_20191114-0944.csv
- ./error-5_20191114-0944.csv
```

All 5 files were created and are filled up with stuff, which seems weird. Two of those should totally fail to run!

The diagnostic output seems to be appropriately small, there could be something going on here:
```
bbales2@frog:~/cmdstanr$ ls -alh error-*944*
-rw-rw-r-- 1 bbales2 bbales2 937K Nov 14 09:44 error-1_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 938K Nov 14 09:44 error-2_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 937K Nov 14 09:44 error-3_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 937K Nov 14 09:44 error-4_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 938K Nov 14 09:44 error-5_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 2.7M Nov 14 09:44 error-diagnostic-1_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 1.2K Nov 14 09:44 error-diagnostic-2_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 1.2K Nov 14 09:44 error-diagnostic-3_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 2.7M Nov 14 09:44 error-diagnostic-4_20191114-0944.csv
-rw-rw-r-- 1 bbales2 bbales2 2.7M Nov 14 09:44 error-diagnostic-5_20191114-0944.csv
```

Saving output twice fails:
```
fit$save_output_files()
Moved 4 output files and set internal paths to new locations:
- NA
- NA
- NA
- NA
- NA
Warning messages:
1: In file.remove(self$output_files()) :
  cannot remove file './error-1_20191114-0943.csv', reason 'No such file or directory'
2: In file.remove(self$output_files()) :
  cannot remove file './error-2_20191114-0943.csv', reason 'No such file or directory'
3: In file.remove(self$output_files()) :
  cannot remove file './error-3_20191114-0943.csv', reason 'No such file or directory'
4: In file.remove(self$output_files()) :
  cannot remove file './error-4_20191114-0943.csv', reason 'No such file or directory'
```

If you do that really quickly you get:
```
Error in file.copy(from = current_paths, to = destinations, overwrite = TRUE) : 
  file can not be copied both 'from' and 'to' 
```

> We could use something other than a timestamp. It seemed nicer to add time to the user's file basename than a bunch of random letters (a la tempfiles)

Fair point. And timestimes are kinda readable and have the advantage of arranging themselves by date if we put the date first. How about modelname-timestamp-chain-tmpfilesomething? I have no strong feelings on this the thing that seemed weird to me was the warning in the docs:

```
If files with the specified names already exist they are overwritten, but
#' this shouldn't occur unless the `timestamp` argument has been intentionally
#' set to `FALSE`.
```

Just seems like an avoidable type of problem.

bbbales2:
(the saving twice also fails for diagnostics and data)

Also I didn't test optimization or vb. I guess I'll check those next round.

bbbales2:
Oh yeah the handy randomly failing model is:
```
data {
  int<lower=1> N;
}

transformed data {
  int run = 0;
  int b = bernoulli_rng(0.5);

  if(b > 0) {
    run = 1;
  }
}

parameters {
  vector[N] mu;
}

model {
  if(run == 1) {
    mu ~ normal(0, 1);
  } else {
    mu ~ normal(1 / 0.0, 1);
  }
}
```

jgabry:
Thanks @bbbales2! How about adding seconds to the timestamp? That fixes the saving twice issue and doesn't require adding random characters like tempfile style does. What do you think? I went ahead and pushed that so you can try it out but I'm open to going a different route. 

I also think I fixed the other issue with the right number of files getting saved. It now should save all output and diagnostic files and set all the paths (at least until #76).

jgabry:
@bbbales2 I’ll add a test or several with the failing chains example, thanks! What do you think about my idea of adding seconds to the timestamp to avoid the repeat saving issue? Or do you still prefer random characters/digits?

bbbales2:
Was just cloning to check this again.

> What do you think about my idea of adding seconds to the timestamp to avoid the repeat saving issue? Or do you still prefer random characters/digits?

It's fine whichever way honestly. The documentation clearly describes how it works. The overwrite just seemed a bit strange to me but I dunno if it matters.

If I was using this featured I'd probably encode some sorta unique identifier in basename anyway.

bbbales2:
The file sizes make sense now and the double saves don't give errors.

Should we do absolute paths? I guess these things might go away sometime in the short future, but:

```
> fit$cmdstan_diagnose()
Running bin/diagnose ./error-2_20191115-145605.csv ./error-3_20191115-145605.csv ./error-4_20191115-145605.csv \
  ./error-5_20191115-145605.csv
File ./error-2_20191115-145605.csv not found
File ./error-3_20191115-145605.csv not found
File ./error-4_20191115-145605.csv not found
File ./error-5_20191115-145605.csv not found
No valid input files, exiting.
> fit$cmdstan_summary()
Running bin/stansummary ./error-2_20191115-145605.csv ./error-3_20191115-145605.csv \
  ./error-4_20191115-145605.csv ./error-5_20191115-145605.csv
File ./error-2_20191115-145605.csv not found
File ./error-3_20191115-145605.csv not found
File ./error-4_20191115-145605.csv not found
File ./error-5_20191115-145605.csv not found
No valid input files, exiting.
```

jgabry:
> The overwrite just seemed a bit strange to me but I dunno if it matters.

On the off chance that (a) `timestamp` is manually set to FALSE _and_ (b) the user doesn't add a timestamp or tempfile thing to `basename` themselves _and_ (c) there already exist a files in `dir` named `<basename>-<chain_id>` do you think we should error or overwrite those files? This is only relevant if all of (a), (b), and (c) are true, otherwise nothing is ever overwritten. I'm happy to throw an error instead if you prefer. 

> If I was using this featured I'd probably encode some sorta unique identifier in basename anyway.

True. you can always just set `timestamp=FALSE` and do it in basename yourself. 

> Should we do absolute paths? I guess these things might go away sometime in the short future,

Good call. Should be fixed now.

jgabry:
I also added tests inspired by your randomly failing chains model (I like that one!). I tweaked it a bit so i can control how often it fails (even guarantee it) from the outside, which is helpful when testing: 

```stan
data {
  real<lower=0,upper=1> pr_fail;
}
transformed data {
  int fail = bernoulli_rng(pr_fail);
}
parameters {
  real theta;
}
model {
  if (fail) {
    theta ~ normal(1 / 0.0, 1);
  } else {
    theta ~ normal(0, 1);
  }
}
```

jgabry:
(Also we can ignore appveyor failures as long as Travis is passing. Appveyor isn’t fully set up yet)

bbbales2:
There's extra dots in the output filenames now but it works fine:

```
> fit$save_output_files()
Moved 5 output files and set internal paths to new locations:
- /home/bbales2/cmdstanr/./error-1_20191116-112913.csv
- /home/bbales2/cmdstanr/./error-2_20191116-112913.csv
- /home/bbales2/cmdstanr/./error-3_20191116-112913.csv
- /home/bbales2/cmdstanr/./error-4_20191116-112913.csv
- /home/bbales2/cmdstanr/./error-5_20191116-112913.csv
```

> (a), (b), and (c) are true

Running cmdstanr on a cluster with a shared filesystem, two models named the same thing, and they finish in the same second, and write to the same output folder. Sure it's careless but who reads docs anyway :D?

Tempfile isn't gonna guarantee against race conditions but it'll put a few extra zeros in front of that probability.

But it's cool I just don't know if the names of these files are that important and so I was happy with the tempfile naming tmp. This discussion probably goes with a discussion on serializing cmdstanr objects, since long term storage seems like what's in order here.

jgabry:
Thanks! 

Regarding the temp/timestamp issue, ok yeah let’s just be extra safe about it! I’ll add some random tempfile-like thing to the names. 

> If you wanna fix the '/./' in the paths that'd be cool.

I hadn’t even noticed that! Will fix.

jgabry:
> How about modelname-timestamp-chain-tmpfilesomething?

@bbbales2 That's what I ended up going with: 

```
> fit$save_output_files()
Moved 2 output files and set internal paths to new locations:
- /Users/jgabry/Documents/Stan/cmdstanr/bernoulli-201911161358-1-31f27.csv
- /Users/jgabry/Documents/Stan/cmdstanr/bernoulli-201911161358-2-31f27.csv
```

The `201911161358` is the timestamp (2019 Nov 16 13:58) and the `31f27` is a random 5-character alphanumeric string. So it's `basename-timestamp-chain-random`.

jgabry:
Ok great, thank!

PR #93, opened 2019-11-20
Install cmdstan via r

### LINKED ISSUE #88: Dont use bash on windows for install_cmdstan()
The install_cmdstan() command currently uses bash, which is not native on Windows and is not installed with RTools. The easies way of installing it currently is to have Git for Windows. Its not the worst of dependencies one can have, but it would still be nice to get rid of it. This was pointed out [here](https://discourse.mc-stan.org/t/cmdstanr-installation-error-command-not-found-win-processx-c-977/11970).

The solution would be to have to most of the work done in R and only call curl, mingw32-make, tar, etc in the command line. I dont think powershell scripting or .bat files are worth the trouble.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #88 

(@rok-cesnovar I know I've opened a lot of PRs recently, so no worries if it takes you some time to get to this!)

This PR changes `install_cmdstan()` so that it installs CmdStan without needing the script `make_cmdstan.sh`. Standard R functions are used for downloading and unzipping the tarball before running make via `processx::run()`. 

I also updated the travis and appveyor scripts to use `install_cmdstan()` instead of `make_cmdstan.sh`. The only reason I haven't removed `make_cmdstan.sh` entirely is that it's still used by `cmdstan_git_checkout_branch()`, which I haven't touched in this PR (although we could do something similar for that).  

#### Copyright and Licensing

**Columbia University**


tests/testthat/test-install.R
```diff
@@ -1,35 +1,51 @@
...
+    dir <- NULL
   } else {
     dir <- tempdir()
   }
   if (dir.exists(file.path(dir, "cmdstan"))) {
-    expect_output(
-      install_cmdstan(dir = dir, cores = 1, quiet = FALSE, overwrite = FALSE),
-      "Please remove or rename the 'cmdstan' folder"
+    expect_warning(
+      install_cmdstan(dir = dir, overwrite = FALSE),
```
rok-cesnovar: I think there is a quiet = FALSE missing here? This tests is probably skipped on Travis and that is why its not failing?

R/install.R
```diff
@@ -128,3 +152,82 @@ cmdstan_git_checkout_branch <- function(repo_branch,
...
+    stderr_line_callback = function(x,p) { if(quiet) message(x) }
+  )
+}
+
+build_status_ok <- function(process_log, quiet = FALSE) {
+  if (is.na(process_log$status) || process_log$status != 0) {
+    if (!quiet) {
+      suggestion <- "See the error message(s) above."
+    } else {
+      suggestion <- "Please try again with 'quiet=FALSE' to get the full error output."
```
rok-cesnovar: The error messages will be output here in any case because of the 

    stderr_line_callback = function(x,p) { if(quiet) message(x) }

So maybe just write out "suggestion <- "See the error message(s) above or try again with 'quiet=FALSE' to get the full installation output."

R/install.R
```diff
@@ -128,3 +152,82 @@ cmdstan_git_checkout_branch <- function(repo_branch,
...
+  processx::run(
+    make_cmd(),
+    args = c(paste0("-j", cores), "build"),
+    wd = dir,
+    echo_cmd = FALSE,
+    echo = !quiet,
+    spinner = quiet,
+    error_on_status = FALSE,
+    stderr_line_callback = function(x,p) { if(quiet) message(x) },
+    timeout = 600 # timeout after 10 minutes to avoid infinite loop problem?
```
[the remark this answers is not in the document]
jgabry: Good call. I’ll make it an argument to install_cmdstan.

jgabry:
Thanks for the review! Will make those changes shortly

tests/testthat/test-install.R
jgabry: It was actually a different reason, but fixed now!

jgabry:
@rok-cesnovar Waiting to see if travis and appveyor pass now, but I addressed all the comments so I think this is ready for review again.  

> Should we alse delete the bash script in this PR?

The only reason we can't delete the bash script is because it's still used by the `cmdstan_git_checkout_branch()` function. We can definitely change `cmdstan_git_checkout_branch()` to not use it either, but maybe as a separate PR?

jgabry:
Ok great thanks. I’ll merge if CI passes.

jgabry:
Looks like Travis is passing but timing out during codecov. I’ll try adding travis_wait as described here:

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Issue #96, opened 2019-11-23
Dimensions dropped from data object, causing exception


cdriveraus:
In my stan model I have a data object like:
matrix[ntipred ? nsubjects : 0, ntipred ? ntipred : 0] tipredsdata;

When ntipred is 0, and hence it is a 0x0 matrix, cmdstanr seems to drop one of the 0 dimensions from the tipredsdata object. In R the structure of the object is correct, num[0 , 0 ] , but cmdstanr reports it as dims found = (0) . 

"Exception: mismatch in number dimensions declared and found in context; processing stage=data initialization; variable name=tipredsdata; dims declared=(0,0); dims found=(0) "

rok-cesnovar:
Thanks for reporting. So just to make sure you are inputing a R data.matrix of size 0,0?

cdriveraus:
Correct.

bbbales2:
@rok-cesnovar do you know how to pass zero by zero matrices into the json Cmdstan?

If that's not possible, can we switch back to using stan_rdump until the 0x0s are okay with json? Should just be a matter of making a copy of the function here:

https://github.com/stan-dev/cmdstanr/blob/master/R/utils.R#L261

and changing the json write to:

```stan_rdump(names(data), file, env = list2env(data))```

and changing the reference here:

https://github.com/stan-dev/cmdstanr/blob/master/R/model.R#L661

I can submit a pull req.

**Comment by cdriveraus** [comment]
just a pointer that you're probably on top of -- many types of objects can be zero dimensional and have a corresponding single or multi dimensional array, which can also be zero dimensional. I just remember stan going through the procedure that some things were fixed for zero dimensions while others were not :)

**Comment by rok-cesnovar** [comment]
@bbbales2 Will have a look, it should definitely be doable. I just need to make a simple example that I can test on.

**Comment by bbbales2** [comment]
I just played with this a little and didn't figure out how to pass in 0x0s with cmdstan:
```
data {
  matrix[0, 0] x;
}

parameters {
  real a;
}

model {
  a ~ normal(0, 1);
}
```

Here's data:

```
data = list(N = 0, x = array(1, dim = c(0, 0)))
write_stan_json(data, "tmp.json")
stan_rdump(names(data), "tmp.dat", env = list2env(data))
```

**Comment by bbbales2** [comment]
It serializes to (Ignore the N):
```
{
  "N": 0,
  "x": []
}
```

The declaration:

```
"x" : [[]]
```

[...]

PR #100, opened 2019-11-25
Switch back to using Rdata temporarily while issue with 0x0s gets resolved for JSON

bbbales2:
This swaps stan_rdump in for write_stan_json to handle https://github.com/stan-dev/cmdstanr/issues/96. We expect to go back to write_stan_json once cmdstan can handle multidimensional size zero arguments via json.

#### Submission Checklist

- [] Run unit tests

#### Copyright and Licensing

Columbia University

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,12 +24,12 @@ Imports:
     checkmate,
     posterior (>= 0.0.1),
     processx,
+    rstan,
     R6
 Suggests: 
     jsonlite (>= 1.2.0),
     knitr,
     rmarkdown,
-    rstan,
     testthat (>= 2.1.0)
 VignetteBuilder: knitr
 Remotes:  
--- a/R/model.R
+++ b/R/model.R
@@ -665,7 +665,7 @@ process_data <- function(data) {
     path <- absolute_path(data)
   } else if (is.list(data) && !is.data.frame(data)) {
     path <- tempfile(pattern = "standata-", fileext = ".dat")
-    stan_rdump(names(data), file = path, env = list2env(data))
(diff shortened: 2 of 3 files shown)


bbbales2:
So I'm gonna try a lawyer trick and say I don't need to write tests for stan_rdump cause we assume rstan is handling those. I guess the big change here is that we'd be leaning on the rstan package again.

rok-cesnovar:
You don't need tests for stan_rdump, but I think some tests have .json resources they run with. Not completely sure.

bbbales2:
> I can tackle making a rdump function so we dont depend on rstan once the read_csv goes through.

This seems more like a problem with cmdstan's json reader than any problem with cmdstanr. I don't see a need to do extra work to work around this (when it'll need fixed anyway).

jgabry:
Thanks. Any idea on the timeline for this getting fixed in CmdStan? 

Since most models aren't affected by this I think it would still be good to allow JSON as an option since it's faster. Thoughts?  

(Also looks like CI is failing)

rok-cesnovar:
> Any idea on the timeline for this getting fixed in CmdStan?

It can be fixed soon if we agree on the fix. The issue is that JSON as a format has no theoretical way of specifying that an array is dim(0,0). Its either dim(0) or dim(1,0). So we need to resort to a slight "hack" to do this.

rok-cesnovar:
Lets merge this so we fix the issue. I will make an new PR to allow the use of json or to just rdump whenever we have this case of 0x0.

bbbales2:
@rok-cesnovar thanks, I just left a cmdstan issue over here for the JSON parsing thing: https://github.com/stan-dev/cmdstan/issues/775

PR #110, opened 2020-01-16
With Cmdstan 2.22+ use JSON by default

### LINKED ISSUE #109: Make JSON input the default option
In order to fix #96 we reverted back to using Rdump for the cmdstan input. Now that @mitzimorris has fixed the underlying Cmdstan issue that was the root of #96, we should move back to using JSON as the default input option.

We can do it after the release of 2.22 on the 18th January. We could do it now, but that would mean that those affected by this issue would need to use a git clone of cmdstan.

rok-cesnovar:
Reverts stan-dev/cmdstanr#100 since the underlying cmdstan issue will be resolved for 2.22

Closes #109

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -664,8 +664,8 @@ process_data <- function(data) {
   } else if (is.character(data)) {
     path <- absolute_path(data)
   } else if (is.list(data) && !is.data.frame(data)) {
-    path <- tempfile(pattern = "standata-", fileext = ".dat")
-    rstan::stan_rdump(names(data), file = path, env = list2env(data))
+    path <- tempfile(pattern = "standata-", fileext = ".json")
+    write_stan_json(data = data, file = path)
   } else {
     stop("'data' should be a path or a named list.", call. = FALSE)
   }
--- a/R/model.R
+++ b/R/model.R
@@ -669,6 +669,9 @@ process_data <- function(data) {
       write_stan_json(data = data, file = path)
     } else {
       path <- tempfile(pattern = "standata-", fileext = ".dat")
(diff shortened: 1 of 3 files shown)


rok-cesnovar:
This is ready for review. Codecov on appveyor seems to not be working, so I decided to comment it out until we sort out the reason.

R/model.R
```diff
@@ -664,8 +664,13 @@ process_data <- function(data) {
   } else if (is.character(data)) {
     path <- absolute_path(data)
   } else if (is.list(data) && !is.data.frame(data)) {
-    path <- tempfile(pattern = "standata-", fileext = ".dat")
-    rstan::stan_rdump(names(data), file = path, env = list2env(data))
+    if (cmdstan_version() >= "2.22") {
+      path <- tempfile(pattern = "standata-", fileext = ".json")
+      write_stan_json(data = data, file = path)
+    } else {
+      path <- tempfile(pattern = "standata-", fileext = ".dat")
```
jgabry: Before calling `rstan::stan_rdump` we should check that rstan can be loaded. Can you add something like

```r
if (!requireNamespace("rstan", quietly = TRUE)) {
  stop("For CmdStan < 2.22 the rstan package is required for writing data.")
}
```
or some other message indicating rstan needs to be installed?

rok-cesnovar: Good call. Thanks

rok-cesnovar:
Appveyor tests pass but for some reason the codecov tests fail. I am not sure why.

jgabry:
I don't know why codecov is failing either, but this looks good. I'll merge it now.

PR #126, opened 2020-02-26
Refactor method for accessing draws and diagnostics

### LINKED ISSUE #123: Proposal for methods for accessing posterior draws and sampler diagnostics
After discussing with @mitzimorris, here is a proposal for user-facing methods for accessing the parameter (and generated quantities) draws and the sampler diagnostics: 

Suppose for example that we run 4 chains each with 1000 warmup and 1000 post-warmup iterations. For simplicity also assume the model has just 2 parameters. 

```r
# get post-warmup draws of the two parameters as 3-D array
fit$draws(inc_warmup=FALSE) # dim 1000 x 4 x 2

# get warmup and post-warmup draws together as 3-D array
fit$draws(inc_warmup=TRUE)  # dim 2000 x 4 x 2

# get post-warmup sampler diagnostics (accept_stat__, stepsize__, treedepth__, n_leapfrog__, divergent__, energy__) as 3-D array 
fit$sampler_diagnostics(inc_warmup=FALSE) # dim 1000 x 4 x 6

# get warmup and post-warmup sampler diagnostics together as 3-D array 
fit$sampler_diagnostics(inc_warmup=TRUE) # dim 2000 x 4 x 6
```

For R users these objects will be `draws_array` objects from the `posterior` package and then can easily be transformed to other formats using the functionality of `posterior`. 

# Questions

* where does `lp__` belong? Should it be a column in the output from `$draws()` or from `sampler_diagnostics()`? 

* what do you think of the names for the methods?

* other thoughts?

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #123 by adding 

```R
fit$draws(inc_warmup=FALSE) 
fit$draws(inc_warmup=TRUE) 
fit$sampler_diagnostics(inc_warmup=FALSE) 
fit$sampler_diagnostics(inc_warmup=TRUE) 
```

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -248,16 +248,20 @@ CmdStanMCMC <- R6::R6Class(
   public = list(
     initialize = function(runset) {
       super$initialize(runset)
-      data_csv <- read_sample_csv(self$output_files())
-      check_divergences(data_csv)
-      check_sampler_transitions_treedepth(data_csv)
-      private$draws_ <- data_csv$post_warmup_draws
-      private$sampler_diagnostics_ <- data_csv$post_warmup_sampler_diagnostics
-      private$sampling_info_ <- data_csv$sampling_info
-      if(!is.null(data_csv$sampling_info$save_warmup) 
-         && data_csv$sampling_info$save_warmup) {
-        private$warmup_draws_ <- data_csv$warmup_draws
-        private$warmup_sampler_diagnostics_ <- data_csv$warmup_sampler_diagnostics
+      if (!length(self$output_files())) {
+        warning("No chains finished successfully. Unable to retrieve the fit.")
+      } else {
+        data_csv <- read_sample_csv(self$output_files())
+        check_divergences(data_csv)
+        check_sampler_transitions_treedepth(data_csv)
+        private$draws_ <- data_csv$post_warmup_draws
+        private$sampler_diagnostics_ <- data_csv$post_warmup_sampler_diagnostics
+        private$sampling_info_ <- data_csv$sampling_info
+        if(!is.null(data_csv$sampling_info$save_warmup) 
+          && data_csv$sampling_info$save_warmup) {
+          private$warmup_draws_ <- data_csv$warmup_draws
+          private$warmup_sampler_diagnostics_ <- data_csv$warmup_sampler_diagnostics
+        }
       }
     },
     num_chains = function() {
@@ -303,6 +307,7 @@ CmdStanMCMC <- R6::R6Class(
     sampler_diagnostics_ = NULL,
     warmup_sampler_diagnostics_ = NULL,
     warmup_draws_ = NULL,
+    draws_ = NULL,
     read_csv_ = function() {
       if (!length(self$output_files())) {
         stop("No chains finished successfully. Unable to retrieve the fit.",
--- a/R/fit.R
+++ b/R/fit.R
@@ -273,6 +273,36 @@ CmdStanMCMC <- R6::R6Class(
       } else {
         cat(paste(self$runset$procs$chain_output(id), collapse="\n"))
       }
+    },
+
+    draws = function(inc_warmup = FALSE) {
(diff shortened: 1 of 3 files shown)


R/fit.R
```diff
@@ -273,6 +273,36 @@ CmdStanMCMC <- R6::R6Class(
...
+
+    draws = function(inc_warmup = FALSE) {
+      if (!length(self$output_files())) {
+        stop("No chains finished successfully. Unable to retrieve the fit.",
+             call. = FALSE)
+      }
+      if (is.null(private$draws_)) {
+        private$read_csv_()
+      }
+      if(inc_warmup && private$sampling_info_$save_warmup) {
```
rok-cesnovar: Do we want to warn users if they use inc_warmup = TRUE but save_warmup was false for the sampling?

rok-cesnovar:
There are 2 more open questions here: 
- where does lp__ belong? Its currently in draws()
- names

R/fit.R
```diff
@@ -273,6 +273,36 @@ CmdStanMCMC <- R6::R6Class(
       } else {
         cat(paste(self$runset$procs$chain_output(id), collapse="\n"))
       }
+    },
+
+    draws = function(inc_warmup = FALSE) {
+      if (!length(self$output_files())) {
+        stop("No chains finished successfully. Unable to retrieve the fit.",
```
jgabry: I think this needs to go after after checking if `private$draws_` is NULL. It shouldn't error if the temp output files are missing if the user has already read in the csv previously.

R/fit.R
jgabry: Yeah, at least a warning. Actually maybe an error makes more sense. But definitely either a warning or error

R/fit.R
```diff
@@ -273,6 +273,36 @@ CmdStanMCMC <- R6::R6Class(
...
+      if(inc_warmup && private$sampling_info_$save_warmup) {
+        posterior::bind_draws(private$warmup_draws_, private$draws_, along="iteration")
+      } else {
+        private$draws_
+      }
+    },
+
+    sampler_diagnostics = function(inc_warmup = FALSE) {
+      if (!length(self$output_files())) {
+        stop("No chains finished successfully. Unable to retrieve the fit.",
```
jgabry: This error will already be caught by `private$read_csv_()` so we can delete these lines.

R/fit.R
jgabry: Actually we don't need to throw an error at all. This error will already be caught by `private$read_csv_()` so we can delete these lines.

rok-cesnovar: Good call. Removed.

R/fit.R
rok-cesnovar: Removed.

R/fit.R
rok-cesnovar: Done. Added errors. Please let me know if the error messages need improving.

jgabry: > There are 2 more open questions here:
> 
> * where does lp__ belong? Its currently in draws()
> * names

For `lp__` I think I slightly prefer it to be accessed via `fit$sampler_diagnostics()` rather than `fit$draws()` but I'm not 100% sure. @bbbales2 @mitzimorris what do you think?

@rok-cesnovar For names, do you mean the names of the methods?

jgabry: Actually nevermind on moving `lp__`. let's keep it in `draws()` so that it's included in the `$summary()` output.

jgabry:
Let's keep `lp__` with draws instead of sampler_diagnostics so it's included in the `$summary()` output. In terms of names, I like `draws()` and `sampler_diagnostics()`. You?

rok-cesnovar:
I like them too. I saw that you mentioned that in the issue, thats why I wrote it here.

jgabry:
Ok cool, I'm going to push some documentation updates and then merge this!

PR #142, opened 2020-03-08
Compile in tmp folder

### LINKED ISSUE #75: Paths with spaces in them
Compilation method doesn't handle file paths with spaces in them:

```
[14:18:31] hhau:Desktop $ tree
.
├── dir\ withaspace
│   └── ex-three-gaussians-melded-joint.stan
└── ex-three-gaussians-melded-joint.stan

1 directory, 2 files
[14:18:32] hhau:Desktop $ Rscript -e "invisible(cmdstanr::cmdstan_model('ex-three-gaussians-melded-joint.stan'))"               
Compiling Stan program...
[14:18:45] hhau:Desktop $ Rscript -e "invisible(cmdstanr::cmdstan_model('dir withaspace/ex-three-gaussians-melded-joint.stan'))"     
Compiling Stan program...
Error in processx::run(command = make_cmd(), args = exe, wd = cmdstan_path(),  : 
  System command error
Calls: <Anonymous> ... <Anonymous> -> <Anonymous> -> <Anonymous> -> <Anonymous> -> throw
Execution halted
[14:18:50] hhau:Desktop $ tree
.
├── dir\ withaspace
│   └── ex-three-gaussians-melded-joint.stan
├── ex-three-gaussians-melded-joint
├── ex-three-gaussians-melded-joint.hpp
├── ex-three-gaussians-melded-joint.o
└── ex-three-gaussians-melded-joint.stan

1 directory, 5 files
```

I can't quite figure out the structure of the `CmdStanModel` object to find the culprit - I think that calls to `system` need appropriate extra quotation a la: https://github.com/stan-dev/rstantools/pull/44/commits/504f0d691d50776a614b89fd3270e83fd69b7db5

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Closes #75 

Changes how a model is compiled to a similar approach cmdstanpy uses: 
1. check if stan file was modified
2. move the stan file to tmpfile and compile
3. move only the executable back to the stan file folder

Other changes: 
- adds a force_recompile flag
- this now supports paths and model names with spaces

#### Copyright and Licensing

Rok Češnovar


jgabry:
Thanks. The code looks good, but for some reason I get an error from the test in `test-model-compile.R` when it does  `mod$compile(threads = TRUE)`. I get the same error if I try `threads=TRUE` outside of a test too. Any ideas? 

```
mod$compile(threads = TRUE)
Compiling Stan program...
A change in the compiler flags was found. Forcing recompilation.

ld: illegal thread local variable reference to regular symbol __ZN4stan4math22AutodiffStackSingletonINS0_4variENS0_15chainable_allocEE9instance_E for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [/var/folders/h6/14xy_35x4wd2tz542dn0qhtc0000gn/T/RtmpJMGi8h/model-ccf14fc03a52] Error 1
Error in processx::run(command = make_cmd(), args = c(tmp_exe, include_paths,  : 
  System command error
Called from: throw(new_process_error(res, call = sys.call(), echo = echo, 
    stderr_to_stdout, res$status))
```

rok-cesnovar:
Good catch. I could reproduce with clang on linux, g++ was not problematic with this, which is always fun.

The problem is that we need to also recompile "cmdstan/stan/src/stan/model/model_header.hpp.gch" in this case. I think this file is only used with clang.

rok-cesnovar:
Not really relevant, but do you experience this same issue with master cmdstanr? I would be surprised if you dont.

mitzimorris:
> Good catch. I could reproduce with clang on linux, g++ was not problematic with this, which is always fun.
> 
> The problem is that we need to also recompile "cmdstan/stan/src/stan/model/model_header.hpp.gch" and "cmdstan/stan/src/stan/model/model_header.d" in this case. I think the former is not created with g++, not sure why the latter is not a problem.

correct, model_header.hpp.gch is only created for clang++

R/model.R
```diff
@@ -239,18 +264,22 @@ compile_method <- function(quiet = TRUE,
...
     wd = cmdstan_path(),
     echo_cmd = !quiet,
     echo = !quiet,
     spinner = quiet,
     stderr_line_callback = function(x,p) { if(quiet) message(x) },
     error_on_status = TRUE
   )
-
+  
+  file.copy(tmp_exe, exe)
```
bbbales2: I think this should be

```
file.copy(tmp_exe, exe, overwrite = TRUE)
```

otherwise the outdated binary never gets overwritten.

R/model.R
```diff
@@ -214,17 +215,41 @@ compile_method <- function(quiet = TRUE,
...
+                           compiler_flags = NULL,
+                           force_recompile = FALSE) {
   make_local_changed <- set_make_local(threads,
                                        opencl,
                                        opencl_platform_id,
                                        opencl_device_id,
                                        compiler_flags)
+  exe <- cmdstan_ext(strip_ext(self$stan_file()))
+  recompile <- make_local_changed || force_recompile ||
+               !file.exists(exe) || (file.mtime(exe) < file.mtime(self$stan_file()))
```
bbbales2:
```
(file.mtime(exe) < file.mtime(self$stan_file()))
```

file.mtime(exe) is an NA if exe doesn't exist. So NAs work their way into the logic. I don't know how R treats NAs in logicals.

Maybe just break this out into more explicit ifs instead of ORing to avoid NAs. Unless of course if(NA) is defined and I'm just not familiar with R and this is all okay.

rok-cesnovar: So the check here is 
```R
!file.exists(exe) || (file.mtime(exe) < file.mtime(self$stan_file()))
```
If the exe file doesnt exist this is TRUE || NA which is TRUE. If it does exist, then the second clause will be either TRUE or FALSE and its fine.

```R
> TRUE || NA
[1] TRUE
> FALSE || NA
[1] NA

```

rok-cesnovar: I will make this more elaborate. All this "cleverness" hurts more then it gains.

R/model.R
rok-cesnovar: Good call.

bbbales2:
I think there's a missing parenthesis still.

Also can we add a test for this? Like:
```
test_that("compile() method overwrites binaries", {
  skip_on_cran()
  mod$compile(quiet = TRUE)
  old_time = file.mtime(mod$exe) // I'm just making up the mod$exe variable
  mod$compile(quiet = TRUE, force_recompile = TRUE)
  new_time = file.mtime(mod$exe)
  assert_less_than(old_time, new_time) // Also making up this assert syntax.
})
```

bbbales2:
I approve of your changes. Check out the doc changes I made and merge if you approve those.

rok-cesnovar:
Thanks! I will wait for the tests and merge.

Issue #136, opened 2020-03-03
Consistent argument names - "num_cores" should be "cores"; "num_chains" should be "chains"


mitzimorris:
**Is your feature request related to a problem? Please describe.**

The names "cores" and "chains" are used in both RStan and CmdStanPy.  "num_chains" might be more correct, but it's confusing to users coming from RStan.

**Describe the solution you'd like**
Please rename to just "chains" and "cores"

**Describe alternatives you've considered**
Was going to make CmdStanPy match this, but then it was pointed out that RStan and PyStan also use "chains" and "cores".

jgabry:
I thought the point was to be consistent with CmdStan rather than RStan and PyStan? I know `num_chains` and `num_cores` don't exist for CmdStan but those names are consistent with CmdStan's `num_warmup` and `num_samples`.

jgabry:
This is related to https://github.com/stan-dev/cmdstanpy/issues/221

jgabry:
@mitzimorris @rok-cesnovar @bbbales2 I think if we use `num_warmup` and `num_samples` then we should use `num_chains` and `num_cores`. If we want to go with `cores` and `chains` I think we should rethink the rest of the names to make them consistent with those. 

I'm ok with @bob-carpenter's preference (in stan-dev/cmdstanpy#221) of picking names that we want to use for CmdStan3, but in that case we should go over _all_ the names again with that goal in mind. 

What do you all think?

mitzimorris:
sounds good!
I would love to get traction on CmdStan3 - there's a proposal in the design-docs repo - discuss there?
https://github.com/stan-dev/design-docs/pull/15

rok-cesnovar:
I honestly dont have a strong preference on names, other than things are consistent in at least the cmdstan* world. Redesigning names with cmdstan3 sounds like a good plan to me.

mitzimorris:
spreadsheet comparing sampler controls available from RStan/PyStan and CmdStan -
https://docs.google.com/spreadsheets/d/1lU6vcY3GF_ftlFdCe2kDhr5FC4YJPWauC5bp2D9zz74/edit?usp=sharing

cells in yellow flag names that we'd like to change.
cells in green flag names we don't want to expose.

how does this work for CmdStanR?
alternative to names "num_samples", "num_warmup":  "iter_warmup", "iter_sampling" - more RStan-like?

bob-carpenter:
This is really helpful.  

[...]

Issue #133, opened 2020-03-03
Explore options for faster csv reading than utils::read.csv


jgabry:
At some point we should investigate if we can get a speedup by using other methods of reading in csv output. Currently we use the standard R function `utils::read.csv()`.

mcol:
`data.table::fread` is massively faster for large files, and has an option to produce a `data.frame` if that's what you are looking for.

rok-cesnovar:
Thanks! I will take a look at that.

rok-cesnovar:
fread looks fast, but it doesn't support ignoring lines with "#". Most comments suggest using grep with fread to do that, which isn't really portable. They have an open issue to add this but its been longstanding.

[readr](https://readr.tidyverse.org/news/index.html ) is another option I looked into 

[vroom](https://www.tidyverse.org/blog/2020/01/vroom-1-1-0/) looks the most promising.
It has some issues with comments after the column names are specified:
```
lp__,accept_stat__,stepsize__,treedepth__,n_leapfrog__,divergent__,energy__,mu,sigma
# Adaptation terminated
# Step size = 0.672434
# Diagonal elements of inverse mass matrix:
# 0.909635, 0.066384
```
but that can be fixed in R afterwards:

For now I tried reading a csv with 41 params (sampler  + model params) with 90k iterations:
- utils::read.csv 2.46s
- readr 1.3s
- vroom 0.08s (they use lazy reading so not sure this is a totally fair comparison yet, but looks amazing)

```R
N <- 10
utils <- c()
for(i in 1:N) {
  t1 <- Sys.time()
  tmp1 <- utils::read.csv(csv_file, comment.char = "#", header = TRUE)
  t2 <- Sys.time()
  utils <- c(utils, (t2-t1))
}
print(paste0("utils::read.csv ", mean(utils)))

readr_times <- c()
for(i in 1:N) {
  t1 <- Sys.time()
  tmp2 <- readr::read_csv(csv_file, comment = "# ", trim_ws = TRUE)
  t2 <- Sys.time()
  readr_times <- c(readr_times, (t2-t1))
}
print(paste0("readr ", mean(readr_times)))

vroom_times <- c()
for(i in 1:N) {
  t1 <- Sys.time()
  tmp3 <- vroom::vroom(csv_file, comment = "# ", delim = ',', trim_ws = TRUE, col_types = c(lp__ = "d"))
  # fix the table
  tmp3 <- tmp3[!is.na(tmp3$lp__),]
[...]

Issue #119, opened 2020-02-06
Only save selected parameters in output


andrjohns:
Not sure whether this should be a ```cmdstan``` or ```cmdstanr``` feature request, but it would be very useful if it were possible to only write the parameter estimates for selected parameters to the output (like RStan does with the ```pars``` argument).

I'm working with models where I'm only interested in particular post-processed parameters (sign-corrections in latent variable models), but when I use ```cmdstanr``` I end up with very large output files, which can be a bit cumbersome.

rok-cesnovar:
This can definitely be done in cmdstanr, though that only reduces the size of the samples R object, the output file will stay as big. For now, you can trim the returned samples object manually in R, though doing that through cmdstanr would definitely be nicer.

In Cmdstan, the related issue is https://github.com/stan-dev/cmdstan/issues/553. https://github.com/stan-dev/cmdstan/issues/619 is also somewhat related.

mcol:
I think that having a way to reduce the size of the R object would aready be a very helpful feature. What function should one look at to implement this?

rok-cesnovar:
The way on how we read in the csv data is going to significantly change after https://github.com/stan-dev/cmdstanr/pull/95

The main function is `read_sample_csv`. Solving this would probably require adding some additional args to this function.

rok-cesnovar:
With the vroom package we can select/drop parameters by name or even use selectin helpers (like endswith). Do we want all of them or just the basic select. 

Also, any ideas on the argument name for this?

jgabry:
Is this going to be an argument to the model fitting methods (`$sample()`, `$variational()`, etc.)?

jgabry:
And I think those helpers would probably be useful.

Issue #20, opened 2019-10-09
Check for (and install?) required tool chain 


jgabry:
Should check what CmdStanPy does: https://github.com/stan-dev/cmdstanpy/blob/master/cmdstanpy/install_cxx_toolchain.py

PR #144, opened 2020-03-16
Generalize compiling

### LINKED ISSUE #120: Enable running pre-built models
This is a question to see how other see this feature. This is a use-case that would help maintaining some of the projects I am involved with, but I do not want to force it in cmdstanr if others dont see it as useful.

**Description:** 

Enable the users to run pre-build models using cmdstanr. This is useful for when you want to ship a pre-built Stan model as an executable to the end-user (for their OS) and enable them to easily use it via R. This is mainly so that they dont have to deal with any installation of compilers and toolchains if they are only interested in one specific model. Its probably not a really common use case.

Proposed use would be: 
`mod <- cmdstan_model(exe = "path/to/pre-built/exe/file)`
and the rest is then as usual.

### LINKED ISSUE #121: simplify/generalize model compilation options
**Is your feature request related to a problem? Please describe.**

CmdStanModel constructor and `compile` method take a series of named arguments for the CmdStan makefile variables, currently:

- arg `threads` for Makefile var `STAN_THREADS`
- arg `opencl` for stanc3 compiler arg `--use-opencl`
- args `opencl_device_id.` for Makefile vars `OPENCL_DEVICE_ID`
- ditto `opencl_platform_id`

We need to minimize dependencies between CmdStanX wrappers and the Stanc3 compiler and CmdStan makefiles.

**Describe the solution you'd like**

Implement a more general organization of user-facing compiler options instead of having many specific named options which will:

- Differentiate between arguments for options for `bin/stanc` vs. options to calls to `make` at the user-facing API.
- Provide getter/setter methods for compiler and makefile options.  These options should be  object-specific, therefore, don't rewrite user's `make/local` file.
- Check `bin/stanc` compiler options, because compiler throws an error on unrecognized options
- Pass `make` options into compiler unchecked
- disallow Stanc3 compiler debug flags - debugging the compiler is outside of the scope of the CmdStanX wrappers.

**Additional context**

We need an extensible solution that provides similar user-facing API for CmdStanPy and CmdStanR and future CmdStanX wrappers.  If the Stanc3 compiler introduces new arguments, then the CmdStanX wrappers need to change accordingly but otherwise, the CmdStanX wrappers should just work with new CmdStan releases.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #121 and also fixes #120

Adds `cpp_options`, `stanc_options` and `model_name` to `$compile()`

#### Copyright and Licensing

Rok Češnovar


rok-cesnovar:
This is now ready for review.

rok-cesnovar:
I refactored this again and I think this should be the way to go. Seemless for the user, not hacky and the least intrusive.

Cmdstanr helps keep "flagged" version of main.o: 
```
main_noflags.o
main_threads.o
main_opencl.o
...
```
based on supplied cpp_options, it moves the correct main_*.o to main.o. Same for model_header.hpp.gch if used (used with clang).

I made a test gist: https://gist.github.com/rok-cesnovar/7a5a11b89cb1759fd65cfa0b6cbe1848
It should produce something like:
```
[1] "First without flags"
Compiling Stan program...
Time difference of 51.15341 secs
[1] "second without flags"
Compiling Stan program...
Time difference of 8.325357 secs
[1] "first with threads"
Compiling Stan program...
Time difference of 48.44692 secs
[1] "second with threads"
Compiling Stan program...
Time difference of 7.581914 secs
[1] "third without flags"
Compiling Stan program...
Time difference of 7.606306 secs
[1] "third with threads"
Compiling Stan program...
Time difference of 8.220567 secs
```

jgabry:
Thanks @rok-cesnovar! Looks like some tests are failing but otherwise looks OK to me.

jgabry:
Actually I just noticed a few very minor things: 

* When compiling we're always printing the same message "Compiling Stan program..." regardless of whether we're also compiling `main.o`.  Can we add a message about also compiling `main.o` when that happens? That way the user isn't confused as to why it's taking longer. 

* The doc says that `include_paths`, `cpp_options`, and `stanc_options` are "character vectors", but the default for `cpp_options` and `stanc_options` is an empty `list()` instead of an empty character vector. Should these be specified as lists or vectors? Also the default of `include_paths` is `NULL`. Should these all have the same default? 

* Also on the topic of the new `cpp_options` and `stanc_options` arguments, can you provide a few examples demonstrating how to use those arguments?  These could go in the `@examples` section inside the `'dontrun{}` section that's already there.

rok-cesnovar:
> * The doc says that `include_paths`, `cpp_options`, and `stanc_options` are "character vectors"

Ah yeah, copy paster error, options should be lists, while the paths should be character vectors.

Will add examples, was hoping to also make a vignette with the tutorial from @bbbales2 on the use of reduce_sum. But that comes later.

jgabry:
> was hoping to also make a vignette with the tutorial from @bbbales2 on the use of reduce_sum. But that comes later.

That would be great! 

Also, does this PR break anything in that reduce_sum tutorial? I think @bbbales2 used the `threads = TRUE` argument, which is gone now right?

bbbales2:
> I think @bbbales2 used the threads = TRUE argument, which is gone now right?

I take it when this gets merged I'll need to go update that?

rok-cesnovar:
Yes, at that point we will do that with:

`mod <- cmdstan_model(file, cpp_options = list(stan_threads = TRUE))`

rok-cesnovar:
I can also make a PR to fix that once we merge this.

bbbales2:
Cool beans. I made an issue to try to help myself remember: https://github.com/stan-dev/example-models/issues/175

jgabry:
Thanks!

rok-cesnovar:
I think this should be good now.

The vroom PR coming tomorrow hopefully will address #133 and #119 and then we  only need to address #7 and #136 for the beta. And maybe #20

jgabry:
Thanks @rok-cesnovar! travis was failing due to a doc issue (all tests were passing), so I fixed that and I'll merge this now.

rok-cesnovar:
Thanks!

PR #164, opened 2020-05-12
Faster csv reading

### LINKED ISSUE #133: Explore options for faster csv reading than utils::read.csv

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #133 

With 512 parameters and 2000 samples the reading times goes from 8s with what we have on master to 4 seconds.

I will address only reading in selected paramters in a separate PR.

#### Copyright and Licensing

Rok Češnovar


R/read_csv.R
```diff
@@ -204,8 +204,14 @@ read_sample_csv <- function(output_files) {
...
     # read sampling data
-    draws <- utils::read.csv(output_file, header = TRUE, comment.char = "#")
-    if (nrow(draws) > 0) {
+    suppressWarnings(      
+      draws <- vroom::vroom(output_file, comment = "# ", delim = ',', trim_ws = TRUE, col_types = c(lp__ = "d"), num_threads = cores)
+    )
+    if(ncol(draws) == 0) {
+      stop("The supplied csv file does not contain any sampling data!")
+    }
+    draws <- draws[!is.na(draws$lp__),]
```
rok-cesnovar, line 213: This is needed because what vroom reads in is the following:
```
     lp__ accept_stat__ stepsize__ treedepth__ n_leapfrog__ divergent__ energy__   theta
    <dbl>         <dbl>      <dbl>       <dbl>        <dbl>       <dbl>    <dbl>   <dbl>
 1  NA           NA         NA              NA           NA          NA    NA    NA     
 2  NA           NA         NA              NA           NA          NA    NA    NA     
 3  NA           NA         NA              NA           NA          NA    NA    NA     
 4  NA           NA         NA              NA           NA          NA    NA    NA     
 5  -6.87         0.955      0.931           1            3           0     7.08  0.192 
 6  -7.96         0.729      0.931           2            3           0     9.23  0.0950
 7  -7.85         1          0.931           1            1           0     8.07  0.100 
 8 -11.1          0.746      0.931           1            3           0    12.0   0.0272
 9  -8.30         1          0.931           2            3           0    11.1   0.0811
10  -6.84         0.967      0.931           1            3           0     8.53  0.199 
# … with 999 more rows
# A tibble: 1,009 x 8  
```
Those above are there because of our csv format. I think we could get away with always skipping first 4 rows, but this is safer.

rok-cesnovar:
Ignore the review request, need to fix the tests first.

jgabry:
Ok no problem. Will wait for now.

jgabry:
Any idea why a few tests are failing on appveyor but not travis?

jgabry:
fyi, I just merged the PR with the name changes and fixed a merge conflict

rok-cesnovar:
> Any idea why a few tests are failing on appveyor but not travis?

Yeah, vroom reads files lazily, the file is initially memory mapped and windows with RTools 3.5 does not wont to delete it if it hasnt been closed. Works just fine with RTools 4.0 toolchain. Investigating :)

rok-cesnovar:
vroom also lets us read multiple files at once so I will also add that here.

rok-cesnovar:
@jgabry This is now ready for review. I decided to turn off the lazy reading as it creates issues if you have not read the file and want to remove them (what save_output_files() does). So there is room for improvement.

rok-cesnovar:
If you havent started looking at this, then please wait and I will make another push soon. Just need to add tests.

I added the arguments for selecting the parameters/sampling_diagnostics you wish to read from the CSV file. On $sample(), only lp__, treedepth__ and divergent__ are read (lp__ is read because it makes it easier for vroom to parse what is a valid row). This makes the validation of samples really fast compared to what we have on master as validation currently read the entire CSV file.

Validation timing on this PR vs master is shown below. I dont expect anyone turning off validation based on these numbers. We can keep the arg if someone wants to just not have the divergence warnings.

| num of samples \ branch - num of parameters | PR - 18 | PR - 1026 | master - 18 | master - 1026 |
|---------------------------------------------|---------|-----------|-------------|---------------|
| 500                                         | 0.10    | 0.20      | 0.11        | 2.57          |
| 1000                                        | 0.15    | 0.24      | 0.15        | 4.5           |
| 2000                                        | 0.18    | 0.30      | 0.22        | 7.6           |
| 4000                                        | 0.24    | 0.46      | 0.41        | 15.5          |
| 8000                                        | 0.35    | 0.66      | 0.81        | 32.1          |

These are also example timings if you are only interested in 2 parameters of X compared to what we had on master where you had to read the entire CSV file.

This also makes the following

```R
a <- read_sample_csv(fit$output_files(), col_select = c("lp__", "treedepth__", "divergent__"), cores = 4)
check_divergences(a)
check_sampler_transitions_treedepth(a)
```
much faster than `$cmdstan_diagnose()` if you are only interested in divergences/treedepth. On 1k parameters with 2000 samples its 0.25s vs 8s.

`cmdstan_diagnose` does a few more stuff so I am not saying get rid of it.
But we can also think about exposing `check_divergences` and `check_sampler_transitions_treedepth`.

rok-cesnovar:
I think should be quite a popular replacement for rstan::read_stan_csv() for people running barebones cmdstan on clusters that want to read results.

jgabry:
> If you havent started looking at this, then please wait and I will make another push soon. Just need to add tests.

Ok no problem, I'll wait. 

> I dont expect anyone turning off validation based on these numbers.

Yeah, the timings look great! Thanks for making that table. 

> But we can also think about exposing `check_divergences` and `check_sampler_transitions_treedepth`

Yeah good idea. (Maybe we can rename `check_sampler_transitions_treedepth` to just `check_treedepth`?)

> On $sample(), only lp__, treedepth__ and divergent__ are read

That makes sense. What's happens if the user later wants some of the other sampler diagnostics? Maybe the first time they call the `$sampler_diagnostics()` method it should read in all of them?

rok-cesnovar:
Yes. The way I envisioned is $sampler_diagnostics() reads in all of them? While in draws() you could filter them.
$draws(parameters = c("mu", "theta")) is how I have it working now.

jgabry:
Ok great! 

I don't know if we should call the argument `parameters` (it's a fine name but rstan uses `pars` so we could use that, or posterior uses `variable` since it also can refer to generated quantities, so we should think about the name a bit more but don't worry about that now)

rok-cesnovar:
This now finally ready for review.

jgabry:
Awesome thanks! Will take a look this weekend, or Monday at the latest.

rok-cesnovar:
Sure, no rush. Whenever you have time.

jgabry:
I started going through this and making a few edits, and I also have a few questions. A bit busy at the moment so will have to finish later tonight or tomorrow.

R/fit.R
```diff
@@ -334,27 +340,37 @@ CmdStanMCMC <- R6::R6Class(
       }
     },
 
-    draws = function(inc_warmup = FALSE) {
-      if (is.null(private$draws_)) {
-        private$read_csv_()
+    draws = function(inc_warmup = FALSE, parameters = NULL) {
```
jgabry, line 343: What do you think we should do with the name `parameters`? rstan, bayesplot, and other packages use `pars`. posterior uses `variables`. I think we should go with `pars` or `variables` instead of `parameters` but I'm not sure which.

jgabry: Also we will probably eventually want to allow turning on matching via regex or some other mechanism so that people can select parameters more easily. Right now if you have `vector[3] theta` in your Stan program you can't do `parameters="theta"`, you have to do `parameters=list("theta[1]", "theta[2]", "theta[3]")`. This doesn't necessarily have to be included in this PR though.

R/read_csv.R
```diff
@@ -159,32 +161,62 @@ read_sample_info_csv <- function(csv_file) {
...
 #'
 #' @export
-#' @param output_files Paths to the CSV files to read.
+#' @param files Paths to the CSV files to read.
+#' @param parameters (list) When `NULL`, the function will return
+#' draws for all model parameters. When supplied a list, the function
+#' will return  draws from the listed model parameters.
+#' @param sampler_diagnostics (list) When `NULL`, the function will return
+#' draws for all sampler diagnostics . When supplied a list, the function
+#' will return  draws from the listed sampler diagnostics .
```
jgabry, line 173: Is there a reason that `parameters` and `sampler_diagnostics` are lists instead of character vectors (i.e., why `list("x", "y", "z")` and not `c("x", y", "z")`? If there's not a particular reason, then I think character vectors would be more standard in R (and I think most of the code you already have dealing with the lists internally will still work with character vectors).

R/read_csv.R
```diff
@@ -196,13 +228,13 @@ read_sample_csv <- function(output_files) {
...
     } else {
       csv_file_info <- read_sample_info_csv(output_file)
-      # check if sampling info matches
       check <- check_sampling_csv_info_matches(sampling_info, csv_file_info)
       if (!is.null(check$error)) {
         stop(check$error)
       }
       not_matching <- c(not_matching, check$not_matching)
       sampling_info$id <- c(sampling_info$id, csv_file_info$id)
+      sampling_info$seed <- c(sampling_info$seed, csv_file_info$seed)
```
jgabry, line 237: I added this since the seed is different for each chain also

R/read_csv.R
```diff
@@ -212,58 +244,94 @@ read_sample_csv <- function(output_files) {
...
-    # read sampling data
-    draws <- utils::read.csv(output_file, header = TRUE, comment.char = "#")
+    repaired_model_params <- repair_variable_names(sampling_info$model_params)
+    repaired_sampler_diagnostics <- repair_variable_names(sampling_info$sampler_diagnostics)
+    if (!is.null(parameters)) {
+      for (p in parameters) {
+        if (!(p %in% repaired_model_params)) {
+          stop("The specified parameter ", p, " was not found in the sampling output!")
+        }
+      }
```
jgabry, line 254: I think it would be better to check these all at once instead of in a loop, so that if the user has multiple parameters that are not found they don't have to keep trying and finding out another one is wrong. 

What do you think? Something along these lines could work:

```r
not_found <- parameters[!parameters %in% repaired_model_params]
if (length(not_found)) {
  stop("Can't find parameter(s) ", paste(not_found, collapse = ", "), " in the sampling output!")
}
```

R/read_csv.R
```diff
@@ -212,58 +244,94 @@ read_sample_csv <- function(output_files) {
...
+      for (p in sampler_diagnostics) {
+        if (!(p %in% repaired_sampler_diagnostics)) {
+          if (p == "lp__") {
+            stop("lp__ is a part of model parameters not sampler diagnostics.")
+          } else {
+            stop("The specified sampler diagnostic ", p, " was not found in the sampling output!")
+          }
+        }
+      }
+    }
```
jgabry, line 266: Same comment as above about catching these all at once

R/read_csv.R
```diff
@@ -212,58 +244,94 @@ read_sample_csv <- function(output_files) {
...
+        col_select <- c(col_select, sampling_info$model_params[sampling_info$model_params!="lp__"])
+      } else {
+        col_select <- c(col_select, unrepair_variable_names(unlist(parameters, use.names=FALSE)))
+      }
+      selected_sampler_diagnostics <- sampling_info$sampler_diagnostics[sampling_info$sampler_diagnostics %in% col_select]
+      selected_model_params <- sampling_info$model_params[sampling_info$model_params %in% col_select]
+    }
+    if (length(sampling_info$sampler_diagnostics) == 0 && length(sampling_info$model_params) == 0) {
+      stop("The supplied csv file does not contain any parameter names or data!")
+    }
```
jgabry, line 286: I might be missing something, but I think we can throw this error much earlier right after creating `sampling_info`, right?

R/read_csv.R
rok-cesnovar: Thanks!

R/read_csv.R
rok-cesnovar: I wanted to support
- NULL which returns all parameters
- list() which returns no parameters
- list("a", "b"...) 

NULL or c() is the same. We could use NULL, NA and c(...) if that would be better.

jgabry: Oh I see. `NA` could be ok. Another option is an empty string, for example, 

```r
sampler_diagnostics = "" # none
sampler_diagnostics = c("divergent__", "treedepth__") # selection
sampler_diagnostics = NULL # all
```

Also, pinging @paul-buerkner since I trust his take on what is sensible from an R perspective. Paul, any opinion on this?

R/fit.R
jgabry: @paul-buerkner Any opinion on `pars` vs `variables` for the argument name?

R/read_csv.R
rok-cesnovar: Yeah, moved it to read_sample_info_csv to fail sooner.

R/read_csv.R
rok-cesnovar: Fixed.

R/read_csv.R
rok-cesnovar: Fixed.

rok-cesnovar:
I fixed the comments that were straightforward, will wait for the decision on names and the parameters "API".

R/fit.R
rok-cesnovar: Lets go with pars if anyone doesnt ming. That is what the rstan users are accustomed to. I dont really like it, params is at least to me a more standard abbreviation. But probably best to avoid confusion.

R/read_csv.R
rok-cesnovar: I like empty string. No parameter can match that. Will do that.

R/fit.R
mitzimorris: I really would prefer `variables` to pars.  Just writing up my reasons why as part of CmdStan3 design doc now - because I don't know where else to discuss this.

jgabry: > I really would prefer variables to pars. Just writing up my reasons why as part of CmdStan3 design doc now - because I don't know where else to discuss this.

@mitzimorris Sorry I missed this comment. Are you still OK with what we talked about in the meeting today? That is, going with `pars` for now for consistency with with rstan/pystan/bayesplot, but then potentially deprecating and changing it at a later date (e.g. with CmdStan3)?

rok-cesnovar:
I am a bit stuck here, cant reproduce the Travis test errors locally on Ubuntu. Will spin an AWS instance tomorrow to see what is hapenning.

jgabry:
@rok-cesnovar Just tried the tests on my mac too and they were failing, but I may have found the issue. It looks like there were two executables in the `tests/testthat/resources/stan` directory that shouldn't have been there. I just pushed a commit removing them. 

My best guess is that you ran a few of the tests manually while debugging and the resulting executables didn't get removed automatically because the testthat teardown doesn't run if you manually run individual tests. That's happened to me before too.

rok-cesnovar:
Thanks! Ready for another look

jgabry:
Should all the tests in `tests/testthat/test-csv.R` still be commented out or can we turn them on again now?

rok-cesnovar:
Ah no, forgot about those. Will uncomment ASAP.

jgabry:
Is the `cores` argument to `read_sample_csv()` used anywhere? I don't see it getting passed to vroom or being used inside `read_sample_csv()` anywhere.

rok-cesnovar:
Vroom has a cores argument but you dont gain that much and the default setting gives good enough performance so I didnt use it in the end. Forgot to cleanup, sorry.

jgabry:
No worries, I'l remove cores now while Im reviewing

jgabry:
Also, I think you were wise to prefer variables to pars. I just decided to go for it and switch to variables (just pushed). I think that's just more sensible and we shouldn't worry that rstan uses pars. Sorry for taking so long to decide that!!

jgabry:
> Also, I think you were wise to prefer variables to pars. I just decided to go for it and switch to variables (just pushed). I think that's just more sensible and we shouldn't worry that rstan uses pars. Sorry for taking so long to decide that!!

I assume you're ok with that? If not we can revert my recent commit

rok-cesnovar:
Yes offcourse I am fine with that!

jgabry:
Ok I think I found one issue we need to resolve, but otherwise this is great! 

The one thing we overlooked is that only the first one of these works: 

```r
# this works
read_sample_csv(fit$output_files(), variables = "theta")

# this errors because it doesn't know about theta, only theta[1], theta[2], etc. 
fit$draws(variables = "theta")
```

In other words, variable selection using the name of the vector "theta" (and not the individual elements) only works when calling `read_sample_csv()` directly, and not when using the new `variables` (formerly `pars`) argument to `$draws()`.

jgabry:
The problem I described in my previous comment is due to the fact that we are handling variable selection in read_sample_csv() whereas we are outsourcing it to the posterior package for the $draws() method. 

posterior has subset_draws() which has a regex argument, which maybe could be used to get what we need here, but it’s not ideal.

rok-cesnovar:
> In other words, variable selection using the name of the vector "theta" (and not the individual elements) only works when calling `read_sample_csv()` directly, and not when using the new `variables` (formerly `pars`) argument to `$draws()`.

Fixed. 

I also found a bug with reading `lp__` after a "filtered" draws() call:

```R
a <- fit$draws(variables = c("eta"))
b <- fit$draws()
```
I also fixed this and added additional tests for this bug.

Final timing report:

|                                 | 19        | 643       | 1283       | 1923       | 2563       | 3203       |
|---------------------------------|-----------|-----------|------------|------------|------------|------------|
| develop                         | 0.3489878 | 5.9398425 | 10.4866965 | 16.3929579 | 21.5311923 | 27.2760901 |
| PR                              | 0.2396772 | 1.7050161 | 3.2770655  | 4.4713771  | 6.0561974  | 6.7868321  |
| PR - read 50% of parameters     | 0.1603785 | 1.1780379 | 2.0339272  | 2.9903495  | 3.5180790  | 4.1634433  |
| PR - read 2 columns (validation | 0.1399992 | 0.3635323 | 0.6058638  | 0.8312213  | 1.1903284  | 1.3976476  |
| PR - read 1                     | 0.1341414 | 0.3551099 | 0.6020284  | 0.8367260  | 1.2255943  | 1.3311992  |

jgabry:
A few things I noticed:

* Probable bug: if all draws have already been read in then no error is thrown if a variable is selected that doesn't actually exist:

```r
fit <- mod$sample(data = stan_data)
d <- fit$draws(variables = "ABCD") # correct error

# first read in all 
d <- fit$draws()
d <- fit$draws(variables = "ABCD") # now no error
```

* The `draws()` method for CmdStanVB objects doesn't have a `variables` argument. That wasn't part of this PR, so that's ok, but it seemed weird to have `draws()` work differently for MCMC and VB. So for now, I added the `variables` argument and implemented it using `posterior::subset_draws()` (see commit e5a24d3), which @paul-buerkner just updated to allow selecting non-scalar parameters. Eventually it would be good to only read in the minimum set of variables needed (like this PR does for MCMC), but that doesn't have to be part of this PR. Anyway, @rok-cesnovar no immediate action needed from you here, just wanted to check that you're ok with this temporary solution for VB?

jgabry:
For the bug I mentioned in my previous comment, would using the `posterior::subset_draws()` help at all now that it can select non-scalar parameters by name?

rok-cesnovar:
> * just wanted to check that you're ok with this temporary solution for VB?

Yes, definitely ok with the solution.

> would using the `posterior::subset_draws()` help at all now that it can select non-scalar parameters by name

Yes, it would, but there was a simpler fix (forgot to add a check). I think we should offload as much to posterior as we can, but I would wait until we have the R package repo, versioned dependencies with install_github() aren't super nice, at least in my experience.

jgabry:
> Yes, it would, but there was a simpler fix (forgot to add a check). I think we should offload as much to posterior as we can, but I would wait until we have the R package repo, versioned dependencies with install_github() aren't super nice, at least in my experience.

Ok yeah, that makes sense. 

On another note, I just pushed a commit changing "parameter" to "variable" in some of the error messages (I assume that's ok). Will wait for travis/appveyor to finish and then merge!

Issue #117, opened 2020-02-03
MPI execution


rok-cesnovar:
Another one of @avehtari great idea, supporting MPI execution. Basically, support the equivalent of

`mpiexec -n 4 --tag-output ./radon sample data file=radon.data.R`

```
modelc1 = cmdstan_model("bernoulli_logit_glm_rhs.stan", quiet = FALSE, mpi = TRUE)
fit1 = modelc1$sample(data = data, mpi = TRUE, n = 4 )`
```

The API is very much up for debate.

rok-cesnovar:
Also see https://discourse.mc-stan.org/t/cross-chain-warmup-adaptation-using-mpi/12912/15

wds15:
Maybe it's worth to think for a moment of how we specify MPI processes and threads to be used. I don't really have a good idea atm.

wds15:
One more thought: Traditionally MPI is used in clusters and you do send your program into a queue with a little script which fires off everything. These scripts typically are very simple shell scripts. From that angle it would not make a lot of sense to include MPI into CmdStanR as it does not really make sense to call out to R just to start the MPI program.

Should we recommend to users  to run MPI on their desktops, then it does again make sense, of course - but for that purpose we have threads. So for me a R user will only ever use the CPUs on his own machine in which case you do not need MPI convenience features in CmdStanR.

This obviously changes if we are going to have MPI only features.

rok-cesnovar:
I am seeing this primarily for: 
-  developer and early-adopter users oriented MPI testing (new MPI features that we need to test locally); basically what the reason leading to this issue
- users that want to test if everything is running with MPI as they thought it should before sending it to the cluster 

I doubt we will have MPI-only features but that could be a use case also.

wds15:
As a dev oriented feature it does make sense, indeed.

yizhang-yiz:
I can confirm the following works:
```r
> library("cmdstanr")
> set_cmdstan_path("/Users/yiz/Work/temp/cmdstan")
cmdstan_make_local(cpp_options=list("CXX"="mpicxx", "CC"="mpicxx", "STAN_MPI"="1","STANC2"="true","TBB_CXX_TYPE"="clang"))
[1] "STAN_MPI=1"         "STANC2=true"        "TBB_CXX_TYPE=clang"
[4] "CXX=mpicxx"         "CC=mpicxx"          "STAN_MPI=1"        
[...]

PR #185, opened 2020-05-27
Improve threading interface

### LINKED ISSUE #163: Interface for number of threads
The current interface to setting the number of within chain threads is

```
set_num_threads(4)
```

Should we maybe move it to `$sample(chain_threads=4)`? Though the cores and threads together might be confusing?

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #163 by: 
- deprecating set_num_threads
- adding the threads_per_chain argument to $sample()

One thing I am not really sure what to do is if threads_per_chain > cores.
#### Copyright and Licensing


R/utils.R
```diff
@@ -336,23 +337,23 @@ prepare_precompiled <- function(cpp_options = list(), quiet = FALSE) {
 NULL
 
 #' @rdname stan_threads
+#' @deprecated
 #' @export
 #' @return The value of the environment variable `STAN_NUM_THREADS`.
 num_threads <- function() {
+  warning("'stan_threads()' is deprecated. Please use the 'sampling_info()' method of the fit object to obtain the number of threads used.")
```
jgabry, line 344: It should say `num_threads()` is deprecated not  `stan_threads()`, right?

R/model.R
```diff
@@ -419,7 +422,8 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
 #'   * `validate_csv`: (logical) When `TRUE` (the default), validate the
 #'   sampling results in the csv files. Disable if you wish to manually read in
 #'   the sampling results and validate them.
-#'
+#'   * `threads_per_chain`: (positive integer) The number of threads to use per MCMC chaing
```
jgabry, line 425: typo: "MCMC chaing" should be "MCMC chain"

R/model.R
```diff
@@ -419,7 +422,8 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
...
+#'   to run parallel sections of model.
```
jgabry, line 426: I think we might want to add a little bit more to the description here to clarify the distinction between threads and cores for users who aren't super familiar with it. @bbbales2 are suggestions for how to document this?

jgabry:
> One thing I am not really sure what to do is if threads_per_chain > cores.

Good question. @bbbales2 @mitzimorris any thoughts on this? I'm not sure if this should be an error or warning or something else.

bbbales2:
I'll try this later today.

jgabry:
> I'll try this later today.

Thanks!

R/utils.R
rok-cesnovar: Fixed. Thanks!

R/model.R
rok-cesnovar: Fixed

R/model.R
rok-cesnovar: I expanded to: 

`The number of threads to use in parallel sections (for example reduce_sum or map_rect) of a MCMC chain. This is contrast with `cores` that specifies the maximum amount of CPU cores used across all chains.`

Comments/edits welcome.

bbbales2:
> One thing I am not really sure what to do is if threads_per_chain > cores.

There's always the outstanding issue of hyperthreading. I guess the full threat is `cores * threads_per_chain > cores on computer` which I suppose we could check and throw a warning for. I'm not convinced we should unless there's good reason to think we're significantly hurting performance by making too many threads.

This also makes me wonder if the argument should be `threads_per_core`? @jgabry?

---

1. When I add/remove `cpp_options = list(stan_threads = TRUE)`, do I need to force the recompile manually?

2. I'm happy to upgrade use of `set_num_threads` to a deprecation message that errors if it's too difficult to make it work.

3. "Running MCMC with 1 chain(s) on 1 core(s) with 4  thread(s) per chain...", there are two spaces between the number 4 and "thread(s)".

4. When I try to run a model with the new threading interface I get the printout:

```
> fit3 = mp$sample(data = qrr_1_data, chains = 1, cores = 1, threads_per_chain = 4)
Running MCMC with 1 chain(s) on 1 core(s) with 4  thread(s) per chain...
```

And then the whole thing hangs without going any further.

5. If I build a model with threading, restart my rsession, rerun the build command (which correctly reports the binary is already there), and then start sampling, sampling works, but it runs without threading. I get the message at the end:

```
Warning message:
In mpq$sample(data = qrr_2_data, chains = 1, cores = 1, threads_per_chain = 4) :
  'threads_per_chain' is set but the model was not compiled with 'cpp_options = list(stan_threads = TRUE)' so 'threads_per_chain' will have no effect!
```

6. I wonder if that warning message about "'threads_per_chain' will have no effect!" should go at the beginning of sampling?

I'm testing using code and a model from this post on discourse https://discourse.mc-stan.org/t/qr-decomposition-and-reduce-sum-undercutting-each-others-runtime-gains/15210 (which were convenient to me now).

R code is:
```
library(cmdstanr)
library(tidyverse)

df1 <- data.frame(x <- runif(5e+02, 0, 1),
                  x2 <- x^2,
                  y <- rnorm(5e+02, 1 + 2*x - 1*x2, x)) %>%
  select(y=3,
         x=1,
         x2=2)

df1 <- data.frame(x <- runif(5e+02, 0, 1),
                  x2 <- x^2,
                  y <- rnorm(5e+02, 1 + 2*x - 1*x2, x)) %>%
  select(y=3,
         x=1,
         x2=2)

qrr_1_data <-
  list(N=dim(df1)[1],
       M=dim(as.matrix(df1[,c(2:3)]))[2],
       K=length(seq(0.25, 0.75, 0.25)),
       tau=seq(0.25, 0.75, 0.25),
       X=df1[,c(2:3)] %>% data.matrix(),
       y=df1[,1],
       grainsize = 1)

qrr_2_data = qrr_1_data
qrr_2_data$X = t(qrr_1_data$X) # It looks like the QR code takes a transposed X

mpq = cmdstan_model("mpq.stan", cpp_options = list(stan_threads = TRUE))

fit4 = mpq$sample(data = qrr_2_data, chains = 1, cores = 1, threads_per_chain = 4)

fit4$time()$total
```

Model code is:
```
functions{
  real q_loss(real u, real tau){
    return (u * (tau - ((u < 0) ? 1 : 0)));
  }
  real ald_lpdf(real y, real mu, real tau){
    return log(tau) + log(1 - tau)
    - q_loss(y - mu, tau);
  }
  real partial_sum(real[] y_slice,
                   int start, int end,
                   real[] alpha, vector[] beta_tilde, matrix Q, real[] tau,
                   int K){
    real target_tmp = 0.0;
    for(k in 1:K){
      for(n in start:end){
        target_tmp += ald_lpdf(y_slice[n-start+1] | (alpha[k] + dot_product(beta_tilde[k], Q[n])), tau[k]);
      }
    }
    return target_tmp;
  }
}
data {
  int<lower=1> N; //number of observations
  int<lower=1> M; //number of predictors
  int<lower=1> K; //number of quantiles
  real<lower=0, upper=1> tau[K]; //quantiles
  matrix[M, N] X; // predictor matrix
  real y[N]; //response variable
  int<lower=1> grainsize;
}
transformed data {
  matrix[M, N] X_centered;
  matrix[N, M] Q;
  matrix[M, M] R;
  matrix[M, M] R_inv;

  for (m in 1:M){
    X_centered[m] = X[m] - mean(X[m]);
  }

  // Compute, thin, and then scale QR decomposition
  Q = qr_Q(X_centered')[, 1:M] * N;
  R = qr_R(X_centered')[1:M, ] / N;
  R_inv = inverse(R);
}
parameters {
  vector[M] beta_tilde[K];
  real alpha[K];
}
model{
  target += reduce_sum(partial_sum, y, grainsize,
                       alpha, beta_tilde, Q, tau,
                       K);
}
```

bbbales2:
(We shouldn't include that linked code in tests cause of licensing or whatever, by the way. I assume this would reproduce with other model code, just that stuff was conveniently loaded up in my rstudio)

jgabry:
@bbbales2 thanks for reviewing. And good call on `threads_per_core` I think. I'll leave the rest of your comments for @rok-cesnovar.,

rok-cesnovar:
I will first address the minor stuff:

> 1. When I add/remove `cpp_options = list(stan_threads = TRUE)`, do I need to force the recompile manually?

Yes, there is no way of knowing what flags an executable was built with. We could add some suffix to the executable or stuff like that, but that should be addressed separately. Its also a question of whether we can do that in a "thin wrapper".

> 2\. I'm happy to upgrade use of `set_num_threads` to a deprecation message that errors if it's too difficult to make it work.

I am not sure I understand this one. Did it not work?

> 3\. "Running MCMC with 1 chain(s) on 1 core(s) with 4  thread(s) per chain...", there are two spaces between the number 4 and "thread(s)".

Thanks. Fixed.

> 4\. When I try to run a model with the new threading interface I get the printout:
> 
> ```
> > fit3 = mp$sample(data = qrr_1_data, chains = 1, cores = 1, threads_per_chain = 4)
> Running MCMC with 1 chain(s) on 1 core(s) with 4  thread(s) per chain...
> ```
> 
> And then the whole thing hangs without going any further.

This is the case that is problematic because thread_per_chain > cores. Will explain below. This should either stop or warn and use threads_per_chain = 1.

> 5. If I build a model with threading, restart my rsession

Ok, good call. I forgot about that case. This goes back to the first case. We have no way of knowing what flags an executable was compiled with. Will remove this warning. This needs to be addressed at the cmdstan level first.

> 6. I wonder if that warning message about "'threads_per_chain' will have no effect!" should go at the beginning of sampling?

It should be removed and fixed at the cmdstan level. See answer for 5.

bbbales2:
> This is the case that is problematic because thread_per_chain > cores

Ooh, I thought the issue was 'cores' as in number of cores on the computer, not 'cores' the argument. Is this some interaction with processx?

bbbales2:
> I am not sure I understand this one. Did it not work?

It didn't, but also it seems like a different sort of interface so there could end up being weird logic bugs.

Like, `set_num_threads` is a module level interface (I don't even know what happens if two things in one R session both set it). The other thing is a per-function call interface.

Does the second take precedence to the first? Or the first to the second? I could see an argument for either way and so it seems confusing to support both. I think the message makes sense since we've already got people using cmdstanr, but this seems like a version-bump/error-message situation.

rok-cesnovar:
Ok, now for the threads_per_core. 
I probably doc-ed this too lightly. 

The cores argument sets the maximum number of cores used. In rstan or even in cmdstanr wihout threading this is how it works right now: 
- 2 chains, 2 cores -> 2 parallel chains
- 4 chains, 2 cores -> 2 parallel chains followed by 2 parallel chains
- 4 chains, 4 cores -> 4 parallel chains

Then we introduce threading. If we just throw in thread_per_chain to this, we would get:

- 2 chains, 2 cores, 2 threads_per_core -> 2 parallel chains using 4 cores at a time
- 4 chains, 2 cores, 2 threads_per_core -> 2 parallel chains using 4 cores at a time followed by 2 parallel chains using 4 cores at a time
- 4 chains, **4 cores**, 2 threads_per_core -> 3 parallel chains using **8 cores** 

Aki pointed this out and I agree here. If i specify use 4 cores and then 6 are use that is confusing.

So instead what is happening here is:

- 4 chains, 4 cores, 2 threads_per_core -> 2 parallel chains using **4 cores** at a time  followed by 2 chains using 4 cores again

rok-cesnovar:
> It didn't, but also it seems like a different sort of interface so there could end up being weird logic bugs.

> Like, set_num_threads is a module level interface (I don't even know what happens if two things in one R session both set it). The other thing is a per-function call interface.

Ok yeah, lets error there. Good call. Right now the second one takes precedens but yes, too confusing.

bbbales2:
> Aki pointed this out and I agree here. If i specify use 4 cores and then 8 are used that is confusing and weird.

Huh, maybe we should change the `cores` argument to something about chains in parallel?

Also, I'm not sure about the justification. There's this idea with threading in parallelization that if you start more threads than you have cores on your computer than you'll end up slowing things down.

But there's also a sort of opposite justification for threading as well: if you have a lot of calculations going in parallel and one stalls for some reason, the scheduler can switch to another and continue making use of the processor.

Ofc. I guess that thing is more of how threading is used in big I/O calculations, which we aren't really doing. Anyhow I think there's multiple justifications for how threading should be used, and I'd rather not make assumptions about the user.

I suspect doing too many threads will slow stuff down, but I suspect it'll be a 10-15% sorta slowdown (which would be unfortunate, but not a showstopper), not a 50-100% sorta slowdown. If the slowdowns are going to look like the second, then I think we may need to do something.

rok-cesnovar:
Well, if you have 4 cores you can always specifiy: chains = 4, cores = 8, threads_per_chain = 2 which will run all chains.

bbbales2:
> I'd rather not make assumptions about the user.

I'd rather not make *that* assumption about the user, I guess is what I mean. 'That' assumption being that they'll be too troubled to figure out how many cores are on their computer/how many cores are implied by a `cores` + `threads_per_core` argument.

But I guess now that I type that out explicitly it is confusing and I'm bringing along my own assumptions here.

rok-cesnovar:
There was the idea of simultaneous_chain argument also. We can also go that route. In which case the cores argument becomes obsolete.

4 chains, 4 simultaneous_chains, no threading -> 4 parallel chains
4 chains, 4 simultaneous_chains, 2 threads_per_chain -> 4 parallel chains, 8 cores used
4 chains, 2 simultaneous_chains, 2 threads_per_chain -> 2 parallel chains, 4 cores used

Now that I wrote this out, it kind of makes sense. I did not like it at first. It does put the burden of calculating max cores used on the user. I would remove cores if we go this route. The problem I see is that simultaneous chains is a mouthful and a pain to spell. I have a Grammarly extension to Chrome that fixed it for me every single time I wrote it :) But that might be just me being a non-native speaker/writer.

bbbales2:
> There was the idea of simultaneous_chain argument also

Yeah, that is a mouthful. I suppose eventually we'll have to deal with MPI processes too.

I guess, let me add here that a precursor to using `reduce_sum` in a model is deciding that you have more computational resources than N chains can fill. So I think it's fair to assume the user knows what they're getting into with cores/threads/MCMC chains.

> This goes back to the first case. We have no way of knowing what flags an executable was compiled with. Will remove this warning. This needs to be addressed at the cmdstan level first.

Does that mean this pull waits on something at the cmdstan level? Or there's a workaround (this is the issue with reloading the rsession and the object state going away)?

rok-cesnovar:
MPI should land fairly soon, so its very much relevant. Open PRs + generated quantities then MPI.

> So I think it's fair to assume the user knows what they're getting into with cores/threads/MCMC chains.

Agreed.

> Does that mean this pull waits on something at the cmdstan level?
  
Currently a workaround. But given that it is session-dependant it's bad. Let's leave it to the user until 2.24 lands, which should have 

I was thinking something like:
`./model info` would print meta info on the model.

jgabry:
A question that just occurred to me (sorry if this was discussed elsewhere already) is whether we should reset the `STAN_NUM_THREADS` environment variable after running the model. That is, if the user says to use e.g. 2 threads and we internally set `Sys.setenv("STAN_NUM_THREADS" = 2)` should we undo that after CmdStan finishes? Otherwise that environment variable would stick around and would be used in subsequent runs even if the user doesn't specify the threads argument, right?

rok-cesnovar:
> whether we should reset the `STAN_NUM_THREADS` environment variable after running the model

Yes, we probably should. Will add that.

jgabry:
What's the status of this PR? Ready for another look? No rush, just checking in case I was supposed to look at this again

rok-cesnovar:
I think the argument names: cores or simultaneus chains is still undecided. Ben was not a fan. Once that is settled I need to add restoring STAN_NUM_THREADS.

bbbales2:
I'm more worried about this being kinda broken until 2.24 cause the session restart thing.

But yeah the parallelization mechanism/naming too is a thing.

rok-cesnovar:
I have another idea that might work and is not too clever. I will give it a try and tag you when its ready.

rok-cesnovar:
Refactor done. Behaviour explained in the comments. Session restart works fine.
```R
library(cmdstanr)

file <- file.path(cmdstan_path(), "examples", "bernoulli", "bernoulli.stan")
data_list <- list(N = 10, y = c(0,1,0,0,0,0,0,0,0,1))

mod <- cmdstan_model(file)
mod_threads <- cmdstan_model(file, cpp_options = list(stan_threads = TRUE))

# OK
mod$sample(data = data_list)
# warning (using threads_per_chain without threading compiled)
mod$sample(data = data_list, cores = 4, threads_per_chain = 4)

# OK
mod_threads$sample(data = data_list, cores = 4, threads_per_chain = 4)
# stop (compiled with threads but not set)
mod_threads$sample(data = data_list)

# stop (threads_per_chain >  cores)
mod_threads$sample(data = data_list, cores = 1, threads_per_chain = 4)
```

rok-cesnovar:
This now passes all tests and is ready for another look.

See above comment and the `tests/testthat/test-threads.R` file in this PR on how the API would look like. 

What is does behind the scenes is it adds a `_threads` or `_opencl` or `_mpi` suffix (or any combination) to the executable name in order to differentiate from its use without flags. This gives us the added benefit of  two models objects made from the same stan model code but with different flags that point to two different executables.

What I did have to do, is eliminate that a user could create a model object just from the executable without the model and actual compiling), which was a really niche use case and was bound to be bug-prone as cmdstanr has no knowledge of how the executable was compiled.

If https://github.com/stan-dev/cmdstan/issues/887 is done for 2.24, which it will be in some form, if there will be no strong objections to the basic idea, we will be able to eliminate the need for the suffices in executable names and will also be able to create model objects from executables only.

So if you are fine with the above we just need to settle on the arguments. 

I guess the competing ideas are:
- A: `cores`, `chains`, `threads_per_chain` where `floor(cores/threads_per_chain)` is calculated to determine the actual number of simultaneous running chains. If cores < threads_per_chain we error.
- B: `chains`, `simultaneous_chains`, `threads_per_chain` where simultaneous_chains defaults to chains. 

Its currently implemented as A, but I dont really oppose to B either.

jgabry:
Thanks @rok-cesnovar. I think I'm ok with whatever you and @bbbales2 decide on for this. @bbbales2 whenever you have a chance can you take another look? Thanks in advance!

> If [stan-dev/cmdstan#887](https://github.com/stan-dev/cmdstan/issues/887) is done for 2.24, which it will be in some form, if there will be no strong objections to the basic idea, we will be able to eliminate the need for the suffices in executable names and will also be able to create model objects from executables only.

Great!

jgabry:
Looks like there's a minor merge conflict now

rok-cesnovar:
> I think I'm ok with whatever you and @bbbales2 decide on for this. @bbbales2 whenever you have a chance can you take another look?

Do you have a preference Ben?

bbbales2:
I'll check this later today

mitzimorris:
> A: cores, chains, threads_per_chain where floor(cores/threads_per_chain) is calculated to determine the actual number of simultaneous running chains. If cores < threads_per_chain we error.

FWIW, I like this naming and logic

bbbales2:
A is fine. I tested the threading/non-threading and it works for me. I didn't test `_mpi`. @jgabry I'll leave the code review to you.

rok-cesnovar:
> I didn't test `_mpi`

MPI with cmdstanr only compiles for now. Running MPI executables is part of https://github.com/stan-dev/cmdstanr/issues/117 and will be addressed immediately after beta.

jgabry:
Ok great thanks for trying it out @bbbales2.  I'm gonna take one more look through the code now before merging

jgabry:
Here's something I just noticed. Is (3) below supposed to ignore `cpp_options` or use it? Currently it doesn't use it but I think it either should use it or should throw an error saying that `cpp_options` shouldn't be specified if `compile=FALSE`. 

So I guess the question is, should `...` be allowed in `cmdstan_model()` when `compile=FALSE`? 

```r
stan_program <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")

# (1) compile via cmdstan_model()
mod <- cmdstan_model(stan_program, compile = TRUE, cpp_options = list(stan_threads = TRUE))
mod$cpp_options() # correctly includes stan_threads

# (2) compile after cmdstan_model()
mod <- cmdstan_model(stan_program, compile=FALSE)
mod$compile(cpp_options = list(stan_threads = TRUE))
mod$cpp_options() # correctly includes stan_threads

# (3) compile after cmdstan_model() but specify cpp_options via cmdstan_model()
mod <- cmdstan_model(stan_program, compile=FALSE, cpp_options = list(stan_threads = TRUE))
mod$compile()
mod$cpp_options() # empty list
```

jgabry:
Another question that just occurred to me: does threading (e.g. with `reduce_sum`) work only for MCMC or also for ADVI and optimization? If the latter, should we add a threading argument to `$optimize()` and `$variational()`?

bbbales2:
> Another question that just occurred to me: does threading (e.g. with reduce_sum) work only for MCMC or also for ADVI and optimization?

I have never tested this. That's a good point. @wds15, have we ever tested threading stuff with optimization/VI?

jgabry:
Also @bbbales2 the Reduce Sum case study is going to need to be updated again when this gets merged (to use the `threads_per_chain` argument instead of `set_num_threads()`).

Sorry you keep having to update it because of CmdStanR changes!

rok-cesnovar:
@bbbales2 if possible I would wait until we have the R package repo done for the case study (it should hopefully be today). That should make the install step a bit easier (No need to deal with the posterior package dependency).

There is no reason threading would not work for optimization/VI. Will make a separate issue for that.


> Here's something I just noticed.

Will take a look, I think there is an easy fix here.

jgabry:
> Will take a look, I think there is an easy fix here.

Ok cool. @rok-cesnovar I'm going to push some minor edits to the doc for the `threads_per_chain` argument and then once you push your fix I think we're ready to go. I agree that we can do a separate issue for threading support for variational and optimization.

jgabry:
Last thing: 

> Once that is settled I need to add restoring STAN_NUM_THREADS.

does this still need to be done?

wds15:
We never tested threading with anything else than MCMC sampling, but I don’t see why it would not just work fine.

jgabry:
Also, regarding the `cpp_options` issue do we need to do the same thing for `stanc_options` and `include_paths`?

rok-cesnovar:
> Is (3) below supposed to ignore `cpp_options` or use it?

Fixed now.

> does this still need to be done?

I honestly think no. If using `cmdstanr` with threading you need to set it every time so no reason to restore. And Sys.setenv has no effect outside the R session.

There is one potential case where this would cause issues: using threading with cmdstanr & rstan in the same R session.

> issue do we need to do the same thing for `stanc_options` and `include_paths`?

Yes, will add.

rok-cesnovar:
Ready, I think.

R/model.R
```diff
@@ -225,17 +236,47 @@ compile_method <- function(quiet = TRUE,
                            force_recompile = FALSE,
                            #deprecated
                            threads = FALSE) {
-  if (all(nzchar(self$exe_file()))) {
-    exe <- cmdstan_ext(strip_ext(self$stan_file()))
-  } else {
-    exe <- cmdstan_ext(strip_ext(self$exe_file()))
+  if (length(cpp_options) == 0) {
+    if (!is.null(private$cpp_options_)) {
```
jgabry, line 240: @rok-cesnovar If `private$cpp_options_` doesn't have anything then, the way you've coded the CmdStanModel object, it will be an empty `list()` rather than `NULL`. So `!is.null()` will always return TRUE here. I'll push a quick commit to fix this.

jgabry:
Ok I fixed the issue I commented on, but there's still one failing test:

```r
test-model-compile.R:160: failure: switching threads on and off works without rebuild
`after_mtime` is not strictly more than `before_mtime`. Difference: 0
```

This was failing earlier today too but I'm not sure why. If I have more time later I'll see if I can figure it out.

wds15:
Sorry for entering late here... I recall that we now have `threads_per_chain` and `cores`, right?

Up to now `cores` essentially meant how many chains we fire up at the same time. What is the meaning now?

If there is a summary to point me to, let me know. I don't want to steal your time.

rok-cesnovar:
You are not late :)

In the end we had 2 options: 

A: cores, chains, threads_per_chain where floor(cores/threads_per_chain) is calculated to determine the actual number of simultaneous running chains. If cores < threads_per_chain we error.

B: chains, simultaneous_chains, threads_per_chain where simultaneous_chains defaults to chains. So remove cores.

> Up to now `cores` essentially meant how many chains we fire up at the same time.

Yes, because that is how rstan did it because with single threaded, cores is defactor simultaneous chains. For the threaded case that gets weird.

rok-cesnovar:
We went with A, but there is still time to reconsider if you have some thoughts. Better to get it right than early :) 

I have to say that I like B, the only problem is the name `simultaneous_chains` which is a bit of a long word. Option A is more similar to what we have right now.

wds15:
Option A would not allow to over-subscribe the cores. I know that this is nothing we should encourage, but we should definitely enable it. The thing is that it can be a good thing to fire up more threads than you have cores - for example, if you have faster and slower running chains. 

I would prefer to option B as well here. How about an option C where simultaneous_chains is set by default to floor(cores/threads_per_chain)  ? That could be a compromise to leave old syntax mostly intact, but also allow more complex specification.

So my main criticism here is that user would not be able to over-subscribe - and while that is likely anyway inefficient I can imagine that it will be useful in some settings unless we prove otherwise.

rok-cesnovar:
> Option A would not allow to over-subscribe the cores.

Why not? You could run this on a 4-core system.
```
mod$sample(cores = 8, chains = 4, threads_per_chain = 2)
```

Its maybe weird but if you oversubscribe this is essentially what you do. You set the number of cores larger than you actually have. No disagreement from me that oversubscribing can be useful, but 

> How about an option C where simultaneous_chains is set by default to floor(cores/threads_per_chain)? That could be a compromise to leave old syntax mostly intact, but also allow more complex specification.

But what does cores mean in this setting? 
```
mod$sample(cores = 4, chains = 4, simultaneous_chains = 4, threads_per_chain = 2)
```
This means it uses 8 cores actually, but you specified 4. That seems weird.

rok-cesnovar:
> but there's still one failing test:

The problem was that
```
  mod$compile(force_recompile = TRUE, cpp_options = list(stan_threads = TRUE))
  mod$compile(force_recompile = TRUE)
```
both used `stan_threads = TRUE`, because we used the `private$cpp_options_` which can also be the `cpp_options` from the previous compile. I split that up now. Hopefully, that fixes it.

wds15:
Ok, if oversubscription works, fine. How could I run 4 chains, 2 should run concurrently, but I would like to use 4 threads per chain being run?

In the option to go with cores, chains, simultaneous_chains & threads_per_chain I would suggest to use the cores option as a means to set defaults sensibly, but once you use chains, simultaneous_chains & threads_per_chain then everything would be defined. Defining cores in this situation should not be allowed and lead to an error.

rok-cesnovar:
> 4 chains, 2 should run concurrently, but I would like to use 4 threads per chain being run

With:
```r
mod$sample(cores = 8, chains = 4, threads_per_chain = 4)
```
One has to divide cores/threads_per_chain in their head in this case, which is the deficiency of this approach as I think I already mentioned.

> In the option to go with cores, chains, simultaneous_chains & threads_per_chain I would suggest to use the cores option as a means to set defaults sensibly, but once you use chains, simultaneous_chains & threads_per_chain then everything would be defined. 

So this would be the same and both would work? Did I get that right?
```r
mod$sample(cores = 8, chains = 4, threads_per_chain = 4) 
mod$sample(simultaneous_chains = 2, chains = 4, threads_per_chain = 4) 
```

But 
```r
mod$sample(simultaneous_chains = 2, chains = 4, threads_per_chain = 4, cores = 4) 
```
would error? 

I am fine with that. Hard to say how confusing this would be to the users, to have 2 ways of doing the same thing.

If I had to pick one, I would now pick
```r
mod$sample(simultaneous_chains = 2, chains = 4, threads_per_chain = 4) 
```

I would really like to settle this debate once and for all, even if it means delaying this PR for another week or two. Rstan will need to have a similar interface and so will cmdstan if we get to multiple chains in one exe.

wds15:
What you wrote about the option C I have in mind looks correct to me. Users would have a few more options on how they specify things. I would probably suggest to deprecate the cores argument to get rid of it in the long run, but for now it's good to keep cores around as users are used to it and it can be used to set sensible defaults.

So that would be my preference, yes... let's hear the others. Hopefully that is a good compromise to go forward with.

wds15:
Since we are at it... what about MPI? How is that supposed to fit in this?

I am fine with not worrying about MPI, but maybe there is an easy solution. The issue is then that all chains need to start at the same time and we need to specify mpicores_per_chain or something like this. It gets even worse if we wanted to allow MPI to be combined with threads (so between machine communication with MPI and on a machine threading). I don't have a good suggestion for this ATM, but I also don't find it worth the effort to delay this threading interface change.

rok-cesnovar:
Havent thought about that much for now, but 
```r
mod$sample(mpi_procs = 4)
```
woul cover everything right now. And if we ever support MPI+threading
```r
mod$sample(mpi_procs = 4, threads_per_proc = 4)
```
it can also be a separate function
```r
mod$mpi_sample(mpi_procs = 4, threads_per_proc = 4)
```

I think it's just better to have this separate. Too much shadowing can be confusing.

wds15:
Keeping it separate is a good approach. So I would support having `mpi_ sample` which we can work out when we get there.

bbbales2:
It's hard to argue with any of the logic here, so I'm also left thinking this is the best:

```r
mod$sample(simultaneous_chains = 2, chains = 4, threads_per_chain = 4) 
```

What if we replaced `simultaneous_chains` with `parallel_chains`?

Then the default way to run Stan on a computer with 4 cores becomes something like:

```
mod$sample(chains = 4, parallel_chains = 4) 
```

Which doesn't seem bad.

jgabry:
Thanks @bbbales2, I like `parallel_chains`!

rok-cesnovar:
Do we just go with `parallel_chains`? I also like that. Will make the changes asap if everyone is on board.

mitzimorris:
regarding switch from "cores" to "parallel_chains" - much better name!
note that it's moving away from RStan - personally, I'm all for that.

jgabry:
Yeah i'm on board with `parallel_chains`

rok-cesnovar:
Done, if tests pass, this is ready. We still use `num_cores` internally, but I left those. That is a separate issue as that is also used in VI/optimization.

jgabry:
Great, thanks @rok-cesnovar!

jgabry:
Also, I just pushed another commit to use `parallel_chains` instead of `cores` in the vignette

jgabry:
thanks for fixing the tests. i just canceled some travis and appveyor builds that we know will fail so that they will check your most recent fix sooner

rok-cesnovar:
Thanks!

jgabry:
tests are passing now but the travis build using the latest (unstable) R-devel is failing. looks like R-devel is stricter with checking links in documentation. I'll fix this.

jgabry:
Everything is passing, I'm gonna merge it!

PR #195, opened 2020-06-12
Github Actions CI

### LINKED ISSUE #176: Migrate CI to Github Actions
I took a deep dive in Github Actions today and feel like we should at least think about moving Travis/Appveyor CI to Github Actions.

It's very elegant and a ton of R and codecov Github actions are available for reuse. It offers Linux, Windows and Catalina testing in one place, free for public repositories

rok-cesnovar:
Closes #176


jgabry:
Cool, thanks for starting to set this up!  One question for now: I do think we'll need to be running R cmd check with R-devel but I saw you removed it in one of the commits. Is that temporary?

rok-cesnovar:
Yes, temporary. The Mac R-devel takes some time to run (~30 minutes - installing dependencies takes much longer than release) so I just disabled it for now, as I know it works.

I just need to figure out ubuntu 18.04 r-devel and why 3.5 fails.

rok-cesnovar:
R-devel on Ubuntu is currently unsupported as it would require building from source (Travis builds that themselves for their machines). The R-studio people are supposedly working on providing one for Github Actions soon. So we can add that once it's out. But we have Mac/Windows ones instead, which is probably more than enough for the meantime.

For R 3.5 there is a [bug in rmarkdown right now](https://github.com/rstudio/rmarkdown/issues/1831) that doesn't allow building vignettes.
A fix was merged, it just needs to be published.

So we would have these:
- macOS with 4.0.1
- macOS with R-devel
- windows with 4.0.1
- windows with R-devel
- windows with 3.6.3
- ubuntu with 4.0.1
- ubuntu with 3.6.3

This was done based on this template: https://github.com/r-lib/actions/tree/master/examples#standard-ci-workflow

jgabry:
This looks great, thanks! 

> R-devel on Ubuntu is currently unsupported as it would require building from source (Travis builds that themselves for their machines). The R-studio people are supposedly working on providing one for Github Actions soon. So we can add that once it's out. But we have Mac/Windows ones instead, which is probably more than enough for the meantime.

Yeah that's more than enough in the meantime. 


Also, I see the .codecov.yaml file is deleted on this branch. Is the plan to add a separate workflow for that (I see they have one in the examples at `r-lib/actions`)? Or can than be included in the same workflow? Not urgent, just curious.

rok-cesnovar:
Ah yes, forgot on that. Was meaning to try that out also. That example is all we need.

rok-cesnovar:
Test coverage works and so do the tests.

Windows tests currently don't have caching, but that is also coming soon.

jgabry:
Awesome, thanks! 

> Windows tests currently don't have caching, but that is also coming soon.

Do you want to wait to merge until then or should we go ahead and add this later? Either way is fine by me.

rok-cesnovar:
No no, I think there is no need to wait as even if it takes 25 minutes it doesn't matter as free Github Actions allows running up to 20 workers at a time so this won't be a bottleneck. And that is much more than free Travis/Appveyor.

rok-cesnovar:
So if you agree we can merge this.

jgabry:
Sounds good, merging now!

PR #204, opened 2020-06-16
add method fit$inv_metric()

### LINKED ISSUE #202: Easy access to inverse_metric after fitting
Right now `read_sample_csv()` provides access to the estimated inverse metric but it's not accessible via the fitted model object. Should we add a method `fit$inverse_metric()`? If so, then: 

(a) Thoughts on name `fit$inverse_metric()`? 

(b) Sometimes it will be a vector (diagonal) and other times a matrix (if dense). Should we allow it to have different return types or should we always return a matrix that is usually but not always a diagonal matrix? Using the same return type is nice but for large matrices it would wasteful to always return the whole matrix unless we use one of R's sparse matrix types (but that requires adding another R package dependency)

jgabry:
Closes #202

#### Submission Checklist

- [x] Run unit tests

#### Summary

* Adds method `$inv_metric()` for extracting the inverse metric for each chain after MCMC. By default always returns a list of (potentially diagonal) matrices. If argument `matrix=FALSE` then returns a list of (diagonal) vectors, if a diagonal metric was used.

* I called the method `inv_metric` instead of `inverse_metric` to be consistent with the argument name `inv_metric` to `$sample()`. So I also went ahead and renamed `inverse_metric` to `inv_metric` in `read_sample_csv()` for consistency. This renaming is responsible for most of the files changed. 

#### Copyright and Licensing

**Columbia University**


bbbales2:
Tested and works for me.

bbbales2:
If anything change the docs from:

"Should a list of matrices be returned? By default a list of matrices is always returned, even if a diagonal metric was used when fitting the model. If a diagonal metric was used then setting matrix=FALSE will return a list of vectors instead, which uses less memory."

To:

"Return inverse metric. If a diagonal metric was used, setting `matrix = FALSE` returns just the diagonals of the matrices instead of the full matrices. `matrix = FALSE` has no effect for dense metrics."

Is backticks how we indicate options?

@rok-cesnovar sorry I didn't get my comment in in time lol. It's fine this change can happen somewhere else.

rok-cesnovar:
Yeah, trigger-happy for merges is my middle name :) We can push a fix later.

jgabry:
Thanks. I'll push a separate commit updating the doc per @bbbales2's suggestion

PR #212, opened 2020-06-20
Respect variables order

### LINKED ISSUE #211: read_sample_csv doesn't respect order of variables argument
**Describe the bug**

This is just a minor bug in that the CSV is read correctly, but the order specified in the `variables` argument isn't respected. That is, `variables = c("a", "b")` returns the draws in the same order as `variables = c("b", "a")`. 

This isn't a big deal but it's inconvenient because, for example, these two lines create the same plot but the second should really have the x and y axes flipped:

```r
bayesplot::mcmc_scatter(fit$draws("alpha", "beta[1]"))
bayesplot::mcmc_scatter(fit$draws("beta[1]", "alpha"))
```

I think the issue seems to be that the `matching_variables()` helper function that is called by `read_sample_csv()` always returns the same order: 

```r
> matching_variables(variable_filters = c("a", "b"),  variables = letters)
$matching
[1] "a" "b"

$not_found
NULL


> matching_variables(variable_filters = c("b", "a"),  variables = letters)
$matching
[1] "a" "b"

$not_found
NULL
```

**To Reproduce**

```r
fit <- cmdstanr_example("logistic")

# these return the draws in the same order
identical(
 fit$draws(c("alpha", "beta[1]")),
 fit$draws(c("beta[1]", "alpha")) 
)
```

**Expected behavior**

`fit$draws(c("beta[1]", "alpha")) ` should return the draws with `beta[1]` before `alpha`.

**CmdStanR version number**
`0.0.0.9006`

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

fixes #211 

Fixes `matching_variables` and `remaining_columns_to_read` so that variables are returned in the correct order. @rok-cesnovar I made some comments below to explain why I made the changes I did. 

#### Copyright and Licensing

**Columbia University**

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -578,6 +578,11 @@ CmdStanMCMC <- R6::R6Class(
       if (!length(self$output_files(include_failed = FALSE))) {
         stop("No chains finished successfully. Unable to retrieve the draws.")
       }
+      if (inc_warmup && !private$sampling_info_$save_warmup) {
+        stop("Warmup draws were requested from a fit object without them! ",
+             "Please rerun the model with save_warmup = TRUE.", call. = FALSE)
+      }
+
       to_read <- remaining_columns_to_read(
         requested = variables,
         currently_read = dimnames(private$draws_)$variable,
@@ -597,10 +602,6 @@ CmdStanMCMC <- R6::R6Class(
       }
       variables <- repair_variable_names(matching_res$matching)
       if (inc_warmup) {
-        if (!private$sampling_info_$save_warmup) {
-          stop("Warmup draws were requested from a fit object without them! ",
-               "Please rerun the model with save_warmup = TRUE.")
-        }
         posterior::bind_draws(private$warmup_draws_, private$draws_, along="iteration")[,,variables]
       } else {
         private$draws_[,,variables]
--- a/R/read_csv.R
+++ b/R/read_csv.R
@@ -504,8 +504,12 @@ remaining_columns_to_read <- function(requested, currently_read, all) {
     all_remaining <- all[!(all %in% currently_read)]
     unread <- c()
     for (p in requested) {
-      if (any(all_remaining == p) || any(startsWith(all_remaining, paste0(p,".")))) {
+      if (any(all_remaining == p)) {
         unread <- c(unread, p)
+      }
+      is_unread_element <- startsWith(all_remaining, paste0(p,"["))
+      if (any(is_unread_element)) {
+        unread <- c(unread, all_remaining[is_unread_element])
       }
     }
   }
--- a/R/read_csv.R
+++ b/R/read_csv.R
@@ -514,7 +514,7 @@ remaining_columns_to_read <- function(requested, currently_read, all) {
     }
   }
   if (length(unread)) {
-    unread
+    unique(unread)
   } else {
     ""
   }
--- a/R/utils.R
+++ b/R/utils.R
@@ -383,16 +383,17 @@ matching_variables <- function(variable_filters, variables) {
   not_found <- NULL
   variable_filters <- unrepair_variable_names(variable_filters)
   variables <- unrepair_variable_names(variables)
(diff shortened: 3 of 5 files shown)


R/fit.R
```diff
@@ -578,6 +578,11 @@ CmdStanMCMC <- R6::R6Class(
       if (!length(self$output_files(include_failed = FALSE))) {
         stop("No chains finished successfully. Unable to retrieve the draws.")
       }
+      if (inc_warmup && !private$sampling_info_$save_warmup) {
+        stop("Warmup draws were requested from a fit object without them! ",
+             "Please rerun the model with save_warmup = TRUE.", call. = FALSE)
+      }
+
```
jgabry, line 585: unrelated to the bug, but we should throw this error immediately instead of at the end of the function after doing the work of reading in the csv

R/read_csv.R
```diff
@@ -504,9 +504,13 @@ remaining_columns_to_read <- function(requested, currently_read, all) {
...
     unread <- c()
     for (p in requested) {
-      if (any(all_remaining == p) || any(startsWith(all_remaining, paste0(p,".")))) {
+      if (any(all_remaining == p)) {
         unread <- c(unread, p)
       }
+      is_unread_element <- startsWith(all_remaining, paste0(p,"["))
+      if (any(is_unread_element)) {
+        unread <- c(unread, all_remaining[is_unread_element])
+      }
```
jgabry, line 513: instead of including `p` we need to include elements of `all_remaining` that start with `p`. right?

R/utils.R
```diff
@@ -383,16 +383,17 @@ matching_variables <- function(variable_filters, variables) {
...
   variables <- unrepair_variable_names(variables)
-  selected <- rep(FALSE, length(variables))
+  selected <- c()
   for (p in variable_filters) {
-      matches <- variables == p | startsWith(variables, paste0(p, "."))
-      if (!any(matches)) {
-        not_found <- c(not_found, p)
-      }
-      selected <- selected | matches
+    matches <- which(variables == p | startsWith(variables, paste0(p, ".")))
```
jgabry, line 388: if we use `which` we get the position (index) of the match, so this preserves the order

jgabry:
It's the weekend, so no rush on reviewing this

R/read_csv.R
rok-cesnovar: Yes, this is much better. Thanks.

tests/testthat/test-csv.R
```diff
@@ -340,6 +347,19 @@ test_that("remaining_columns_to_read() works", {
...
+  c_mat <- paste0("c[", c(1,2,1,2), ",", c(1,1,2,2), "]")
+  all_vars <- c("a", b_vec, c_mat)
+  expect_equal(
+    remaining_columns_to_read(c("a", "b[2]"), c("a", c_mat), all_vars),
+    "b[2]"
+  )
+  expect_equal(
+    remaining_columns_to_read(c("a", "b", "c"), c(paste0("b[", c(1,4), "]"), "c[1,2]"), all_vars),
+    c("a", "b[2]", "b[3]", "c[1,1]", "c[2,1]", "c[2,2]")
+  )
```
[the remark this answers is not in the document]
jgabry: Good call. I just pushed a commit that makes `remaining_columns_to_read()` call  `unique()` before returning. Now it will be

```r
  expect_equal(
    remaining_columns_to_read(c("a", "b[2]", "b"), c("a", c_mat), all_vars),
    c("b[2]", "b[1]", "b[3]", "b[4]")
  )
```

where the elements of `b` are out of order because `b[2]` was asked for before `b` (i.e., the second `b[2]` is removed). This is a bit weird but I think probably the right behavior if the input is `c("b[2]", "b")`.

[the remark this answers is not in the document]
rok-cesnovar: Yes, its probably weird but as requested. Will merge once tests pass.

PR #216, opened 2020-06-23
Remove remaining uses of rstan 

### LINKED ISSUE #215: Drop rstan from Suggests?
Currently we have a soft dependency on rstan because for CmdStan < 2.22, we use `rstan::stan_rdump()` to write the data file (I believe this is because of a problem with JSON pre 2.22): 

https://github.com/stan-dev/cmdstanr/blob/373a891ef02eff9e8809bde1cb2f10a7c936e191/R/model.R#L868-L871

I think it would be preferable to get rid of all dependence on rstan, so I propose that instead of calling rstan inside CmdStanR we instead throw an error and tell the user to call `rstan::stan_rdump()` themselves. This would affect a very small number of users (only those who insist on using an old CmdStan) and allow us to remove rstan entirely while still supporting R dump files. 

Thoughts? I have a branch that does this so I can make PR if no objection.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #215 by removing all use of rstan from the package. An informative error is thrown if CmdStan < 2.22 and any objects are found with a dimension of 0.  In the process I created a new file `data.R` to keep the data related functions together (`write_stan_json`, `process_data`, and some utility functions only used for data-related things).

#### Copyright and Licensing

**Columbia University**


jgabry:
Yeah but thinking about it now I actually think we can remove `--no-build-vignettes` if we have calls to rstan inside of R markdown chunks that are set to `eval=FALSE`. But actually it doesn't really matter for us since when we eventually submit to CRAN we will need to generate the vignettes locally first instead of letting CRAN generate and check them for us (because CRAN won't have CmdStan installed).

Issue #174, opened 2020-05-16
cmdstanr fit object doesn't contain samples


ssp3nc3r:
I'm not sure whether you want this logged as a bug, feature request, or document change, but it is not clear that the fit object, `fit_mcmc` in the example, does not contain actual samples, only file paths to read them. 

Implications are that when the fit object is saved, _e.g._, 

`saveRDS(fit_mcmc, "fit_mcmc.rds") 
`

and shared on another computer, it won't have access to the samples. Below I show the result of fitting the example model in the help files, transferring them to another computer, and attempting to get a summary of the samples:

```
> fit_mcmc$cmdstan_summary()
Running bin/stansummary \
  /var/folders/pg/qc4jx60s2ys3smt604547ckw0000gn/T/RtmpNc0Dqi/bernoulli-202005161343-1-05db98.csv \
  /var/folders/pg/qc4jx60s2ys3smt604547ckw0000gn/T/RtmpNc0Dqi/bernoulli-202005161343-2-05db98.csv
File /var/folders/pg/qc4jx60s2ys3smt604547ckw0000gn/T/RtmpNc0Dqi/bernoulli-202005161343-1-05db98.csv not found
File /var/folders/pg/qc4jx60s2ys3smt604547ckw0000gn/T/RtmpNc0Dqi/bernoulli-202005161343-2-05db98.csv not found
No valid input files, exiting.
```

Is the current or recommended approach to use the class function,

`fit_mcmc$save_output_files()`

to move `csv` files into the working directory, and then save the fit object together with the output files, and share the group of files together?

mitzimorris:
I think this is a document change - this should be made clearer as part of the vignettes -
it is documented nicely at the function level:

see http://mc-stan.org/cmdstanr/reference/fit-method-save_output_files.html

but agreed, you need to understand what the outputs are, and why.  in this case, the CmdStanMCMC object only reads the sample into memory the first time you access the draws.  if you save the output files to a permanent location, you will be able to access them later, although you'd probably need to resolve filesystem access, etc.

rok-cesnovar:
While writing this, I saw @mitzimorris already answered. Here is my answer anyway :)

Yes, the samples do not get read in the fit object automatically. That is by design. 

If you would do 

```R
fit <- mod$sample(...)
fit$draws()

saveRDS(fit, ...)
```

[...]

Issue #200, opened 2020-06-15
Vignettes for beta release


rok-cesnovar:
We discussed this already, just creating an issue so we dont forget. Also listing ideas that come to mind but please comment if you have any other ideas.

- [ ] threading vignette

Showcase how to use threading. Use the redcard model from the case study from @bbbales2 

- [ ] filtering returned parameters from `$draws()`
Showcase how one can read in only selected parameters.

- [ ] variational inference

- [ ] optimization

- [ ] standalone generate_quantities

- [ ] OpenCL vignette
Showcase using OpenCL with a model with one of the GLMs. The only thing I am not completely sure here is how does this effect tests as not everyone can run these.

jgabry:
Thanks for creating the checklist. 

> * The only thing I am not completely sure here is how does this effect tests as not everyone can run these.

We can exclude that vignette from testing if necessary.

bbbales2:
Small aside, I'm not really sure originally where the redcard model came from. I copied the code from [here](https://github.com/rmcelreath/cmdstan_map_rect_tutorial) but that indicates the data came from further upstream.

bbbales2:
Hmm what's the current status of vignettes/tutorials? These all sound like edge case things. There should be doc that goes over the non-edge case stuff (compile model, run model, look at summaries, extract draws, link to posterior, use with Bayesplot, change adapt delta, change max treedepth, save the model results to disk, etc.).

jgabry:
@bbbales2 The one vignette we currently have at 

http://mc-stan.org/cmdstanr/articles/cmdstanr.html

includes compiling, running a model, looking at summary, extracting draws, plotting a posterior, albeit only brief demonstrations of each of those. 

It doesn't yet include things like changing adapt delta and max_treedepth. Not sure if those should go in that vignette or a separate one (e.g., a troubleshooting vignette), but I agree we should demonstrate that. 

If you want to help flesh out the existing vignette (or just tell me what to flesh it out with) that'd be great!

bbbales2:
> If you want to help flesh out the existing vignette

I'm behind in all the things as is lol. Just trying to make your life harder.

bbbales2:
> albeit only brief demonstrations

I think a good vignette should basically be a bunch of ready to copy-paste code, so brief demonstrations are the thing. The full docs are there to handle the non-brief stuff.

bbbales2:
[...]

PR #217, opened 2020-06-23
Add finalizer method and new vignette

### LINKED ISSUE #28: Implement finalize() methods for the R6 classes?
Do we need this? It can be used to delete the potentially large temp files created by cmdstanr when a user deletes an object with rm() and when R exits. I guess this does come with a small risk that users could lose files by accident, so implementing this should come with adding an argument to the fitting methods (e.g. sample) so the user can specify that files should be saved at that time, without having to wait to call the `save_output_files()` method after fitting the model.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #28 
Advances #200 

This PR adds a finalize method for the CmdStanRun R6 class that removes the temporary files when the object is garbage collected. I documented this in the `output_dir` argument doc, but decided to also add a new vignette that demonstrates this, along with more details about how CmdStanR works.

Here's a rendered version of the new vignette: 

* https://htmlpreview.github.io/?https://github.com/stan-dev/cmdstanr/blob/R6-finalizers/docs/articles/cmdstanr-internals.html

For some reason the title gets cut off when using htmlpreview.github.io but the tentative title is "How does CmdStanR work?", although if you have a better title please let me know!

#### Copyright and Licensing

**Columbia University**


vignettes/cmdstanr-internals.Rmd
```diff
@@ -0,0 +1,293 @@
...
+is no longer empty: 
+
+```{r for-csv-reading}
+draws <- fit$draws() # force CSVs to be read into R
+str(fit)
+```
+
+For models with many parameters, if only some parameters are requested (e.g., by
+specifying the `variables` argument to `fit$draws()`) then CmdStanR will only
+read in the requested variables (unless they have already been read in).
```
rok-cesnovar, line 241: Optional: We could also close https://github.com/stan-dev/cmdstanr/issues/174 by mentioning that in order to store all samples one should call

`fit$draws()`

and then save the fit object.

jgabry: Good idea. I'll add that.

jgabry: Actually, I think it would be necessary to also call `fit$sampler_diagnostics()` to make sure those get saved too. I'll add that also. 

But should we just add a `save` method to the fitted model objects that forces reading everything in before saving? It could look like this:

```r
save = function(file, ...) {
  self$draws()
  self$sampler_diagnostics()
  saveRDS(self, file, ...)
}
```

rok-cesnovar: That would probably be the best yes. Great idea!

R/fit.R
```diff
@@ -599,7 +651,7 @@ CmdStanMCMC <- R6::R6Class(
         all = private$sampling_info_$model_params
       )
 
-      if (is.null(to_read) || (length(to_read) > 0)) {
+      if (is.null(to_read) || any(nzchar(to_read))) {
```
jgabry, line 654: Adding the test for the save_object method resulted in errors, and debugging led me to realize that  
`length(to_read) > 0` is TRUE if `to_read = ""`. For some reason in R `length("")` is 1 not 0 (I always forget that) so we need to use nzchar. 

This means that we were actually reading in everything again even if `to_read` was an empty string! Fixed now.

vignettes/cmdstanr-internals.Rmd
jgabry: Ok I added a method `fit$save_object()`. In doing that I fixed a big bug that meant that we were actually reading in all draws again even if they had already been read in! See comment below. So I'm glad we decided to add the save method or I wouldn't have noticed!

R/fit.R
rok-cesnovar: Ah, didnt know that. I am learning my R as I go :)

jgabry: No worries!. I _should_ have known that but I always forget.

PR #232, opened 2020-06-30
Add hpp_file() and message on main.o compiling

### LINKED ISSUE #229: hpp_file method
CmdStanPy has an open issue to add an `hpp_file` method (https://github.com/stan-dev/cmdstanpy/issues/239) for CmdStanModel objects, so we could also add one to CmdStanR.

### LINKED ISSUE #230: Message to user to expect slow compilation the first time?
@rok-cesnovar The first time the user compiles a model after installing CmdStan they should expect it to take a lot longer than usual (presumably to build the precompiled header, etc.), right? If so, maybe we should print a message the first time they try to compile something to warn them to expect a long compilation time. I can imagine that once we release CmdStanR we could get lots of questions on the forums about that if we don't warn them to expect it. What do you think?

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #229 and fixes #230

#### Copyright and Licensing


tests/testthat/test-csv.R
```diff
@@ -3,9 +3,9 @@ context("read_cmdstan_csv")
 if (not_on_cran()) {
   set_cmdstan_path()
   fit_bernoulli_optimize <- testing_fit("bernoulli", method = "optimize")
-  fit_bernoulli_variational <- testing_fit("bernoulli", method = "variational")
+  fit_bernoulli_variational <- testing_fit("bernoulli", method = "variational", seed = 123)
```
rok-cesnovar, line 6: This has nothing to do with this PR. Ever since I added these, some tests kept failing randomly. It seems that variational fails with some seeds for the bernoulli model. This should fix that.

jgabry: Ok yeah I've noticed that too.

R/model.R
```diff
@@ -313,7 +317,9 @@ compile_method <- function(quiet = TRUE,
 
   temp_stan_file <- tempfile(pattern = "model-", fileext = ".stan")
   file.copy(self$stan_file(), temp_stan_file, overwrite = TRUE)
-  tmp_exe <- cmdstan_ext(strip_ext(temp_stan_file)) # adds .exe on Windows
+  temp_file_no_ext <- strip_ext(temp_stan_file)
+  tmp_exe <- cmdstan_ext(temp_file_no_ext) # adds .exe on Windows
+  private$hpp_file_ <- paste0(temp_file_no_ext, ".hpp")
```
jgabry, line 322: Should we rename the hpp file to include the suffix potentially added via `exe_suffix` (e.g. if compiled for threading)?

R/utils.R
```diff
@@ -216,6 +216,7 @@ prepare_precompiled <- function(cpp_options = list(), quiet = FALSE) {
     model_header_gch_used <- FALSE
   }
   if (!file.exists(main_path_w_flags)) {
+    message("Re-compiling the main object file and precompiled headers. This might take up to a few minutes ...")
```
jgabry, line 219: I think it could be good to explain a bit more in the message, for example:

```r
message(
"Compiling the main object file and precompiled headers (may take up to a few minutes). ",
"This is only necessary the first time a model is compiled after installation or when certain ",
"C++ options are changed."
)
```

It's a bit long but I think more clear. What do you think? 

Also, do you know for which compiler option changes this will be necessary?

rok-cesnovar: This looks good. Will update in the morning.
This is be required for threads, mpi and opencl.

R/model.R
rok-cesnovar: That will be problematic because of how the makefiles work and expect the hpp path to be. I dont think its needed.

rok-cesnovar:
Thanks. Will fix the text in the morning and then it should be good to go.

R/model.R
jgabry: Oh ok no problem

jgabry:
> Thanks. Will fix the text in the morning and then it should be good to go.

Sounds good, thanks!

R/utils.R
rok-cesnovar: What about: 
```R
message(
      "Compiling the main object file and precompiled headers (may take up to a few minutes). ",
      "This is only necessary for the first compilation after installation or when ",
      "threading, MPI or OpenCL are used for the first time."
    )
```

PR #233, opened 2020-06-30
Inits as list or function

### LINKED ISSUE #231: Allow passing inits as a list of R objects or function like RStan
For inits, right now CmdStanR requires specifying initialization files, but we could allow passing in a list of initial values like RStan does. This came up in a brms issue: https://github.com/paul-buerkner/brms/issues/940

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #231 

Initial values can now be specified as a list of lists or as a function. This example demonstrates and can be used to verify:

```r
# this init list and init function should create same initial values
init_list <- list(
  list(alpha = 0, beta = c(-0.5, 0, 0.5)),  # chain 1
  list(alpha = -1, beta = c(1, -0.25, -1))  # chain 2
)

init_fun <- function(chain_id) { # chain_id is optional but used here to match init_list results
  if (chain_id == 1) {
    list(alpha = 0, beta = c(-0.5, 0, 0.5))
  } else if (chain_id == 2) {
    list(alpha = -1, beta = c(1, -0.25, -1))
  }
}

fit_init_list <- cmdstanr_example("logistic", init = init_list, chains = 2)
fit_init_fun <- cmdstanr_example("logistic", init = init_fun, chains = 2)

# check that init files were written correctly when using list
# the output should match the contents of init_list above
jsonlite::read_json(fit_init_list$metadata()$init[1], simplifyVector = TRUE)
jsonlite::read_json(fit_init_list$metadata()$init[2], simplifyVector = TRUE)

# check that init files were written correctly when using function
# should be identical to when using list
jsonlite::read_json(fit_init_fun$metadata()$init[1], simplifyVector = TRUE)
jsonlite::read_json(fit_init_fun$metadata()$init[2], simplifyVector = TRUE)
```

#### Copyright and Licensing

**Columbia University**


jgabry:
@bbbales2 If you have time, do you mind trying to break this?

rok-cesnovar:
I was able to break it with: 

```
init_list <- list(
  list(),  # chain 1
  list()  # chain 2
)

init_fun <- function(chain_id) { # chain_id is optional but used here to match init_list results
  if (chain_id == 1) {
    list()
  } else if (chain_id == 2) {
    list()
  }
}
```
which makes the chains fail. Is that expected for empty lists?

I also think 
```r
jsonlite::read_json(fit_init_list$metadata()$init[1], simplifyVector = TRUE)
```
should be fit$init() or something like that.

jgabry:
> which makes the chains fail. Is that expected for empty lists?

Good catch. It should error before trying to sample. Should be fixed now. 

> I also think
> 
> ```r
> jsonlite::read_json(fit_init_list$metadata()$init[1], simplifyVector = TRUE)
> ```
> 
> should be fit$init() or something like that.

Good idea. I'll add that. Is there any way to tell which initial values CmdStan used when they are generated randomly by CmdStan? It would be nice to let the user check the initial values used by CmdStan regardless of whether the user specified them or not.

rok-cesnovar:
> Is there any way to tell which initial values CmdStan used when they are generated randomly by CmdStan?

Not sure, do you mean like this?
```
fit_no_init <- cmdstanr_example("logistic", chains = 2, quiet = FALSE)
> fit_no_init$metadata()$init
[1] 2 2
```

I don't think we can get much more than that.

jgabry:
If the user doesn't specify inits then CmdStan will generate them randomly, so in theory CmdStan could write those random inits to CSV so the user could inspect them and `fit$init()` could always return them. RStan provides this: 

```r
library(rstan)
fit <- stan_demo("eight_schools") # no inits specified by user 
inits <- get_inits(fit) # can access Stan's randomly generated inits
str(inits)
```
```
List of 4
 $ :List of 3
  ..$ mu   : num 1.91
  ..$ theta: num [1:8(1d)] -0.501 -1.292 -1.546 1.17 1.459 ...
  ..$ tau  : num 2.75
 $ :List of 3
  ..$ mu   : num 0.0567
  ..$ theta: num [1:8(1d)] -0.72868 0.52871 -0.00591 -1.73703 1.85056 ...
  ..$ tau  : num 1.02
 $ :List of 3
  ..$ mu   : num -0.693
  ..$ theta: num [1:8(1d)] 0.562 -1.896 0.149 1.035 1.533 ...
  ..$ tau  : num 3.96
 $ :List of 3
  ..$ mu   : num 1.21
  ..$ theta: num [1:8(1d)] 0.256 0.054 0.801 -1.615 -0.743 ...
  ..$ tau  : num 0.602
```

But I guess this isn't currently possible with CmdStan. Not a huge deal.

DESCRIPTION
```diff
@@ -22,13 +22,13 @@ Roxygen: list(markdown = TRUE, r6 = FALSE)
 SystemRequirements: CmdStan (https://mc-stan.org/users/interfaces/cmdstan)
 Imports: 
     checkmate,
+    jsonlite (>= 1.2.0),
```
jgabry, line 25: @rok-cesnovar At this point we use jsonlite in enough places that we might as well move it from Suggests to Imports.

R/fit.R
```diff
@@ -12,8 +12,9 @@ CmdStanFit <- R6::R6Class(
     },
 
     save_object = function(file, ...) {
-      # CmdStanMCMC has its own implementation
       self$draws()
+      try(self$sampler_diagnostics(), silent = TRUE)
+      try(self$init(), silent = TRUE)
```
jgabry, line 17: We can use `try(..., silent=TRUE)` here to make sure we try to bring in everything before saving but don't error if it's not available. Doing this also for sampler_diagnostics means we don't need a separate method for CmdStanMCMC anymore.

R/fit.R
```diff
@@ -47,6 +48,17 @@ CmdStanFit <- R6::R6Class(
       private$metadata_
     },
 
+    init = function() {
+      if (is.null(private$init_)) {
```
jgabry, line 52: only read in init files if requested

DESCRIPTION
rok-cesnovar: I agree.

R/fit.R
rok-cesnovar: Nice!

jgabry:
Ok great. Looks like all checks are passing so I'm going to merge this.

PR #238, opened 2020-07-02
Feature/compile output dir

### LINKED ISSUE #237: Output directory for $compile
I would like to be able to compile a stan model to a specific directory. 

This would help with model storage and when compiling a model within an R package (my current use case).

This function may be available within the cpp_options using the -o flag, but I cannot figure it out. And I suspect when you go to sample from the model, it might not retain the connection.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #237 

#### Copyright and Licensing


jgabry:
That was fast! I'll take a look.

rok-cesnovar:
I actually wanted something like this also but forgot. I did have it thought out before.

vignettes/cmdstanr-internals.Rmd
```diff
@@ -101,9 +101,15 @@ mod$compile()
...
-containing the Stan program.
+### Executable location
 
+By default, the executable is created in the same directory as the file
+containing the Stan program. You can also specify a different location with the 
+`dir` argument of the `cmdstan_model` function.
+
+```{r compile-method, eval = FALSE}
+mod <- cmdstan_model(stan_file, dir = "path/to/directory/for/executable")
+```
```
jgabry, line 112: There's already a chunk named `compile-method` so knitting this vignette gives the error: 

```
processing file: cmdstanr-internals.Rmd
Error in parse_block(g[-1], g[1], params.src) : 
  duplicate label 'compile-method'
```

Just needs a new chunk name like `compile-with-dir`

tests/testthat/test-model-compile.R
```diff
@@ -190,3 +190,23 @@ test_that("compile errors are shown", {
     "An error occured during compilation! See the message above for more information."
   )
 })
+
+test_that("dir arg works for cmdstan_model", {
+  skip_on_cran()
+  tmp_dir <- tempdir()
+  mod_dir <- cmdstan_model(stan_program, dir = tmp_dir)
+  expect_equal(dirname(mod_dir$exe_file()), tmp_dir)
+  expect_true(file.exists(mod_dir$exe_file()))
```
jgabry, line 199: It doesn't really matter but since we're already using checkmate we could use (here and below)

```
checkmate::expect_file_exists(...)
```

instead of 

```
testthat::expect_true(file.exists(...))
```

R/model.R
```diff
@@ -129,8 +129,8 @@
 #' fit_optim_w_init_list$init()
 #' }
 #'
-cmdstan_model <- function(stan_file, compile = TRUE, ...) {
-  CmdStanModel$new(stan_file = stan_file, compile = compile, ...)
+cmdstan_model <- function(stan_file, dir = NULL, compile = TRUE, ...) {
```
jgabry, line 134: This is fine, but for consistency I wonder if `dir` should be an argument to the `compile()` method that can get passed through `cmdstan_model()` via `...` like the other arguments to `compile()`. What do you think?

R/model.R
```diff
@@ -331,9 +336,13 @@ compile_method <- function(quiet = TRUE,
     exe_suffix <- paste0("_", exe_suffix)
   }
   if (all(nzchar(self$exe_file()))) {
-    exe <- cmdstan_ext(paste0(strip_ext(self$stan_file()), exe_suffix))
+    if (is.null(private$dir_)) {
+      stan_file <- self$stan_file()
+    } else {
+      stan_file <- file.path(private$dir_, basename(self$stan_file()))
```
jgabry, line 344: Do we need to wrap `private$dir_` in `absolute_path()`? We might not (maybe that will already happen later to `stan_file`?)

jgabry:
I'm going to restart the checks. One of the mac builds failed with 

```
 *** caught segfault ***
address 0x7fdb1582ffff, cause 'memory not mapped'
```

which doesn't look good and seems to happen occasionally. But I've never seen this locally on my mac.

R/model.R
rok-cesnovar: Agreed

R/model.R
rok-cesnovar: Good call yes! Absolute path should be used.

jgabry:
@rok-cesnovar A few things: 

1) We keep getting segfaults occasionally on the Mac builds. I have no idea why. Does this happen in the CmdStan unit tests ever or just with CmdStanR? We don't need to solve that for this PR (I don't think it's related), but it's a bit concerning. 

2) I think we can simplify a bunch of the code in `cmdstan_model()`: 

All of this: 

```r
      args <- list(...)
      cpp_options_exists <- "cpp_options" %in% names(args)
      if (cpp_options_exists) {
        private$precompile_cpp_options_ = args$cpp_options
      }
      stanc_options_exists <- "stanc_options" %in% names(args)
      if (stanc_options_exists) {
        private$precompile_stanc_options_ = args$stanc_options
      }
      include_paths_exists <- "include_paths" %in% names(args)
      if (include_paths_exists) {
        private$precompile_include_paths_ = args$include_paths
      }
      dir_arg_exists <- "dir" %in% names(args)
      if (dir_arg_exists) {
        dir <- args$dir
        checkmate::assert_directory_exists(dir, access = "r")
        private$dir_ <- dir
      }      
```

can all be replaced with 

```r
      args <- list(...)
      private$precompile_cpp_options_ <- args$cpp_options %||% list()
      private$precompile_stanc_options_ <- args$stanc_options %||% list()
      private$precompile_include_paths_ <- args$include_paths # will be NULL if not included in args
      private$dir_ <- args$dir # will be NULL if not included in args
      # check that dir exists in compile method instead
```

Do you see any issues with that? If not I'll go ahead and push that and make the corresponding change in `compile_method` to check that dir exists. 

3) I think the `dir` argument should be on of the first arguments to the compile method. If you're ok with that I can change it. 

4) This test needs to expect the full path, not just ABCD (I can fix it): 

```
  expect_error(
    cmdstan_model(stan_program, dir = "ABCD"),
    "Assertion on 'dir' failed: Directory 'ABCD' does not exists."
  )
```

rok-cesnovar:
> 1. We keep getting segfaults occasionally on the Mac builds. I have no idea why. Does this happen in the CmdStan unit tests ever or just with CmdStanR? We don't need to solve that for this PR (I don't think it's related), but it's a bit concerning.

I have never seen this with the cmdstan tests, so its either something with cmdstanr or some issue with the devel version of R ( I think this happens only on the devel version. Will try to play with this next week.

> 2\. I think we can simplify a bunch of the code in `cmdstan_model()`:

Nice! This is much nicer and should be done, definitely. 

Also agree on 3 and 4. Do you want me to make the changes? I have some time now.

jgabry:
> Also agree on 3 and 4. Do you want me to make the changes? I have some time now.

I can do it. I basically did it locally already, so I'll just clean it up and push.

rok-cesnovar:
Thanks!

jgabry:
> I have never seen this with the cmdstan tests, so its either something with cmdstanr or some issue with the devel version of R ( I think this happens only on the devel version. Will try to play with this next week.

Ok thanks.

jgabry:
Ok I pushed those changes but noticed that `hpp_file()` doesn't quite work properly. Can you fix this if you have time? 

I think the issue is that after

https://github.com/stan-dev/cmdstanr/blob/daf827ed8b93071294c2faf8a15dc3262fa94233/R/model.R#L385

it never gets copied over the non-temporary location like the exe file does: 

```r
> stan_program <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")
> mod <- cmdstan_model(stan_program)
Compiling Stan program...
> mod$exe_file() 
[1] "/Users/jgabry/.cmdstanr/cmdstan-2.23.0/examples/bernoulli/bernoulli"
> mod$hpp_file() # still a temp file
[1] "/var/folders/h6/14xy_35x4wd2tz542dn0qhtc0000gn/T//RtmpUeKzTW/model-101247011afbc.hpp"
```

Also if you do `cmdstan_model()` again then `hpp_file()` doesn't even return the temporary path, just `character(0)`: 

```
> mod <- cmdstan_model(stan_program)
Model executable is up to date!
> mod$exe_file()
[1] "/Users/jgabry/.cmdstanr/cmdstan-2.23.0/examples/bernoulli/bernoulli"
> mod$hpp_file()
character(0)
```

rok-cesnovar:
> it never gets copied over the non-temporary location like the exe file does:

I did that intentionally. 

The reasoning is that this is a power-user thing and something I don't envision will be useful all that often, except for maybe reporting errors. And for something with such a niche use-case, I don't think it makes sense to pollute the user's space with the .hpp file for each model. No strong feelings here though, so if you feel it's useful, we can change that.

My ideas for improving the current interface for hpp file would be: 
- add a save_hpp_file(dir) function
- if the hpp file is no longer present in the temp location, throw an error or warning that in order to obtain the hpp file, the user should use force_recompile (for both hpp_file() and save_hpp_file()). 

I will make an issue if you agree with these improvements.

jgabry:
Ok I see, that makes sense. And yeah this doesn't have to be part of this PR, so a separate issue is fine.

rok-cesnovar:
Will do this separately once this PR is merged. Is this good to go or is there anything else?

jgabry:
Good to go!

jgabry:
Looks like all checks passed, so I'll merge this now

jgabry:
> I have never seen this with the cmdstan tests, so its either something with cmdstanr or some issue with the devel version of R ( I think this happens only on the devel version. Will try to play with this next week.

FYI, I just saw a failure on the mac build using the released version of R, so not just a devel issue. Not sure if that's better or worse ;)

PR #246, opened 2020-07-21
Use the hyphened --include-paths stancflag for 2.24+

### LINKED ISSUE #245: --include-paths warning in cmdstan 2.24
**Describe the bug**

Trying out cmdstan 2.24.0 rc1, I get the warning "--include_paths is deprecated. Please use --include-paths."

**To Reproduce**
First, install [cmdstan 2.24.0-rc1](https://github.com/stan-dev/cmdstan/releases/tag/v2.24.0-rc1)

Then:
```r
library(cmdstanr)
file <- file.path(cmdstan_path(), "examples", "bernoulli", "bernoulli.stan")
mod <- cmdstan_model(file, include_paths=".", force_recompile=TRUE)
#> Compiling Stan program...
#> --include_paths is deprecated. Please use --include-paths.
#> Then lots of compiler warnings, probably unrelated

```

**Expected behavior**
No `include_paths` warning.

**Operating system**
Ubuntu 20.04

**CmdStanR version number**
 0.0.0.9008 (e33e2b4)

**Additional context**
Minor request: it would be convenient if `include_paths` included the directory of the model file, as mentioned in https://github.com/stan-dev/cmdstanr/issues/59#issuecomment-549008393, but it's not a big deal.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Cmdstan 2.24 deprecated --include_paths in favor of --include-paths. So we should use that. Fixes #245 

The issue also has a request: 
"It would be convenient if include_paths included the directory of the model file"

This PR does not address that. I am not sure what would be best there. I guess there might be situation where users don't want "." in include paths.

#### Copyright and Licensing

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -454,7 +454,9 @@ compile_method <- function(quiet = TRUE,
     }
   }
   stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
-  prepare_precompiled(cpp_options, quiet)
+  if (cmdstan_version() < "2.24") {
+    prepare_precompiled(cpp_options, quiet)
+  }
   run_log <- processx::run(
     command = make_cmd(),
     args = c(tmp_exe,
--- a/R/model.R
+++ b/R/model.R
@@ -428,8 +428,14 @@ compile_method <- function(quiet = TRUE,
     checkmate::assert_directory_exists(include_paths, access = "r")
     include_paths <- absolute_path(include_paths)
     include_paths <- paste0(include_paths, collapse = ",")
-    stancflags_val <- paste0(stancflags_val, " --include_paths=", include_paths, " ")
+    if (cmdstan_version() >= "2.24") {
+      include_paths_flag <- " --include-paths="
+    } else {
+      include_paths_flag <- " --include_paths="
+    }
+    stancflags_val <- paste0(stancflags_val, include_paths_flag, include_paths, " ")
   }
+
   if (!is.null(cpp_options$stan_opencl)) {
     stanc_options[["use-opencl"]] <- TRUE
   }
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -71,7 +71,7 @@ jobs:
           remotes::install_cran("rcmdcheck")
           install.packages("posterior", repos = c("https://mc-stan.org/r-packages/", getOption("repos")), type="source")
           install.packages("cmdstanr", repos = c("https://mc-stan.org/r-packages/", getOption("repos")), type="source")
-          cmdstanr::install_cmdstan(cores = 2, overwrite = TRUE, release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.24.0-rc1/cmdstan-2.24.0-rc1.tar.gz")
+          cmdstanr::install_cmdstan(cores = 2, overwrite = TRUE, release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.24.0/cmdstan-2.24.0.tar.gz")
         shell: Rscript {0}
 
       - name: Session info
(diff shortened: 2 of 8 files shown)


jgabry:
> The issue also has a request:
> "It would be convenient if include_paths included the directory of the model file"
> 
> This PR does not address that. I am not sure what would be best there. I guess there might be situation where users don't want "." in include paths.

I think that most times they would want `"."` included but you're right that there could be cases where they don't. I'm not sure what to do about that either.

jgabry:
Also, should we update the tests to use the release candidate instead of the latest release?

rok-cesnovar:
> I'm not sure what to do about that either.

I think we could have the logic: "if no paths set use '.', else use whatever set". But that should be fixed at the stanc3 level.

> Also, should we update the tests to use the release candidate instead of the latest release?

Sure yeah. Will add that here.

jgabry:
Actually maybe `cmdstan_default_path()` should be changed to not remove release candidates when setting the default path? 

The way it is currently 

https://github.com/stan-dev/cmdstanr/blob/e33e2b46828c3c72cd1325d9e27caa2828d6a27e/R/path.R#L114-L115

it will ignore the RC if an earlier release is present and, more importantly, if only the RC is installed the path won't get set automatically at all, and that could lead to errors.

rok-cesnovar:
Yes, I agree. Just need to double check that if both X and X-rc are present it doesnt pickup the RC. Will fix tomorrow.

jgabry:
Ok cool, thanks!

jgabry:
@rok-cesnovar Do you know why the tests are failing or want help looking into it?

rok-cesnovar:
I think I do know, just haven't had the time to look into this. Will do today. Cmdstan 2.24 no longer needs the precompiled main.o as it now takes care of it, the test is however written in such a way that it expects. Should be a simple fix.

rok-cesnovar:
@jgabry fixed now. I left the tests for 2.23 in, though we could also delete them as I imagine we will always test with the latest cmdstan version, otherwise, it will become a maintenance burden.

R/model.R
```diff
@@ -448,7 +454,9 @@ compile_method <- function(quiet = TRUE,
     }
   }
   stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
-  prepare_precompiled(cpp_options, quiet)
+  if (cmdstan_version() < "2.24") {
+    prepare_precompiled(cpp_options, quiet)
```
rok-cesnovar, line 458: We can leave the `prepare_precompiled` function in for now, but I would maybe remove it after a release cycle, so when 2.25 is out? What do you think?

This function just makes versions 2.23 and earlier a bit more user friendly, but everything still works even without it, though a bit more care needs to be taken when using the thread,mpi and opencl flags with it.

So the function will become more or less dead code once users move to 2.24.

rok-cesnovar:
Now using the freshly released 2.24 instead of the RC.

R/model.R
jgabry: Sure, that makes. Let's remove it once 2.25 comes out. Should we open an issue to remind ourselves?

jgabry:
Looks good!

> I think I do know, just haven't had the time to look into this.

Thanks, didn't mean to rush you! 

> I left the tests for 2.23 in, though we could also delete them as I imagine we will always test with the latest cmdstan version, otherwise, it will become a maintenance burden.

Yeah deleting them is fine. Do you want to do that before merging?

rok-cesnovar:
> Do you want to do that before merging?

Done. I also bumped the version, 0.0.0.9010 is now recommended for use with > 2.23.

jgabry:
Great, thanks! I assume the checks will pass but let's let them run and then we can merge.

rok-cesnovar:
Yes, lets wait. Also added the link to the new install instructions, so we can close https://github.com/stan-dev/cmdstanr/issues/248

jgabry:
Ok great thanks

jgabry:
Do the new install instructions replace the blog post we've been linking to, or do we need both?

rok-cesnovar:
The blog post probably no longer needs to be linked no.

PR #251, opened 2020-07-24
use read.csv if vroom fails

### LINKED ISSUE #249: Fall back to read.csv in case of a vroom error
Suggested by @maedoc

The idea in pseudocde is: 
```
try {
//read with vroom
} catch {
warning("There was an issue with reading the CSV file with vroom. Using the utils::read_csv")
// read entire file with read_csv
}
```

@jgabry something like this would do? Any suggestions on the message?

jgabry:
Fixes #249

#### Submission Checklist

- [x] Run unit tests

#### Summary

Uses `utils::read_csv()` if `vroom::vroom()` errors. Throws a warning asking users to report. 

#### Copyright and Licensing

**Columbia University**

### DIFF
--- a/R/read_csv.R
+++ b/R/read_csv.R
@@ -228,7 +228,9 @@ read_cmdstan_csv <- function(files,
     draws <- try(silent = TRUE, expr = {
       suppressWarnings(do.call(vroom::vroom, vroom_args))
     })
-    if (inherits(draws, "try-error")) {
+    if (!inherits(draws, "try-error")) {
+      draws <- draws[!is.na(draws$lp__), ]
+    } else {
       if (vroom_warnings == 0) { # only warn the first time instead of for every csv file
         warning(
           "Fast CSV reading with vroom::vroom() failed. Using utils::read.csv() instead. ",
@@ -241,8 +243,6 @@ read_cmdstan_csv <- function(files,
       draws <- utils::read.csv(output_file, comment.char = "#", skip = metadata$lines_to_skip)
       draws <- draws[, col_select]
     }
-    draws <- draws[!is.na(draws$lp__), ]
-
 
     if (nrow(draws) > 0) {
       if (metadata$method == "sample") {
--- a/R/read_csv.R
+++ b/R/read_csv.R
@@ -121,6 +121,8 @@ read_cmdstan_csv <- function(files,
   col_types <- NULL
   col_select <- NULL
   not_matching <- c()
+  vroom_warnings <- 0
+
   for (output_file in files) {
     if (is.null(metadata)) {
       metadata <- read_csv_metadata(output_file)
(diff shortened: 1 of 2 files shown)


R/read_csv.R
```diff
@@ -204,41 +206,44 @@ read_cmdstan_csv <- function(files,
       all_draws <- 1
     }
 
+    vroom_args <- list(
```
jgabry, line 209: In order to not have to do try() multiple times, make list of vroom arguments so we only need one call to vroom() in the code. previously we had separate calls depending on generated quantities or not.

R/read_csv.R
```diff
@@ -204,41 +206,44 @@ read_cmdstan_csv <- function(files,
...
-          n_max = all_draws * 2
+      vroom_args$col_types <- c("lp__" = "d")
+      vroom_args$n_max <- all_draws * 2
+    }
+
+    draws <- try(silent = TRUE, expr = {
+      suppressWarnings(do.call(vroom::vroom, vroom_args))
+    })
+    if (inherits(draws, "try-error")) {
+      if (vroom_warnings == 0) { # only warn the first time instead of for every csv file
```
jgabry, line 232: Checking that we haven't yet thrown the warning let's us warn once instead of repeating the warning every csv file

R/read_csv.R
```diff
@@ -204,41 +206,44 @@ read_cmdstan_csv <- function(files,
...
-          progress = FALSE,
-          skip = metadata$lines_to_skip,
-          n_max = all_draws * 2
+      vroom_args$col_types <- c("lp__" = "d")
+      vroom_args$n_max <- all_draws * 2
+    }
+
+    draws <- try(silent = TRUE, expr = {
+      suppressWarnings(do.call(vroom::vroom, vroom_args))
+    })
```
jgabry, line 230: Not sure how to add a unit test for this, so to verify the behavior you can temporarily change this to 

```r
draws <- try(silent = TRUE, expr = stop())
```

which will force it to error and you can verify that it successfully uses read.csv() instead and throws the warning.

R/read_csv.R
```diff
@@ -121,6 +121,8 @@ read_cmdstan_csv <- function(files,
   col_types <- NULL
   col_select <- NULL
   not_matching <- c()
+  vroom_warnings <- 0
```
jgabry, line 124: Keep track of whether we've thrown a warning so that we can avoid throwing it for every csv file. Once per call to read_cmdstan_csv() seems like enough.

R/read_csv.R
rok-cesnovar: Not sure how to make it error either. I think its fine without a test for now. If we think of something we can add it later.

jgabry:
I think I should have bumped the version number with this so we can tell if the user has this version. Are you ok if I do that now on master?

rok-cesnovar:
Yeah sure!

Issue #240, opened 2020-07-03
Implement equivalent of model_pars and par_dims of stanfit from rstan


rok-cesnovar:
Running:

```r
fit <- cmdstanr_example("logistic")
rstanfit <- rstan::read_stan_csv(fit$output_files())
rstanfit@model_pars
rstanfit@par_dims
```
prints
```
> rstanfit@model_pars
[1] "alpha" "beta"  "lp__" 
> rstanfit@par_dims
$alpha
integer(0)

$beta
[1] 3

$lp__
integer(0)
```
Should we implement the equivalent of this? If yes, any name suggestions?

Or are there any plans to add this to posterior (like posterior::variables()). Retrieving dimensions and names of parameters/variables isn't Stan specific. I can also add this to posterior if it would be better to live there.

jgabry:
I agree this is useful information. I think some version of this could go in posterior but it's a little bit Stan-specific in that the dimension of `alpha` is `integer(0)` instead of `1` (to indicate a scalar instead of a 1-element vector or array). But that's a choice that RStan made to in some way match the Stan program. For posterior my first instinct is that `alpha` should have dimension `1` (but we should think about that and check with @paul-buerkner). 

And maybe we can combine `model_pars` and `par_dims` into a single function `variable_structure()` (or something like that) which takes a `draws` object as input and just returns the named list?

```
# assume the draws object has variables
# alpha, beta[1], beta[2], beta[3], Sigma[1,1], Sigma[1,2], Sigma[2,1], Sigma[2,1], lp__

> variable_structure(draws)
$alpha
[1] 1

$beta
[1] 3

$Sigma
[1] 2 2

$lp__
[1] 1
```

Paul will also probably have suggestions if you open an issue over at posterior.

jgabry:
Although maybe we should also add the equivalent of `model_pars` to `fit$metadata()`. Right now `fit$metadata()$model_pars` gives all element names, so we could add `fit$metadata()$model_pars_named` or something that only lists the unique non-element names.

rok-cesnovar:
> Although maybe we should also add the equivalent of `model_pars` to `fit$metadata()`. Right now `fit$metadata()$model_pars` gives all element names, so we could add `fit$metadata()$model_pars_named` or something that only lists the unique non-element names.

Agree. That also makes read_cmdstan_csv a more complete function standalone.

paul-buerkner:
[...]

PR #256, opened 2020-07-28
Return variable names and dimensions

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Attempt at fixing #240 

For now, this just returns the variable names and dimensions as part of metadata in read_cmdstan_csv. We should add it in fit, but not sure on the name.

Example:

```
> f <- cmdstanr_example("logistic", method = "sample")
> csv <- read_cmdstan_csv(f$output_files())
> f1$metadata$model_params_dims
$lp__
[1] 0

$alpha
[1] 0

$beta
[1] 3

> f1$metadata$model_params_named
[1] "lp__"  "alpha" "beta"
```

#### Copyright and Licensing


rok-cesnovar:
@jgabry we can hold off with this for after beta if you want. I am done for today, but if you have any feedback I can finish this tomorrow. 

Mostly want ideas for names. Does `metadata$model_params_dims` and `metadata$model_params_named` work?
Would fit$draws_dims() work? Or should this be a function on draws rather than a part of fit?

jgabry:
Ok cool thanks I'll take a look.

R/utils.R
```diff
@@ -326,3 +326,28 @@ matching_variables <- function(variable_filters, variables) {
...
+    var_names <- gsub("\\]", "", variable_names)
+    for(var in uniq_variable_names) {
+      pattern <- paste0(var, "\\[")
+      var_indices <- var_names[grep(pattern, var_names)]
+      var_indices <- gsub(pattern, "", var_indices)
+      if(length(var_indices)) {
+        var_indices <- strsplit(var_indices[length(var_indices)], ",")[[1]]
+        dims[[var]] <- as.numeric(var_indices)
+      } else {
+        dims[[var]] <- 0
```
jgabry, line 348: This is consistent with RStan, but what do you think about using 1 instead of 0 for a scalar real? I guess the question is whether we want to differentiate in R between the Stan types 

```
real alpha;
real alpha[1];
```

and I'm not sure if we do or don't. What do you think?

jgabry: I guess the question is what will people use this for?

tests/testthat/test-utils.R
```diff
@@ -128,3 +128,25 @@ test_that("cmdstan_make_local() works", {
...
+  expect_equal(variable_dims(vars), vars_dims)
+
+  vars <- c("c[1,1]", "c[1,2]", "c[1,3]", "c[2,1]", "c[2,2]", "c[2,3]", "b[1]", "b[2]", "b[3]", "b[4]")
+  vars_dims <- list(c = c(2,3), b = 4)
+  expect_equal(variable_dims(vars), vars_dims)
+
+  # wrong dimensions for descending order
+  vars <- c("c[1,1]", "c[1,2]", "c[1,3]", "c[2,3]", "c[2,2]", "c[2,1]", "b[4]", "b[2]", "b[3]", "b[1]")
+  vars_dims <- list(c = c(2,1), b = 1)
+  expect_equal(variable_dims(vars), vars_dims)
```
jgabry, line 151: What about throwing an error if the inputs are in the wrong order instead of testing that it does the expected incorrect behavior? Maybe just check that `all.equal(variable_names, sort(variable.names))` and if not throw an error?

R/read_csv.R
```diff
@@ -321,6 +321,9 @@ read_cmdstan_csv <- function(files,
   metadata$inv_metric <- NULL
   metadata$lines_to_skip <- NULL
   metadata$model_params <- repair_variable_names(metadata$model_params)
+  model_param_dims <- variable_dims(metadata$model_params)
+  metadata$model_params_dims <- model_param_dims
+  metadata$model_params_named <- names(model_param_dims)
```
jgabry, line 326: For the names what about using `metadata$variable_dims` and `metadata$variable_names`?

tests/testthat/test-utils.R
rok-cesnovar: I was trying to avoid the sort here, because params in the csv are always sorted correctly.

Though a sort even with 50k parameters is not going to be a huge bottleneck. If we are fine with sorting here, we can just sort descending and take the first element. Then there is no need to warn.

R/utils.R
rok-cesnovar: In R I think there is no difference so I do agree with just going with 1. 

The only reason for sticking with 0 is if this would be a problem for any packages looking into switchinh to cmdstanr. I dont have a good feeling if it would be.

R/read_csv.R
rok-cesnovar: And then matching fit$variable_dims()? That works for me.

jgabry: I'm ok with that, but I wonder if we need both `fit$metadata()` _and_ methods for accessing elements of `fit$metadata()`. Could we just tell the user to use `fit$metadata()$variable_dims` if they want it? That way we don't need to provide methods for accessing elements of the return object of other methods.  What do you think? I don't feel too strongly about it but worth considering before we add new methods.

tests/testthat/test-utils.R
jgabry: Yeah avoiding an unnecessary sort is always nice! I guess if there's really zero risk of having the input in the wrong order then I think it's fine to avoid the sort.

mitzimorris:
> Mostly want ideas for names. Does `metadata$model_params_dims` and `metadata$model_params_named` work?
> Would fit$draws_dims() work? Or should this be a function on draws rather than a part of fit?

hi, could we co-ordinate w/ CmdStanPy on names?  we went through this a while back - last thing I implemented there was https://github.com/stan-dev/cmdstanpy/pull/252 - called everything "stan_variable".

R/utils.R
jgabry: Yeah I guess that's possible, but probably unlikely. Since CmdStanR is new I think we should do what makes the most sense, which I think for an R users would be `1`.

jgabry:
Yes good point @mitzimorris!

jgabry:
@mitzimorris Does `stan_var_dims()` in CmdStanPy say that a variable 

```
real alpha
```

has dim 0 or 1?

R/utils.R
```diff
@@ -326,3 +326,36 @@ matching_variables <- function(variable_filters, variables) {
...
+#'
+variable_dims <- function(variable_names = NULL) {
+  if (is.null(variable_names)) {
+    return(NULL)
+  }
+  dims <- list()
+  uniq_variable_names <- unique(gsub("\\[.*\\]", "", variable_names))
+  var_names <- gsub("\\]", "", variable_names)
+  for (var in uniq_variable_names) {
+    pattern <- paste0("^", var, "\\[")
```
jgabry, line 350: @rok-cesnovar I added "^" because otherwise this line wouldn't work properly if there was a parameter `x[1]` _and_ another parameter `xx[1]`

jgabry:
@rok-cesnovar I pushed a few edits and I think this is ok now other than changing the names. Are you ok with changing to `"stan_variable"`, which is what @mitzimorris said CmdStanPy uses?

mitzimorris:
> @mitzimorris Does `stan_var_dims()` in CmdStanPy say that a variable
> 
> ```
> real alpha
> ```
> 
> has dim 0 or 1?

property `stan_var_dims` returns a Python `dict`  which maps
Stan program variable names to variable dimensions.
Scalar types have int value '1'.  Structured types have list of dims,
e.g.,  program variable ``vector[10] foo`` has entry ``('foo', [10])``.

given fit object `bernoulli_fit`, `stan_var_dims` returns  `{'theta': 1}`
given fit object `logistic_fit`,  `stan_var_dims` returns `{'beta': (2,)}`

R/read_csv.R
rok-cesnovar: Yes. I am fine with that. Good call.

rok-cesnovar:
@jgabry changed to stan_var_dims and stan_variables. Also added a few more tests. I think its good to go.

jgabry:
Looks good! This is good to go, just one super nitpicky thing (sorry!): I think it's slightly weird to use `stan_variables` but then use `stan_var_dims` instead of `stan_variable_dims`. I would prefer using either `stan_vars` and `stan_var_dims` or `stan_variables` and `stan_variable_dims`. I think I prefer the latter, but I'm fine either way, as long as it's consistent. @mitzimorris and @rok-cesnovar what do you think?

rok-cesnovar:
> I would prefer using either `stan_vars` and `stan_var_dims` or `stan_variables` and `stan_variable_dims`.

I agree. That annoyed me a bit as well to be honest but thought that was discussed and agreed on earlier. I prefer the verbose version as well.

mitzimorris:
agreed.  will change CmdStanPy accordingly.

rok-cesnovar:
@jgabry renamed. Good to merge?

jgabry:
Yeah looks good. Will merge now.

PR #263, opened 2020-07-31
make vroom optional

### LINKED ISSUE #262: Replace vroom with fread
**Describe the issue**

~We should move `vroom` from `Imports` to `Suggests` in the `DESCRIPTION` file. This was suggested by [Michael Peck on Discourse](https://discourse.mc-stan.org/t/lightweight-interfaces-keeping-it-light/16932/11?u=jonah). It has a lot of dependencies, so it's best to use it if present but not force people to install it.~

edit by Rok: 

We can replace vroom with data.table's fread function which makes making vroom optional redundant.

**CmdStanR version number**

0.1.0

jgabry:
fixes #262

#### Submission Checklist

- [x] Run unit tests

#### Summary

Moves vroom from Imports to Suggests and adds code to `read_cmdstan_csv()` to handle the case when vroom is not  installed. Also bumps version to 0.1.1. 

#### Copyright and Licensing

**Columbia**

### DIFF
--- a/R/read_csv.R
+++ b/R/read_csv.R
@@ -105,8 +105,8 @@
 #' }
 #'
 read_cmdstan_csv <- function(files,
-                            variables = NULL,
-                            sampler_diagnostics = NULL) {
+                             variables = NULL,
+                             sampler_diagnostics = NULL) {
   checkmate::assert_file_exists(files, access = "r", extension = "csv")
   metadata <- NULL
   warmup_draws <- NULL
@@ -122,6 +122,15 @@ read_cmdstan_csv <- function(files,
   col_select <- NULL
   not_matching <- c()
   vroom_warnings <- 0
+
+  no_vroom <- !requireNamespace("vroom", quietly = TRUE)
+  if (no_vroom && !identical(variables, "")) {
+    warning(
+      "The vroom package is not installed. Using utils::read.csv() instead. ",
+      "\nInstall the vroom package to speed up reading CmdStan output into R.",
+      call. = FALSE
+    )
+  }
 
   for (output_file in files) {
     if (is.null(metadata)) {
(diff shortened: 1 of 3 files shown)


R/read_csv.R
```diff
@@ -123,6 +123,15 @@ read_cmdstan_csv <- function(files,
   not_matching <- c()
   vroom_warnings <- 0
 
+  no_vroom <- !requireNamespace("vroom", quietly = TRUE)
+  if (no_vroom && !identical(variables, "")) {
+    warning(
```
jgabry, line 128: Only throw a warning about missing vroom if reading in variables. This avoids a warning during model fitting because only a few diagnostics are automatically read in. The user will see the warning the first time they access any variables. Other suggestions welcome.

rok-cesnovar: This should also check sampler diagnostics are "", I think.

rok-cesnovar: I also think if vroom is not present, we should just return everything. I don't think it makes sense to read the entire file in RAM just to then discard some of it and then do it again. Peak RAM usage in this case will be worse than if we warn with something like "no vroom, reading all columns".

jgabry:
> * without vroom we are not able to write GQ in cases where the CSV no longer exists (https://github.com/stan-dev/cmdstanr/blob/master/R/data.R#L138). We should either stop here or use write.csv (I think the former is fine as this case is very rare).

Good point! I totally forgot about that.

> * validation after sampling should be turned off by default when not using vroom. In that case validation reads all samples in memory. With large CSV files, there will be a dead period between when sampling finishes and before the fit object is returned. If we do not want to turn it off by default, we should at least give the user a message that cmdstanr is reading in the CSV files.

I see your point but if we turn off validation then in that case we won't have any divergence warnings for people not using vroom, right? 

> * when no vroom, read_cmdstan_csv should just read everything in always (except if called with variables = "" and sampler_diagnostics = "". utils::read.csv reads everything in memory, no use of discarding it then. If the user is low on RAM, they can discard it themselves.

Yeah good point. 

> * can we tell the user on load that in order to get faster I/O they should install vroom. 

Yeah we can have a message about this when users do `library(cmdstanr)` but that message won't display for people who are using cmdstanr without calling `library` (e.g. as the backend to brms). So it may still be important to have the message when reading the csv. What do you think? 

* Because the way we have it now, users that wont use filtering of variables wont even know there is a faster option. 

Right now it's only if `variables = ""` (no variables) that the message is ignored. There is still a message if `variables = NULL` (all variables).

rok-cesnovar:
> Right now it's only if variables = "" (no variables) that the message is ignored. There is still a message if variables = NULL (all variables).

Oh ok, misread that then. Ignore the comments on the warning message then.

> I see your point but if we turn off validation then in that case we won't have any divergence warnings for people not using vroom, right?

Yeah. Maybe we don't need to change this. Users that care about execution times will probably take the time to install vroom anyways.

Issue #259, opened 2020-07-29
Prepend PATH with path to RTools on Windows


rok-cesnovar:
We have seen a few issues from users that turned out to be muliple g++ installed on the system. Typically some MinGw install or something along those lines.

What we could do is try to prepend the path to rtools on the front of PATH (like we do for TBB) which would then force the use of the right compiler.

We would check the usual paths to rtools and if found use it. If the user installed it in some non-typical location, then we would not prepend it.

jgabry:
I don't know much about this topic, but that sounds ok to me if you think it's a good idea

andrjohns:
Not sure whether is what you're talking about, but for Windows there's an ```RTOOLS40_HOME``` environment variable that gives the install directory for Rtools (I believe the equivalent for version 3.5 and lower is ```RTOOLS_HOME```), which could simplify finding the right compiler

rok-cesnovar:
Good call. `RTOOLS40_HOME` is definitely something we could rely on. We do have to double-check if this gets set for all RTools 4.0 installs. Hopefully, there is no way of not setting it. If that is the case, then it should be easy for R 4.0.

rok-cesnovar:
Ok, silly me. As @andrjohns pointed out, prepending PATH is actually what RTools 4.0 suggests upon installing 

`writeLines('PATH="${RTOOLS40_HOME}\\usr\\bin;${PATH}"', con = "~/.Renviron")`

So maybe we should just check if the user did that? And if they did not, do it for them or suggest they do it themselves?

jgabry:
Sounds good to me

PR #264, opened 2020-08-01
add replacement knitr engine for stan

### LINKED ISSUE #261: knitr engine for Stan chunks in R Markdown documents
Hello! The built-in knitr engine for Stan chunks relies on RStan, which may not work for everyone (especially not for CmdStanR users). I wrote a version of the engine that uses CmdStanR instead of RStan (and supports executable caching to prevent re-compiling every time the Rmd doc is knit/rendered) and demonstrate it here: https://mpopov.com/blog/2020/07/30/replacing-the-knitr-engine-for-stan/

I'm curious if the maintainers would be interested in providing something like this inside the package? Then R Markdown users can override the built-in, RStan-based engine by running something like `cmdstanr::register_knitr_engine()` at the top so whenever they have Stan chunks, those would go through CmdStanR's engine instead.

Here's a {distill} article as another example:

distill

I'd be happy to submit a PR, just wanted to check first.

bearloga:
## Submission Checklist

- [x] Run unit tests

#### Summary

Resolves #261 

**Adds**:
- `eng_cmdstan()` for proceessing `stan` chunks in R Markdown documents
- `register_knitr_engine()`, which registers `eng_cmdstan()` as the `stan` engine to use
- a vignette demonstrating this new engine

**Modifies**:
- `write_stan_tempfile()` now includes an optional dir parameter, which is set to `tempdir()` by default for backwards compatibility

#### Copyright and Licensing

**Mikhail Popov**


jgabry:
Awesome, thanks for submitting this! I should be able to review it this weekend or early in the week.

R/knitr.R
```diff
@@ -0,0 +1,54 @@
...
+eng_stan <- function(options) {
+  output_var <- options$output.var
+  if (!is.character(output_var) || length(output_var) != 1L)
+    stop(
+      "the chunk option output.var must be a character string ",
+      "providing a name for the returned `CmdStanModel` object."
+    )
+  if (options$eval) {
+    if (options$cache)
+      dir <- knitr:::valid_path(options[["cache.path"]], options$label)
```
jgabry, line 44: Thinking ahead to CmdStanR v1.0 (when we want to submit to CRAN), I think CRAN will have a problem with calling a non-exported function from another package with `:::`. But `knitr:::valid_path()` looks like it's super simple internally, so you could avoid having to call it pretty easily.

R/knitr.R
```diff
@@ -0,0 +1,54 @@
...
+    else
+      dir <- tempdir()
+    model_file <- write_stan_tempfile(options$code, dir)
+    csm <- cmdstan_model(model_file)
+    assign(output_var, csm, envir = knitr::knit_global())
+  }
+  options$engine <- "stan"
+  code <- knitr:::one_string(options$code)
```
jgabry, line 52: Same comment as above about calling an internal knitr function

R/knitr.R
```diff
@@ -0,0 +1,54 @@
...
+#' engine for `stan` chunks, replacing knitr's built-in, RStan-based engine. If
+#' `override = FALSE`, this registers a `cmdstan` engine so that both engines
+#' may be used in the same R Markdown document. If there is syntax highlighting,
+#' the `cmdstan` chunks will have `stan` syntax highlighting applied to them.
+#' @references
+#' - [Register a custom language engine](https://bookdown.org/yihui/rmarkdown-cookbook/custom-engine.html)
+#' - [Stan language engine](https://bookdown.org/yihui/rmarkdown/language-engines.html#stan)
+#' @export
+register_knitr_engine <- function(override = TRUE) {
+  if (requireNamespace("knitr", quietly = TRUE)) {
```
jgabry, line 18: I think we should also have 
```r
else {
  stop("Please install the knitr package.")
}
```
otherwise this will just fail silently if knitr isn't installed, right? 

Or maybe instead of that just

```r
if (!requireNamespace("knitr", quietly = TRUE)) {
  stop("Please install the knitr package.")
}
```
at the top of the function and then there's no need to have rest of the body of the function inside an `if` clause?

R/knitr.R
```diff
@@ -0,0 +1,54 @@
+#' Register the knitr engine for Stan
+#'
+#' Registers CmdStanR's knitr engine [eng_stan()] for processing Stan chunks.
```
jgabry, line 3: Can you mention in the doc that there's a vignette with an example?

R/knitr.R
bearloga: Good call! I went with the second suggestion.

R/knitr.R
bearloga: Sure thing!

R/knitr.R
bearloga: I forgot about that requirement, that's right. Good catch! Fixed.

bearloga:
@jgabry You're very welcome! :D Thanks for the review, suggestions, and changes!

R/example.R
```diff
@@ -121,8 +124,11 @@ print_example_program <-
 #' f2 <- write_stan_tempfile(lines)
 #' identical(readLines(f), readLines(f2))
 #'
-write_stan_tempfile <- function(code) {
-  tmp <- tempfile(fileext = ".stan")
+write_stan_tempfile <- function(code, dir = tempdir()) {
```
rok-cesnovar, line 127: Just wanted to chime in to say thanks for this! I kept forgetting to do a feature request issue for this.

bearloga:
> I think this is ready to go and could be merged now, so I'm approving it, but before I merge it I have one question that just occurred to me: @bearloga what do you think about using the name `eng_cmdstan` instead of `eng_stan`? `eng_stan` is what knitr already calls its Stan engine (at least internally). Or did you intentionally use the same name, which could also make sense?

That makes a lot of sense, especially considering what the engine gets registered as when `override = FALSE`, so I'll go ahead and change that.

Thank you for the credit in DESCRIPTION and those "super knitpicky edits"!

rok-cesnovar:
It would probably be good to mention this additions in News.md right?

jgabry:
@bearloga Great, thanks for changing to `eng_cmdstan`. Looks good. 

@rok-cesnovar Yeah we should mention this in news. I'll add that before merging. That probably means we should do a minor version bump. Do you want to do this as `0.2.0` or `0.1.1`?

rok-cesnovar:
I think we should do 0.1.1. We can do 0.2.0 once we merge #263 and fix #259

jgabry:
Ok I bumped the version to 0.1.1 and added an entry to NEWS.md. I will merge this and then update the CmdStanR website. Once the website is updated it will create a url for the new vignette and then I can update the doc to actually link to the vignette instead of just mentioning it.

bearloga:
So sorry for forgetting to update that `%\VignetteIndexEntry`! Wish I could say that was my first time forgetting to update that when renaming a vignette :(

jgabry:
> So sorry for forgetting to update that `%\VignetteIndexEntry`! Wish I could say that was my first time forgetting to update that when renaming a vignette :(

Haha, happens to me all the time!

jgabry:
Ok everything is passing. Merging now. Thanks again for the PR!

bearloga:
You're welcome! Thanks for merging and for a positive PR review experience!

P.S. @rok-cesnovar I'm glad I could unintentionally resolve two issues with one PR :)

jgabry:
Just updated the website. New vignette is live: https://mc-stan.org/cmdstanr/articles/r-markdown.html

PR #273, opened 2020-08-11
Message at startup if new CmdStan is available

### LINKED ISSUE #265: Let users know if a new CmdStan is available
@mitzimorris and I were talking about this. At load time we could compare the latest installed cmdstan to the latest available cmdstan and let the user know if a newer release is available. (If we do this we should make sure to skip this check or wrap it in `try()` if there's no internet connection.)

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #265 

When the package is loaded via library it now checks what the latest CmdStan version is to see if it's more recent than the version being used. If it is then there is a message alerting the user (see example below). This slightly but noticeably increase load time because it has to check GitHub. Also, if there is an error checking the version number from GitHub then there is no message.  

#### Example 

```
> library(cmdstanr)

Loading cmdstanr
This is cmdstanr version 0.1.1
- Online documentation and vignettes at mc-stan.org/cmdstanr
- CmdStan path set to: /Users/jgabry/.cmdstanr/cmdstan-2.23.0
- Use set_cmdstan_path() to change the path

A newer version of CmdStan is available. See ?install_cmdstan() to install it.
```

#### Copyright and Licensing

**Columbia University**

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -16,3 +16,31 @@
 }
 
 
+cmdstanr_initialize <- function() {
+  # First check for environment variable CMDSTAN, but if not found
+  # then see if default
+  path <- Sys.getenv("CMDSTAN")
+  if (isTRUE(nzchar(path))) { # CMDSTAN environment variable found
+    if (dir.exists(path)) {
+      path <- absolute_path(path)
+      suppressMessages(set_cmdstan_path(path))
+    } else {
+      warning("Can't find directory specified by environment variable",
+              " 'CMDSTAN'. Path not set.", call. = FALSE)
+      .cmdstanr$PATH <- NULL
+    }
+
+  } else { # environment variable not found
+    path <- cmdstan_default_path()
+    if (!is.null(path)) {
+      suppressMessages(set_cmdstan_path(path))
+    }
+  }
+
+  if (getRversion() < '3.5.0') {
+    .cmdstanr$TEMP_DIR <- tempdir()
+  } else {
+    .cmdstanr$TEMP_DIR <- tempdir(check = TRUE)
+  }
+  invisible(TRUE)
+}
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,6 +1,14 @@
 .onAttach <- function(...) {
-  ver <- utils::packageVersion("cmdstanr")
-  packageStartupMessage("This is cmdstanr version ", ver)
+  startup_messages()
+}
+
+.onLoad <- function(...) {
+  cmdstanr_initialize()
+}
+
+
+startup_messages <- function() {
+  packageStartupMessage("This is cmdstanr version ", utils::packageVersion("cmdstanr"))
   packageStartupMessage("- Online documentation and vignettes at mc-stan.org/cmdstanr")
   if (is.null(.cmdstanr$PATH)) {
     packageStartupMessage("- Use set_cmdstan_path() to set the path to CmdStan")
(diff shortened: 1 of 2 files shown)


R/zzz.R
```diff
@@ -9,10 +17,43 @@
...
+      && latest_version > cmdstan_version()) {
+      packageStartupMessage(
+        "\nA newer version of CmdStan is available.",
+        " See ?install_cmdstan() to install it."
+      )
+  }
 }
 
 
+cmdstanr_initialize <- function() {
```
jgabry, line 32: This function is all the same, I just moved it to zzz.R from path.R since it's only used in `.onLoad()`

R/zzz.R
```diff
@@ -9,10 +17,43 @@
...
-.onLoad <- function(...) {
-  cmdstanr_initialize()
+  latest_version <- try(suppressWarnings(latest_released_version()), silent = TRUE)
+  if (!inherits(latest_version, "try-error")
+      && latest_version > cmdstan_version()) {
+      packageStartupMessage(
+        "\nA newer version of CmdStan is available.",
+        " See ?install_cmdstan() to install it."
+      )
+  }
```
jgabry, line 28: These lines are where the relevant changes are.

jgabry:
> This looks good to me.
> 
> I have two questions that can also be addressed separately.
> 
> 1. I am not sure if there is a way, but is there a typical way of setting some global property for packages in R that the packages can consume later? Is that the `options()` function? Or should that be R environment variables?
> 
> I am basically thinking if there is a way to set a something
> 
> ```r
> options("CMDSTANR_NO_VER_CHECK"=TRUE)
> library(cmdstanr)
> ```
> 
> and cmdstanr would not check versions on boot.
> 
> Just thinking out loud here. A user could also set it in ~/.Renviron. This should not be recommended but if someone explicitly wants this, should we provide a way of turning that off?
> 

How about we use `options()` with the default value taken from an environment variable? So if someone has `CMDSTANR_NO_VER_CHECK` set (in .Renviron or just set in their R session) then we can use that as the default value 
for `options("CMDSTANR_NO_VER_CHECK")`? If not set then we'll default to checking for the a new version. What do you think? 


> 2. what will happen with packages that use cmdstanr with `cmdstanr::`. The checks should probably not happen there right? Just double checking.

Yeah because I put this in `.onAttach()` instead of `.onLoad()` it will only happen when someone does `library(cmdstanr)`. So, e.g., users of brms with the cmdstanr backend will not see this message.

rok-cesnovar:
> How about we use options() with the default value taken from an environment variable? 

Cool with that.

jgabry:
Ok cool. I guess I should add a line to the message about turning off the check using options or the environment variable? How about this message (see last line)?

```
Loading cmdstanr
This is cmdstanr version 0.1.1
- Online documentation and vignettes at mc-stan.org/cmdstanr
- CmdStan path set to: /Users/jgabry/.cmdstanr/cmdstan-2.23.0
- Use set_cmdstan_path() to change the path

A newer version of CmdStan is available. See ?install_cmdstan() to install it.
To disable this check in the future set the option or environment variable CMDSTANR_NO_VER_CHECK=TRUE.
```

rok-cesnovar:
I like that.

jgabry:
Ok I just added that. I'm not sure how to add a test for it though. I did test it locally and it seems to work as expected.

rok-cesnovar:
Yeah, difficult to test this one. Looks fine to me.

PR #277, opened 2020-08-14
add $syntax_check() and tests

### LINKED ISSUE #276: Separate C++ translation from compilation
For running tests on the generated Stan code in brms, I would like to use the syntax checking of the compiler to at least catch any errors at compilation time. In rstan, it is possible to do translation to C++ (including syntax checking) without actual C++ compilation. It seems that cmdstanr's `$compile()` always does both at the same time. Is it possible to separate these two processes and make them available in separate functions?

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #276 

@paul-buerkner if you have a few minutes, you can test with

`remotes::install_github("stan-dev/cmdstanr@syntax_check")
`
```r
library(cmdstanr)

file_path <- file.path(cmdstan_path(), "examples", "bernoulli", "bernoulli.stan")
mod <- cmdstan_model(file_path, compile = FALSE)
mod$syntax_check()
```

#### Copyright and Licensing

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
 }
 CmdStanModel$set("public", name = "compile", value = compile_method)
 
+#' Check syntax of a Stan program
+#'
+#' @name model-method-syntax-check
+#' @aliases syntax_check
+#' @family CmdStanModel methods
+#'
+#' @description The `$syntax_check()` method of a [`CmdStanModel`] object
+#'   checks the Stan syntax of the model and returns `TRUE` if the compilation
+#'   succeeds. In case of a syntax error, an error is thrown.
+#'   
+#' @section Usage:
+#'   ```
+#'   $syntax_check(
+#'     quiet = TRUE,
+#'     include_paths = NULL,
+#'     stanc_options = list()
+#'   )
+#'   ```
+#'
+#' @section Arguments:
+#'   * `quiet`: (logical) Should the compile command be printed.
+#'   * `include_paths`: (character vector) Paths to directories where Stan
+#'   should look for files specified in `#include` directives in the Stan
+#'   program.
+#'   * `stanc_options`: (list) Any Stan-to-C++ transpiler options to be used
+#'   when compiling the model. See the **Examples** section below as well as the
+#'   `stanc` chapter of the CmdStan Guide for more details on available options:
+#'   https://mc-stan.org/docs/cmdstan-guide/stanc.html.
+#'
+#' @section Value: The `$syntax_check()` method returns `TRUE` if the model
+#'   is valid.
+#'
+#' @template seealso-docs
+#'
+#' @examples
+#' \dontrun{
+#' file <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")
+#'
+#' mod <- cmdstan_model(file, compile = FALSE)
+#' mod$syntax_check()
+#' }
+#'
+NULL
+
+syntax_check_method <- function(quiet = TRUE,
+                           include_paths = NULL,
+                           stanc_options = list()) {
+  if (length(stanc_options) == 0 && !is.null(private$precompile_stanc_options_)) {
+    stanc_options <- private$precompile_stanc_options_
+  }
+  if (is.null(include_paths) && !is.null(private$precompile_include_paths_)) {
+    include_paths <- private$precompile_include_paths_
+  }
+  
+  model_name <- sub(" ", "_", paste0(strip_ext(basename(self$stan_file())), "_model"))
+
+  message("Running syntax check of the Stan program...")
+
+  temp_hpp_file <- tempfile(pattern = "model-", fileext = ".hpp")
+  stanc_options[["o"]] <- temp_hpp_file
+
+  stancflags_val <- ""
+  if (!is.null(include_paths)) {
+    checkmate::assert_directory_exists(include_paths, access = "r")
+    include_paths <- absolute_path(include_paths)
+    include_paths <- paste0(include_paths, collapse = ",")
+    if (cmdstan_version() >= "2.24") {
+      include_paths_flag <- " --include-paths="
+    } else {
+      include_paths_flag <- " --include_paths="
+    }
+    stancflags_val <- paste0(stancflags_val, include_paths_flag, include_paths, " ")
+  }
+
+  if (is.null(stanc_options[["name"]])) {
+    stanc_options[["name"]] <- model_name
+  }
+
+  stanc_built_options = c()
+  for (i in seq_len(length(stanc_options))) {
+    option_name <- names(stanc_options)[i]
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thanks a lot! Will check it out tonight when I am back on my laptop.

Codecov Comments Bot <notifications@github.com> schrieb am Fr., 14. Aug.
2020, 17:05:

jgabry:
Awesome. I'll try to review this later today or this weekend.

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
 
+#' Check syntax of a Stan program
+#'
+#' @name model-method-syntax-check
+#' @aliases syntax_check
+#' @family CmdStanModel methods
+#'
+#' @description The `$syntax_check()` method of a [`CmdStanModel`] object
+#'   checks the Stan syntax of the model and returns `TRUE` if the compilation
+#'   succeeds. In case of a syntax error, an error is thrown.
```
jgabry, line 517:
```suggestion
#' @description The `$syntax_check()` method of a [`CmdStanModel`] object
#'   checks the Stan program for syntax errors and returns `TRUE` (invisibly) if 
#'   parsing succeeds. In case of a syntax error, an error is thrown.
```

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
+#'   
+#' @section Usage:
+#'   ```
+#'   $syntax_check(
+#'     quiet = TRUE,
```
jgabry, line 522: Unlike some of the other methods with a `quiet` argument, I'm not sure we need it here. Is there a case when it's helpful to see the command here? If not, I'd suggest removing the argument. If there is, then it's fine to keep it.

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
+#'     include_paths = NULL,
+#'     stanc_options = list()
+#'   )
+#'   ```
+#'
+#' @section Arguments:
+#'   * `quiet`: (logical) Should the compile command be printed.
```
jgabry, line 529: See comment above about `quiet`

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
+#'   * `include_paths`: (character vector) Paths to directories where Stan
+#'   should look for files specified in `#include` directives in the Stan
+#'   program.
+#'   * `stanc_options`: (list) Any Stan-to-C++ transpiler options to be used
+#'   when compiling the model. See the **Examples** section below as well as the
+#'   `stanc` chapter of the CmdStan Guide for more details on available options:
+#'   https://mc-stan.org/docs/cmdstan-guide/stanc.html.
```
jgabry, line 536: The **Examples** section referred to here is actually in the doc for the compile method. Maybe change to: 

```suggestion
#'   * `stanc_options`: (list) Any Stan-to-C++ transpiler options to be used
#'   when compiling the model. See the documentation for the
#'   [`$compile()`][model-method-compile] method for details.
```

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
+#'
+#' @section Value: The `$syntax_check()` method returns `TRUE` if the model
```
jgabry, line 538:
```suggestion
#' @section Value: The `$syntax_check()` method returns `TRUE` (invisibly) if the model
```

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
...
+
+  stanc_built_options = c()
+  for (i in seq_len(length(stanc_options))) {
+    option_name <- names(stanc_options)[i]
+    if (isTRUE(as.logical(stanc_options[[i]]))) {
+      stanc_built_options <- c(stanc_built_options, paste0("--", option_name))
+    } else {
+      stanc_built_options <- c(stanc_built_options, paste0("--", option_name, "=", stanc_options[[i]]))
+    }
+  }
```
jgabry, line 595: If possible can you add a test just making sure that this works or at least doesn't error?

R/model.R
```diff
@@ -506,6 +506,114 @@ compile_method <- function(quiet = TRUE,
 }
 CmdStanModel$set("public", name = "compile", value = compile_method)
 
+#' Check syntax of a Stan program
+#'
+#' @name model-method-syntax-check
```
jgabry, line 511:
```suggestion
#' @name model-method-syntax_check
```

paul-buerkner:
This feature works well! Just one minor question: It seems that the initial message "Syntax checking the Stan program..." is passed via the same stream as the details of the error message if an error occurs. As a result, I cannot turn of the initial message separately. Is the message really required?

rok-cesnovar:
> Is this message really required?

Yeah, maybe not. The syntax check is instant (if a model had a few thousand lines it would maybe take a second.

Would this work: text that the syntax is OK on stdout and a compile error printed as message()

paul-buerkner:
I think if OK was printed as message, that would also be fine and more consistent, but given the instant execution time, I thought the initial message may not be required.

paul-buerkner:
Alternatively, perhaps there could be an argument to silence these messages? For use in brms, it would be the best to say nothing when it worked and throw in the information when it did. I don't want to make these too much about brms though so you go with the approach you think is best for cmdstanr in general.

rok-cesnovar:
I think we can reuse quiet for this case. 

quiet = TRUE-> Only show the compile error, otherwise just return TRUE 
quiet  = FALSE -> print any other messages

Will wait for @jgabry to comment on that before I proceed.

> I don't want to make these too much about brms

I think the points that you make about the use in brms will apply for other packages that will use cmdstanr. These are all very good points I think.

jgabry:
> quiet = TRUE-> Only show the compile error, otherwise just return TRUE
> quiet = FALSE -> print any other messages

That sounds good to me!

R/model.R
rok-cesnovar: stanc_options are used always for at least the name, so this works and is already being tested. Will add a test where we specify an additional flag.

rok-cesnovar:
This is actually a nice and fast way of checking a model with the pedantic check without having to compile to the C++ level. So thanks @paul-buerkner for the idea! 

We should use this in a vignette to show how users can run the pedantic check in cmdstanr.

paul-buerkner:
Glad it turns out to be useful outside of brms as well!

R/model.R
rok-cesnovar: quiet = FALSE now outputs "Stan program is syntactically correct"
quiet = TRUE only outputs anything on stderr from the compiler

rok-cesnovar:
@jgabry I think I addressed everything. I added a test for the pedantic mode so we test additional flags. Also added the pedantic mode example to the vignette. Not sure if I need to rebuild the vignette or anything. Anyhow, ready for another look.

PR #295, opened 2020-08-30
Fix macOS CI issue

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This fixes the occasional but annoying macOS CI false positive. 

It turns out the problem is with sampling with 100k iterations we had in one test. Reducing that to 10k fixes the occasional CI failures on MacOS. I ran ~50 tests on my fork and got zero fails. There might be other cases where we are close to the memory limit on the mac agents and we will see those pop-up now. But we now at least know what to look at and that these are in fact false positives.

The segfault that occurred happened in the test where we read the 100k draws in memory. I am assuming the virtual machine runs out of memory. If we really want to test with 100k iterations, the fit would have to be split in a separate test file. I don't think it is necessary to test for 100k so I just reduced it to 10k.

#### Copyright and Licensing

### DIFF
--- a/tests/testthat/test-fit-mcmc.R
+++ b/tests/testthat/test-fit-mcmc.R
@@ -12,7 +12,7 @@ if (not_on_cran()) {
                             refresh = 0, save_warmup = TRUE)
   fit_mcmc_2 <- testing_fit("logistic", method = "sample",
                             seed = 1234, chains = 1,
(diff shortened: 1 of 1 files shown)


jgabry:
Awesome, thanks for figuring this out!

Issue #304, opened 2020-10-01
Ability to cache compiled stan-models and re-use them across sessions


mikekaminsky:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to save the compilation time for models that we run over-and-over-again across R sessions.

**Describe the solution you'd like**
Ideally,  we could cache those executables somewhere and skip compilation if the executable has always been compiled once.  I can see the paths to the executables and header files via `model$exe_file()` and `model$hpp_file()`, but what I don't see is a way to create a `CmdStanModel` object using an existing executable / header file. If this is already possible and I just missed it in the docs, I apologize!

**Describe alternatives you've considered**
I'm basically trying to do a small part of what [rstantools](https://github.com/stan-dev/rstantools) does for [rstan](https://github.com/stan-dev/rstan) -- for a variety of reasons, our project doesn't work with rstan and so we are using cmdstanr instead -- I've got the shell of the code working for doing the caching, just need a way to plug the saved executables back into the code we're using (creating a `CmdStanModel` object).

mitzimorris:
cmdstanpy lets you instantiate a CmdStanModel from just an exe file - also has logic to check timestamps on stan and exe file and if the latter is newer, doesn't compile.  CmdStanR should have similar behaviors.

rok-cesnovar:
Hi, thanks for the request.

We currently do not have a way to create a CmdStanModel just from an executable. It might be reasonable to add that.

However, the executable should not recompile every session. Can you maybe provide an example of how you call cmdstan_model?

If I run
```
library(cmdstanr)


model_path <- file.path(cmdstan_path(), "examples", "bernoulli", "bernoulli.stan")
mod <- cmdstan_model(model_path)
```
close the session, restart RStudio or the PC, it does not recompile every time.

rok-cesnovar:
We compare the modified time of the .stan file and the executable and only recompile if the stan file was modified after the executable was created.

jgabry:
Yeah like @rok-cesnovar said I'm not sure that it's necessary to be able to create a CmdStanModel object from an executable if `cmdstan_model()` can be used without recompiling. That seems like it's just as convenient if not more convenient (because the user doesn't need to worry themselves about the executable or its path and whatnot). Although maybe there's something I'm overlooking. 

@mikekaminsky is that not working properly for you? If it's not then we definitely need to fix something. 
[...]

PR #306, opened 2020-10-01
Recompile based on the hashed code instead of file mod time

mike-lawrence:
#### Submission Checklist

- [X] Run unit tests

#### Summary

Per discussion in #304

#### Copyright and Licensing

Mike Lawrence

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -448,7 +448,6 @@ run_autoformatter_method <- function(quiet = FALSE,
   arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='stanc',include_paths=include_paths,
     stanc_options=stanc_options)
 
-  #given bug in stanc (https://github.com/stan-dev/cmdstan/issues/944),
   # if pedantic-mode warnings have been requested, first run without --auto-format & with echo=!quiet
   if('--warn-pedantic' %in% arg_strings){
     pedantic_run_log <- processx::run(
--- a/R/model.R
+++ b/R/model.R
@@ -486,7 +486,17 @@ compile_method <- function(quiet = TRUE,
   } else {
     exe_base <- file.path(dir, basename(self$stan_file()))
   }
-  current_hash <- digest::digest(self$code(),algo='xxhash64')
+
+  # obtain the current hash from the auto-formatted code
+  stanc3_formatted_code <- processx::run(
+      command = stanc_cmd(),
+      args = c(self$stan_file(), "--auto-format"),
+      wd = cmdstan_path(),
+      echo = FALSE
+  )$stdout
+  current_hash <- digest::digest(stanc3_formatted_code,algo='xxhash64')
+
+  #append hash to model name
   exe_name_without_hash <- paste0(strip_ext(exe_base),exe_suffix)
   exe <- cmdstan_ext(paste0(exe_name_without_hash,'_',current_hash))
   model_name <- sub(" ", "_", paste0(strip_ext(basename(self$stan_file())), "_model"))
--- a/R/model.R
+++ b/R/model.R
@@ -812,6 +812,7 @@ sample_method <- function(data = NULL,
                           fixed_param = FALSE,
                           validate_csv = TRUE,
                           show_messages = TRUE,
+                          chain_id = seq_len(chains),
                           # deprecated
                           cores = NULL,
                           num_cores = NULL,
@@ -865,6 +866,7 @@ sample_method <- function(data = NULL,
   checkmate::assert_integerish(chains, lower = 1, len = 1)
   checkmate::assert_integerish(parallel_chains, lower = 1, null.ok = TRUE)
   checkmate::assert_integerish(threads_per_chain, lower = 1, len = 1, null.ok = TRUE)
+  checkmate::assert_integerish(chain_id, lower = 1, len = chains, unique = TRUE, null.ok = FALSE)
   if (is.null(self$cpp_options()[["stan_threads"]])) {
     if (!is.null(threads_per_chain)) {
       warning("'threads_per_chain' is set but the model was not compiled with ",
@@ -901,7 +903,7 @@ sample_method <- function(data = NULL,
     method_args = sample_args,
     model_name = strip_ext(basename(self$exe_file())),
     exe_file = self$exe_file(),
-    proc_ids = seq_len(chains),
+    proc_ids = chain_id,
     data_file = process_data(data),
     save_latent_dynamics = save_latent_dynamics,
     seed = seed,
--- a/R/model.R
+++ b/R/model.R
@@ -674,7 +674,8 @@ CmdStanModel$set("public", name = "check_syntax", value = check_syntax_method)
 #'     window = NULL,
 #'     fixed_param = FALSE,
 #'     validate_csv = TRUE,
-#'     show_messages = TRUE
+#'     show_messages = TRUE,
+#'     chain_id = seq_len(chains)
 #'   )
 #'   ```
 #'
--- a/R/model.R
+++ b/R/model.R
@@ -917,7 +917,7 @@ sample_method <- function(data = NULL,
   )
   cmdstan_args <- CmdStanArgs$new(
     method_args = sample_args,
-    model_name = strip_ext(basename(self$exe_file())),
+    model_name = strip_ext(basename(self$stan_file())),
     exe_file = self$exe_file(),
     proc_ids = chain_ids,
     data_file = process_data(data),
@@ -1014,7 +1014,7 @@ optimize_method <- function(data = NULL,
   )
   cmdstan_args <- CmdStanArgs$new(
     method_args = optimize_args,
-    model_name = strip_ext(basename(self$exe_file())),
+    model_name = strip_ext(basename(self$stan_file())),
     exe_file = self$exe_file(),
     proc_ids = 1,
     data_file = process_data(data),
@@ -1137,7 +1137,7 @@ variational_method <- function(data = NULL,
   )
   cmdstan_args <- CmdStanArgs$new(
     method_args = variational_args,
-    model_name = strip_ext(basename(self$exe_file())),
+    model_name = strip_ext(basename(self$stan_file())),
     exe_file = self$exe_file(),
     proc_ids = 1,
     data_file = process_data(data),
@@ -1251,7 +1251,7 @@ generate_quantities_method <- function(fitted_params,
   )
   cmdstan_args <- CmdStanArgs$new(
     method_args = generate_quantities_args,
-    model_name = strip_ext(basename(self$exe_file())),
+    model_name = strip_ext(basename(self$stan_file())),
     exe_file = self$exe_file(),
     proc_ids = seq_len(chains),
     data_file = process_data(data),
--- a/R/model.R
+++ b/R/model.R
@@ -1004,7 +1004,7 @@ optimize_method <- function(data = NULL,
     output_dir = output_dir
   )
 
-  cmdstan_procs <- CmdStanProcs$new(num_procs = 1)
+  cmdstan_procs <- CmdStanProcs$new(num_procs = 1, show_messages = (is.null(refresh) || refresh != 0))
   runset <- CmdStanRun$new(args = cmdstan_args, procs = cmdstan_procs)
   runset$run_cmdstan()
   CmdStanMLE$new(runset)
@@ -1124,7 +1124,7 @@ variational_method <- function(data = NULL,
     output_dir = output_dir
   )
 
-  cmdstan_procs <- CmdStanProcs$new(num_procs = 1)
+  cmdstan_procs <- CmdStanProcs$new(num_procs = 1, show_messages = (is.null(refresh) || refresh != 0))
   runset <- CmdStanRun$new(args = cmdstan_args, procs = cmdstan_procs)
   runset$run_cmdstan()
   CmdStanVB$new(runset)
--- a/R/model.R
+++ b/R/model.R
@@ -571,9 +571,6 @@ compile_method <- function(quiet = TRUE,
   if (!force_recompile) {
     message("Model executable is up to date!")
     private$cpp_options_ <- cpp_options
-    # private$precompile_cpp_options_ <- NULL
-    # private$precompile_stanc_options_ <- NULL
-    # private$precompile_include_paths_ <- NULL
     self$exe_file(exe)
     return(invisible(self))
   } else {
--- a/R/model.R
+++ b/R/model.R
@@ -452,8 +452,8 @@ run_autoformatter_method <- function(quiet = FALSE,
     command = stanc_cmd(),
     args = c(self$stan_file(), arg_strings, "--auto-format"),
     wd = cmdstan_path(),
-    echo_cmd = !quiet,
-    echo = !quiet,
+    echo_cmd = FALSE,
+    echo = FALSE,
     spinner = quiet && interactive(),
     stderr_line_callback = function(x,p) {
       message(x)
--- a/R/model.R
+++ b/R/model.R
@@ -402,11 +402,6 @@ generate_arg_strings_method <- function(quiet=TRUE,
   if (is.null(stanc_options[["name"]])) {
     stanc_options[["name"]] <- model_name
   }
-  if (make_or_stanc == "make") {
-    char_for_stanc_built_options <- "'"
-  } else {
-    char_for_stanc_built_options <- ""
-  }
 
   stanc_built_options <- c()
   for (i in seq_len(length(stanc_options))) {
@@ -414,8 +409,8 @@ generate_arg_strings_method <- function(quiet=TRUE,
     if (isTRUE(as.logical(stanc_options[[i]]))) {
       stanc_built_options <- c(stanc_built_options, paste0("--", option_name))
     } else {
-      stanc_built_options <- c(stanc_built_options, paste0("--", option_name, "=",
-        char_for_stanc_built_options, stanc_options[[i]], char_for_stanc_built_options))
+      stanc_built_options <- c(stanc_built_options, paste0("--", option_name, "='",
+       stanc_options[[i]], "'"))
     }
   }
 
--- a/R/model.R
+++ b/R/model.R
@@ -486,8 +486,8 @@ compile_method <- function(quiet = TRUE,
   } else {
     exe_base <- file.path(dir, basename(self$stan_file()))
   }
-  current_hash = digest::digest(self$code(),algo='xxhash64')
-  exe_name_without_hash = paste0(strip_ext(exe_base),exe_suffix)
+  current_hash <- digest::digest(self$code(),algo='xxhash64')
+  exe_name_without_hash <- paste0(strip_ext(exe_base),exe_suffix)
   exe <- cmdstan_ext(paste0(exe_name_without_hash,'_',current_hash))
   model_name <- sub(" ", "_", paste0(strip_ext(basename(self$stan_file())), "_model"))
 
@@ -497,11 +497,11 @@ compile_method <- function(quiet = TRUE,
   # - the stan model was changed since last compilation
   if (!file.exists(exe)) {
     force_recompile <- TRUE
-    regex_string = paste0('^(',basename(exe_name_without_hash),'_)[a-f0-9]{16}')
-    regex_string = cmdstan_ext(regex_string)
-    regex_string = gsub('(.exe)$','(.exe)',regex_string)
-    regex_string = paste0(regex_string,'$')
-    existing_exes = list.files(
+    regex_string <- paste0('^(',basename(exe_name_without_hash),'_)[a-f0-9]{16}')
+    regex_string <- cmdstan_ext(regex_string)
+    regex_string <- gsub('(.exe)$','(.exe)',regex_string)
+    regex_string <- paste0(regex_string,'$')
+    existing_exes <- list.files(
       path = dirname(exe_name_without_hash)
       , pattern = regex_string
       , full.names = T
--- a/R/model.R
+++ b/R/model.R
@@ -444,7 +444,9 @@ compile_method <- function(quiet = TRUE,
                            #deprecated
                            threads = FALSE) {
 
-  arg_strings = self$generate_arg_strings(make_or_stanc='make',include_paths=include_paths,
+  make_arg_strings = self$generate_arg_strings(make_or_stanc='make',include_paths=include_paths,
+    cpp_options=cpp_options,stanc_options=stanc_options,threads=threads)
+  stanc_arg_strings = self$generate_arg_strings(make_or_stanc='stanc',include_paths=include_paths,
     cpp_options=cpp_options,stanc_options=stanc_options,threads=threads)
 
  # add path to the TBB library to the PATH variable to avoid copying the dll file
@@ -490,7 +492,7 @@ compile_method <- function(quiet = TRUE,
   # obtain the current hash from the auto-formatted code
   stanc3_formatted_code <- processx::run(
       command = stanc_cmd(),
-      args = c(self$stan_file(), "--auto-format"),
+      args = c(self$stan_file(), "--auto-format",stanc_arg_strings),
       wd = cmdstan_path(),
       echo = FALSE
   )$stdout
@@ -540,7 +542,7 @@ compile_method <- function(quiet = TRUE,
 
   run_log <- processx::run(
     command = make_cmd(),
-    args = c(tmp_exe,arg_strings),
+    args = c(tmp_exe,make_arg_strings),
     wd = cmdstan_path(),
     echo_cmd = !quiet,
     echo = !quiet,
--- a/R/model.R
+++ b/R/model.R
@@ -622,6 +622,7 @@ check_syntax_method <- function(quiet = FALSE,
   arg_strings = self$generate_arg_strings(make_or_stanc='stanc',include_paths=include_paths,
     stanc_options=stanc_options)
 
+
   run_log <- processx::run(
     command = stanc_cmd(),
     args = c(self$stan_file(), arg_strings),
--- a/R/model.R
+++ b/R/model.R
@@ -499,7 +499,8 @@ compile_method <- function(quiet = TRUE,
                            threads = FALSE) {
 
   from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
-  current_hash <- digest::digest(from_autoformatter$stdout,algo='xxhash64')
+  formatted_code = gsub(pattern='\r\n',replacement='\n',x=from_autoformatter$stdout,fixed=T)
+  current_hash <- digest::digest(formatted_code,algo='xxhash64')
 
   make_arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='make',include_paths=include_paths,
     cpp_options=cpp_options,stanc_options=stanc_options,threads=threads)
--- a/R/model.R
+++ b/R/model.R
@@ -361,8 +361,8 @@ generate_arg_strings_method <- function(quiet=TRUE,
                            #deprecated
                            threads = FALSE) {
 
-  if (!is.null(make_or_stanc)) {
-    checkmate::assert_choice(make_or_stanc, choices = c("make", "stanc"))
+  if (is.null(make_or_stanc)) {
+    stop('Internal cmdstanr error: make_or_stanc is NULL')
   }
 
   if (length(cpp_options) == 0 && !is.null(private$precompile_cpp_options_)) {
--- a/R/model.R
+++ b/R/model.R
@@ -353,7 +353,7 @@ CmdStanModel <- R6::R6Class(
 NULL
 
 generate_arg_strings_method <- function(quiet=TRUE,
-                           make_or_stanc=list('make','stanc'),
+                           make_or_stanc = c("make", "stanc"),
                            dir = NULL,
                            include_paths = NULL,
                            cpp_options = list(),
@@ -362,7 +362,7 @@ generate_arg_strings_method <- function(quiet=TRUE,
                            threads = FALSE) {
 
   if (!is.null(make_or_stanc)) {
-    checkmate::assert_choice(make_or_stanc, choices=c('make','stanc'))
+    checkmate::assert_choice(make_or_stanc, choices = c("make", "stanc"))
   }
 
   if (length(cpp_options) == 0 && !is.null(private$precompile_cpp_options_)) {
@@ -402,19 +402,19 @@ generate_arg_strings_method <- function(quiet=TRUE,
   if (is.null(stanc_options[["name"]])) {
     stanc_options[["name"]] <- model_name
   }
-  if(make_or_stanc=='make'){
+  if (make_or_stanc == "make") {
     char_for_stanc_built_options <- "'"
-  }else{
+  } else {
     char_for_stanc_built_options <- ""
   }
 
-  stanc_built_options = c()
+  stanc_built_options <- c()
   for (i in seq_len(length(stanc_options))) {
     option_name <- names(stanc_options)[i]
     if (isTRUE(as.logical(stanc_options[[i]]))) {
-      stanc_built_options = c(stanc_built_options, paste0("--", option_name))
+      stanc_built_options <- c(stanc_built_options, paste0("--", option_name))
     } else {
-      stanc_built_options = c(stanc_built_options, paste0("--", option_name, "=",
+      stanc_built_options <- c(stanc_built_options, paste0("--", option_name, "=",
         char_for_stanc_built_options, stanc_options[[i]], char_for_stanc_built_options))
     }
   }
@@ -423,15 +423,15 @@ generate_arg_strings_method <- function(quiet=TRUE,
     prepare_precompiled(cpp_options, quiet)
   }
 
-  if(make_or_stanc=='make'){
+  if (make_or_stanc == "make"){
     stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
-    these_args = c(
+    these_args <- c(
       cpp_options_to_compile_flags(cpp_options),
       stancflags_val
     )
-  }else{
+  } else {
     stancflags_val <- trimws(c(stancflags_val, c(stanc_built_options)))
-    these_args = c(stancflags_val)
+    these_args <- c(stancflags_val)
   }
   return(these_args)
 }
@@ -445,11 +445,16 @@ run_autoformatter_method <- function(quiet = FALSE,
   temp_hpp_file <- tempfile(pattern = "model-", fileext = ".hpp")
   stanc_options[["o"]] <- temp_hpp_file
 
-  arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='stanc',include_paths=include_paths,
-    stanc_options=stanc_options)
+  arg_strings <-
+    self$generate_arg_strings(
+      quiet = quiet,
+      make_or_stanc = "stanc",
+      include_paths = include_paths,
+      stanc_options = stanc_options
+    )
 
   # if pedantic-mode warnings have been requested, first run without --auto-format & with echo=!quiet
-  if('--warn-pedantic' %in% arg_strings){
+  if ("--warn-pedantic" %in% arg_strings) {
     pedantic_run_log <- processx::run(
       command = stanc_cmd(),
       args = c(self$stan_file(), arg_strings),
@@ -497,12 +502,19 @@ compile_method <- function(quiet = TRUE,
                            #deprecated
                            threads = FALSE) {
 
-  from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
-  formatted_code = gsub(pattern='\r\n',replacement='\n',x=from_autoformatter$stdout,fixed=T)
-  current_hash <- digest::digest(formatted_code,algo='xxhash64')
+  from_autoformatter <- self$run_autoformatter(quiet = quiet, include_paths = include_paths, stanc_options = stanc_options)
+  formatted_code <- gsub(pattern = '\r\n',replacement = '\n', x = from_autoformatter$stdout, fixed = TRUE)
+  current_hash <- digest::digest(formatted_code, algo = "xxhash64")
 
-  make_arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='make',include_paths=include_paths,
-    cpp_options=cpp_options,stanc_options=stanc_options,threads=threads)
+  make_arg_strings <-
+    self$generate_arg_strings(
+      quiet = quiet,
+      make_or_stanc = "make",
+      include_paths = include_paths,
+      cpp_options = cpp_options,
+      stanc_options = stanc_options,
+      threads = threads
+    )
 
  # add path to the TBB library to the PATH variable to avoid copying the dll file
   if (cmdstan_version() >= "2.21" && os_is_windows()) {
@@ -561,9 +573,9 @@ compile_method <- function(quiet = TRUE,
     regex_string <- gsub('(.exe)$','(.exe)',regex_string)
     regex_string <- paste0(regex_string,'$')
     existing_exes <- list.files(
-      path = dirname(exe_name_without_hash)
-      , pattern = regex_string
-      , full.names = T
+      path = dirname(exe_name_without_hash),
+      pattern = regex_string,
+      full.names = TRUE
     )
     file.remove(existing_exes)
   }
@@ -586,7 +598,7 @@ compile_method <- function(quiet = TRUE,
 
   run_log <- processx::run(
     command = make_cmd(),
-    args = c(tmp_exe,make_arg_strings),
+    args = c(tmp_exe, make_arg_strings),
     wd = cmdstan_path(),
     echo_cmd = !quiet,
     echo = !quiet,
@@ -661,7 +673,7 @@ check_syntax_method <- function(quiet = FALSE,
                                 include_paths = NULL,
                                 stanc_options = list()) {
 
-  from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
+  from_autoformatter <- self$run_autoformatter(quiet = quiet, include_paths = include_paths, stanc_options = stanc_options)
   if (!quiet) {
     message("Stan program is syntactically correct");
   }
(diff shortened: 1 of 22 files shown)


mike-lawrence:
Note that I had to add a dependency on the `digest` package for this as I don't think there are any hash functions in base R. Happy to replace with one if I missed it though. Also, I opted for the `xxhash64` as I understand that it is fastest these days.

mike-lawrence:
Oh shoot. I just realized that a bunch of the tests assume the executable filename, so I have to fix that.

jgabry:
Thanks @mike-lawrence! 

Yeah I'm not aware of any functionality in base R for hashing and I like the digest package (it also has basically zero dependencies itself, so that's nice). 

I haven't had a chance to check all the regex stuff yet but the general idea of getting rid of the previous executables is good. Since regular expressions are super confusing it would be good to get more eyes on this. I'll request a review from @rok-cesnovar but I know he's busy working on his thesis and so it might take a little while before this is fully reviewed. 

Also, I'm not 100% sure but I think we have some tests that test recompilation after changes to the Stan program. If we do then those will help us figure out if this is working properly. If we don't have those tests then we should probably add them as part of this PR to make sure that this is working as intended.

R/model.R
```diff
@@ -405,7 +405,9 @@ compile_method <- function(quiet = TRUE,
   } else {
     exe_base <- file.path(dir, basename(self$stan_file()))
   }
-  exe <- cmdstan_ext(paste0(strip_ext(exe_base), exe_suffix))
+  current_hash = digest::digest(self$code(),algo='xxhash64')
+  exe_name_without_hash = paste0(strip_ext(exe_base),exe_suffix)
+  exe <- cmdstan_ext(paste0(exe_name_without_hash,'_',current_hash))
   model_name <- sub(" ", "_", paste0(strip_ext(basename(self$stan_file())), "_model"))
```
rok-cesnovar, line 411: This should replace self$stan_file() with exe_name_without_hash otherwise the model name will be bernoulli_hash_model, which is not what we want.

rok-cesnovar:
> Also, I'm not 100% sure but I think we have some tests that test recompilation after changes to the Stan program.

Those are currently done as: 
 - compile
 - change modify time
 - try compiling again and check that it did

The second step will have to now be changed so it modifies the stan model file (writing two different strings in the model file). @mike-lawrence let me know if you need any help with the tests. Glad to help. Thanks for doing this!

Apart from the test and one minor comment I had this looks great.

rok-cesnovar:
This is the test location: https://github.com/stan-dev/cmdstanr/blob/master/tests/testthat/test-model-compile.R#L58

mike-lawrence:
Hm. The non-test aspects of this might not be ready after all. I've been using it on linux and have been getting left-over executables, so something about the hashing is awry.

rok-cesnovar:
No worries, let me know if you need any help. 

One idea that comes to mind is that we could run the model through the stanc3 auto formatter before hashing it, avoiding recompiling for whitespace/formatting changes. That can also be a separate issue/PR of course.

mike-lawrence:
@rok-cesnovar working on this now, figure it does make sense to include in this PR your idea to not hash the code but instead hash the output of the autoformatter. I notice that check_syntax() generates an .hpp file; I'm a newb when it comes to compiled stuff/C, so would this hpp file be the output of the autoformatter?

rok-cesnovar:
You get autoformatted Stan code if you run `bin/stanc3 --auto-format model.stan`.

So 
```
run_vals <- processx::run(
    command = cmdstanr:::stanc_cmd(),
    args = c(model$stan_file(), "--auto-format"),
    wd = cmdstan_path(),
    echo = TRUE
)
```
you will get the auto-formatted model in stdout (and as an element of run_vals).

mike-lawrence:
Ok, done, now working on fixing the tests...

mike-lawrence:
The file saving tests are failing because those utilities use the name of the executable file rather than the name of the model file. I think it shouldn't have any side effects to use the name of the model file, so I'm going to change to that.

mike-lawrence:
Oh! @rok-cesnovar the stanc3 autoformatter seems to strip off anything before the `data{` section, including user-defined functions, so changes to anything there won't be detected. Am I misusing it somehow?

mike-lawrence:
Is there a way to run only a subset of the tests? I'm finding it tedious to wait for all tests when debugging just a couple.

rok-cesnovar:
Yes, with https://rdrr.io/cran/testthat/man/test_file.html

In Rstudio you can open a test file and click run tests which I think does the same thing.

mike-lawrence:
Making progress. One thing I'm confused about is this bit in `mod$compile()`:

```r
  if (!force_recompile) {
    message("Model executable is up to date!")
    private$cpp_options_ <- cpp_options
    private$precompile_cpp_options_ <- NULL
    private$precompile_stanc_options_ <- NULL
    private$precompile_include_paths_ <- NULL
    self$exe_file(exe)
    return(invisible(self))
  } else {
    message("Compiling Stan program...")
  }
```

Why do we wipe out the private entries if the model has already been compiled? I'm encountering test fails thanks to this (`include_dirs` gets dropped if one does `mod$compile(force=T); mod$compile()` so we get an error on any subsequent call of  `mod$compile(force=T)`

mike-lawrence:
After a big round of embarrassingly messy commits (sorry, still new to git and proper workflows therein), I think this is done!

Except for one test. It turns out that the pedantic mode flag for stanc is ignored when the auto-formating flag is also set. I can't see any logical reason for that behaivour so I've filed an issue over on stanc. I could remove the `auto-format` flag conditional on `warn-pedantic` being an entry in the `stanc_options` list, but that seems like a hack and the behaviour should really be fixed in stanc instead.

mike-lawrence:
Darn, while all tests (except pedantic) pass on my local machine, I see that the tests are failing here. Is there any way to see more info from failing tests here? Even just the expected and actual values would help.

R/model.R
mike-lawrence: Sorry, missed this comment until now. I coincidentally just posted [a related issue](https://github.com/stan-dev/cmdstanr/issues/337). If we determine there that model_name should always use the exe, I'll change things here to comply.

mike-lawrence:
I ended up adding a workaround to the alleged `stanc --warn-pedantic --auto-format` bug, and with a couple other tweaks it looks like this is good to go! 

The code coverage report is saying that it decreases coverage, but I think that's a false alarm as it's listing affected files that I didn't touch. I think this is due to my hamfisted attempt to rebase to the current master?

I know @rok-cesnovar took a look at an earlier version of this PR, but @jgabry could you take a look with special attention to the section in `compile()` where I use regex strings to discern any existing exes to delete? I'm not very knowledgeable with regex and would hate to either delete user files or leave behind exes that should have been deleted.

rok-cesnovar:
Hey @mike-lawrence,

I took the liberty of merging in the master branch (the master branch on your fork is out of sync with the stan-dev one).
Hope you dont mind. 

> I ended up adding a workaround to the alleged stanc --warn-pedantic --auto-format bug, and with a couple other tweaks it looks like this is good to go!

As responded in the cmdstan issue, I think this is not a bug actually. The formatter is not supposed to work with all other flags.

> The code coverage report is saying that it decreases coverage, but I think that's a false alarm as it's listing affected files that I didn't touch. I think this is due to my hamfisted attempt to rebase to the current master?

This was definitely due to the PR showing too many changes. I think that is cleared now.

> Note also that this PR fixes the check_syntax() bug when models have #includes, so if this PR is merged, my other PR should be denied/deleted.

That PR was just merged.

Thanks for working on this!

mike-lawrence:
> Hope you dont mind.

Nope, that's perfect, I didn't know how to do that! I've resolved the resulting conflict and all tests are passing.

@jgabry, so you don't have to scroll up to figure out what's left:

> Could you take a look with special attention to the section in compile() where I use regex strings to discern any existing exes to delete? I'm not very knowledgeable with regex and would hate to either delete user files or leave behind exes that should have been deleted.

(or @rok-cesnovar if you feel like a single set of eyes on the critical part is ok)

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+      cpp_options_to_compile_flags(cpp_options),
+      stancflags_val
+    )
+  }else{
+    stancflags_val <- trimws(c(stancflags_val, c(stanc_built_options)))
+    these_args = c(stancflags_val)
+  }
+  return(these_args)
+}
+CmdStanModel$set("public", name = "generate_arg_strings", value = generate_arg_strings_method)
```
jgabry, line 438: Does this method need to be public? Is the user ever going to call it themselves?

R/model.R
```diff
@@ -352,14 +352,19 @@ CmdStanModel <- R6::R6Class(
 #'
 NULL
 
-compile_method <- function(quiet = TRUE,
+generate_arg_strings_method <- function(quiet=TRUE,
+                           make_or_stanc=list('make','stanc'),
```
jgabry, line 356:
```suggestion
                           make_or_stanc = c("make", "stanc"),
```

R/model.R
```diff
@@ -352,14 +352,19 @@ CmdStanModel <- R6::R6Class(
...
                            dir = NULL,
                            include_paths = NULL,
                            cpp_options = list(),
                            stanc_options = list(),
-                           force_recompile = FALSE,
                            #deprecated
                            threads = FALSE) {
+
+  if (!is.null(make_or_stanc)) {
```
jgabry, line 364: Is there a use case for it ever being NULL or can we error if it's NULL?

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    stanc_options[["use-opencl"]] <- TRUE
+  }
+  if (is.null(stanc_options[["name"]])) {
+    stanc_options[["name"]] <- model_name
+  }
+  if(make_or_stanc=='make'){
+    char_for_stanc_built_options <- "'"
+  }else{
+    char_for_stanc_built_options <- ""
+  }
```
jgabry, line 409: Can you add a comment here to remind us in the future why it was necessary to use `char_for_stanc_built_options` differently for make and stanc? Maybe Rok will remember but I will definitely forget ;)

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+
+  if(make_or_stanc=='make'){
+    stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
+    these_args = c(
+      cpp_options_to_compile_flags(cpp_options),
+      stancflags_val
+    )
+  }else{
+    stancflags_val <- trimws(c(stancflags_val, c(stanc_built_options)))
+    these_args = c(stancflags_val)
```
jgabry, line 434:
```suggestion
    these_args <- c(stancflags_val)
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    }
+  }
+
+  if (cmdstan_version() < "2.24") {
+    prepare_precompiled(cpp_options, quiet)
+  }
+
+  if(make_or_stanc=='make'){
+    stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
+    these_args = c(
```
jgabry, line 428:
```suggestion
    these_args <- c(
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+  if (is.null(stanc_options[["name"]])) {
+    stanc_options[["name"]] <- model_name
+  }
+  if(make_or_stanc=='make'){
+    char_for_stanc_built_options <- "'"
+  }else{
+    char_for_stanc_built_options <- ""
+  }
+
+  stanc_built_options = c()
```
jgabry, line 411:
```suggestion
  stanc_built_options <- c()
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+  for (i in seq_len(length(stanc_options))) {
+    option_name <- names(stanc_options)[i]
+    if (isTRUE(as.logical(stanc_options[[i]]))) {
+      stanc_built_options = c(stanc_built_options, paste0("--", option_name))
```
jgabry, line 415:
```suggestion
      stanc_built_options <- c(stanc_built_options, paste0("--", option_name))
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    } else {
+      stanc_built_options = c(stanc_built_options, paste0("--", option_name, "=",
```
jgabry, line 417:
```suggestion
      stanc_built_options <- c(stanc_built_options, paste0("--", option_name, "=",
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+        char_for_stanc_built_options, stanc_options[[i]], char_for_stanc_built_options))
+    }
+  }
+
+  if (cmdstan_version() < "2.24") {
+    prepare_precompiled(cpp_options, quiet)
+  }
+
+  if(make_or_stanc=='make'){
```
jgabry, line 426:
```suggestion
  if (make_or_stanc == "make"){
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
+    these_args = c(
+      cpp_options_to_compile_flags(cpp_options),
+      stancflags_val
+    )
+  }else{
```
jgabry, line 432:
```suggestion
  } else {
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    stancflags_val <- paste0(stancflags_val, include_paths_flag, include_paths, " ")
+  }
+
+  if (!is.null(cpp_options$stan_opencl)) {
+    stanc_options[["use-opencl"]] <- TRUE
+  }
+  if (is.null(stanc_options[["name"]])) {
+    stanc_options[["name"]] <- model_name
+  }
+  if(make_or_stanc=='make'){
```
jgabry, line 405:
```suggestion
  if (make_or_stanc == "make") {
```

Just for code consistency. Easier to read if we always use the same conventions if possible.

R/model.R
```diff
@@ -352,14 +352,19 @@ CmdStanModel <- R6::R6Class(
...
                            dir = NULL,
                            include_paths = NULL,
                            cpp_options = list(),
                            stanc_options = list(),
-                           force_recompile = FALSE,
                            #deprecated
                            threads = FALSE) {
+
+  if (!is.null(make_or_stanc)) {
+    checkmate::assert_choice(make_or_stanc, choices=c('make','stanc'))
```
jgabry, line 365:
```suggestion
    checkmate::assert_choice(make_or_stanc, choices = c("make", "stanc"))
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+
+  if (!is.null(cpp_options$stan_opencl)) {
+    stanc_options[["use-opencl"]] <- TRUE
+  }
+  if (is.null(stanc_options[["name"]])) {
+    stanc_options[["name"]] <- model_name
+  }
+  if(make_or_stanc=='make'){
+    char_for_stanc_built_options <- "'"
+  }else{
```
jgabry, line 407:
```suggestion
  } else {
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+run_autoformatter_method <- function(quiet = FALSE,
+                                include_paths = NULL,
+                                stanc_options = list()) {
+
+
+  temp_hpp_file <- tempfile(pattern = "model-", fileext = ".hpp")
+  stanc_options[["o"]] <- temp_hpp_file
+
+  arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='stanc',include_paths=include_paths,
+    stanc_options=stanc_options)
```
jgabry, line 449:
```suggestion
  arg_strings <-
    self$generate_arg_strings(
      quiet = quiet,
      make_or_stanc = "stanc",
      include_paths = include_paths,
      stanc_options = stanc_options
    )
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+
+  #given bug in stanc (https://github.com/stan-dev/cmdstan/issues/944),
```
jgabry, line 451: I guess this turned out not to be a bug

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+  # if pedantic-mode warnings have been requested, first run without --auto-format & with echo=!quiet
+  if('--warn-pedantic' %in% arg_strings){
```
jgabry, line 453:
```suggestion
  if ("--warn-pedantic" %in% arg_strings) {
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+    },
+    error_on_status = FALSE
+  )
+  if (run_log$status != 0) {
+    stop("Syntax error found! See the message above for more information.",
+         call. = FALSE)
+  }
+  return(run_log)
+}
+CmdStanModel$set("public", name = "run_autoformatter", value = run_autoformatter_method)
```
jgabry, line 490: Does this method need to be public or will it only ever be used internally?

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+                           dir = NULL,
+                           include_paths = NULL,
+                           cpp_options = list(),
+                           stanc_options = list(),
+                           force_recompile = FALSE,
+                           #deprecated
+                           threads = FALSE) {
+
+  from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
+  formatted_code = gsub(pattern='\r\n',replacement='\n',x=from_autoformatter$stdout,fixed=T)
```
jgabry, line 502:
```suggestion
  from_autoformatter <- self$run_autoformatter(quiet = quiet, include_paths = include_paths, stanc_options = stanc_options)
  formatted_code <- gsub(pattern = '\r\n',replacement = '\n', x = from_autoformatter$stdout, fixed = TRUE)
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+  current_hash <- digest::digest(formatted_code,algo='xxhash64')
+
+  make_arg_strings = self$generate_arg_strings(quiet=quiet,make_or_stanc='make',include_paths=include_paths,
+    cpp_options=cpp_options,stanc_options=stanc_options,threads=threads)
```
jgabry, line 506:
```suggestion
  make_arg_strings <-
    self$generate_arg_strings(
      quiet = quiet,
      make_or_stanc = "make",
      include_paths = include_paths,
      cpp_options = cpp_options,
      stanc_options = stanc_options,
      threads = threads
    )
```

R/model.R
```diff
@@ -369,22 +374,146 @@ compile_method <- function(quiet = TRUE,
...
+                           include_paths = NULL,
+                           cpp_options = list(),
+                           stanc_options = list(),
+                           force_recompile = FALSE,
+                           #deprecated
+                           threads = FALSE) {
+
+  from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
+  formatted_code = gsub(pattern='\r\n',replacement='\n',x=from_autoformatter$stdout,fixed=T)
+  current_hash <- digest::digest(formatted_code,algo='xxhash64')
```
jgabry, line 503:
```suggestion
  current_hash <- digest::digest(formatted_code, algo = "xxhash64")
```

R/model.R
```diff
@@ -414,17 +557,24 @@ compile_method <- function(quiet = TRUE,
...
-    force_recompile <- TRUE
+    regex_string <- paste0('^(',basename(exe_name_without_hash),'_)[a-f0-9]{16}')
+    regex_string <- cmdstan_ext(regex_string)
+    regex_string <- gsub('(.exe)$','(.exe)',regex_string)
+    regex_string <- paste0(regex_string,'$')
+    existing_exes <- list.files(
+      path = dirname(exe_name_without_hash)
+      , pattern = regex_string
+      , full.names = T
+    )
```
jgabry, line 568:
```suggestion
    existing_exes <- list.files(
      path = dirname(exe_name_without_hash),
      pattern = regex_string,
      full.names = TRUE
    )
```

R/model.R
```diff
@@ -414,17 +557,24 @@ compile_method <- function(quiet = TRUE,
...
 
   if (!force_recompile) {
     message("Model executable is up to date!")
     private$cpp_options_ <- cpp_options
-    private$precompile_cpp_options_ <- NULL
-    private$precompile_stanc_options_ <- NULL
-    private$precompile_include_paths_ <- NULL
+    # private$precompile_cpp_options_ <- NULL
+    # private$precompile_stanc_options_ <- NULL
+    # private$precompile_include_paths_ <- NULL
```
jgabry, line 577: Are these not necessary anymore? If not then go ahead and just delete them instead of commenting them out.

R/model.R
```diff
@@ -437,52 +587,10 @@ compile_method <- function(quiet = TRUE,
...
-  stancflags_val <- paste0("STANCFLAGS += ", stancflags_val, paste0(stanc_built_options, collapse = " "))
-  if (cmdstan_version() < "2.24") {
-    prepare_precompiled(cpp_options, quiet)
-  }
   run_log <- processx::run(
     command = make_cmd(),
-    args = c(tmp_exe,
-             cpp_options_to_compile_flags(cpp_options),
-             stancflags_val),
+    args = c(tmp_exe,make_arg_strings),
```
jgabry, line 593:
```suggestion
    args = c(tmp_exe, make_arg_strings),
```

R/model.R
```diff
@@ -556,61 +664,8 @@ NULL
...
-    stderr_line_callback = function(x,p) {
-      message(x)
-    },
-    error_on_status = FALSE
-  )
-  if (run_log$status != 0) {
-    stop("Syntax error found! See the message above for more information.",
-         call. = FALSE)
-  }
+  from_autoformatter <- self$run_autoformatter(quiet=quiet,include_paths=include_paths,stanc_options=stanc_options)
```
jgabry, line 668:
```suggestion
  from_autoformatter <- self$run_autoformatter(quiet = quiet, include_paths = include_paths, stanc_options = stanc_options)
```

tests/testthat/test-model-compile.R
```diff
@@ -29,7 +29,7 @@ test_that("error if no compile() before model fitting", {
 test_that("compile() method works", {
   skip_on_cran()
   # remove executable if exists
-  exe <- cmdstan_ext(strip_ext(mod$stan_file()))
+  exe = cmdstan_ext(paste0(strip_ext(mod$stan_file()),'_d5f293b0266e270c')) #entered by-hand so test isn't circular
```
jgabry, line 32:
```suggestion
  exe <- cmdstan_ext(paste0(strip_ext(mod$stan_file()),'_d5f293b0266e270c')) #entered by-hand so test isn't circular
```

tests/testthat/test-model-compile.R
```diff
@@ -71,7 +115,7 @@ test_that("compile() method works with spaces in path", {
   file.copy(stan_file, stan_model_with_spaces)
 
   mod_spaces <- cmdstan_model(stan_file = stan_model_with_spaces, compile = FALSE)
-  exe <- cmdstan_ext(strip_ext(mod_spaces$stan_file()))
+  exe = cmdstan_ext(paste0(strip_ext(stan_model_with_spaces),'_d5f293b0266e270c')) #entered by-hand so test isn't circular
```
jgabry, line 118:
```suggestion
  exe <- cmdstan_ext(paste0(strip_ext(stan_model_with_spaces),'_d5f293b0266e270c')) #entered by-hand so test isn't circular
```

tests/testthat/test-model-compile.R
```diff
@@ -94,7 +138,7 @@ test_that("compilation works with include_paths", {
   skip_on_cran()
 
   stan_program_w_include <- testing_stan_file("bernoulli_include")
-  exe <- cmdstan_ext(strip_ext(stan_program_w_include))
+  exe = cmdstan_ext(paste0(strip_ext(stan_program_w_include),'_69554fb67cd9e770')) #entered by-hand so test isn't circular
```
jgabry, line 141:
```suggestion
  exe <- cmdstan_ext(paste0(strip_ext(stan_program_w_include),'_69554fb67cd9e770')) #entered by-hand so test isn't circular
```

jgabry (commented):
Ok so I still haven't worked my way up to verifying the regex stuff (sorry for the delay!), but I just went through and make a ton of really tiny suggestions just to bring the code style in line with what we've been using. None of your code seemed wrong or worse, it's just so much easier to read to the source code of a package if the same style is used throughout the codebase. And instead of me just making all the changes I did it by using the "suggestion" feature when reviewing a PR so you can see the changes and accept them. That way you'll also get more familiar with the code style we've been using for future PRs (and I hope there are future PRs, really glad you've been contributing!).

R/model.R
mike-lawrence: Oh, nope, could totally be private. I just was copying existing code and didn't catch that bit.

R/model.R
mike-lawrence: Throwing an error when NULL seems appropriate; I was copying the template for how other functions used checkmate, but especially if we make this a private function, it should never be NULL.

R/model.R
mike-lawrence: oops, my bad; I know `<-` is the style here, I'm just *very* used to `=` instead and missed this one (and possibly others, just thought I'd comment on this first one). I'll do a search for `=` before the next PR : )

R/model.R
mike-lawrence: oh! If it's within cmdstanr-style to use multi-line arguments, that's awesome as that's my preference (though I prefer commas to start lines so missing ones are easier to see/avoid, but I know that's uncommon/pattern-breaking for most).

R/model.R
mike-lawrence: Right! I'll replace with a more descriptive comment.

R/model.R
mike-lawrence: I posted a comment asking what they were even there for in the first place. It seems odd to delete them when the model is compiled. I commented them out as a sort of marker that their removal was yet to be fully decided on. I can't think of any breaking changes that their removal would cause, so I'll commit their proper removal.

R/model.R
mike-lawrence: Ditto with the other one; totally should be private.

R/model.R
jgabry: Yeah in some cases NULL is OK so we let it through but if it's not OK here then I think in that case you can just remove the if clause. I'm pretty sure the checkmate assert function will error by default if NULL.

R/model.R
jgabry: Ok cool. I think we should make it private if it won't be called by the user. Oh and if you switch that to private you'll need to use `private$generate_arg_strings` instead of `self$generate_arg_strings` when calling it. I had that issue once in the past and it took me an embarrassingly long time to figure out why `self$foo` wasn't working anymore.

R/model.R
mike-lawrence: So it should really be "\"" for the stanc case, I'll commit that change momentarily. That needs to be a double-quote because `processx::run()` uses single-quotes to encapsulate entries in the list supplied to it's `args` argument. 

The make case is because the entire `STANCFLAGS +=` string gets encapsulated by double quotes at some point, so it's necessary to use a single quote to encapsulate anything inside that string that needs encapsulating. If we can find where the double-quotes are added around the `STANCFLAGS +=` string and change it to single quotes, we can get rid of this conditional.

mike-lawrence: Oh! I figured it out. processx::run() is smart enough to just use whatever doesn't break any existing encapsulation, so it's the one adding the double quotes to the stancflags string. So I'll get rid of this conditional.

jgabry: > So it should really be """ for the stanc case

In that case I think you'll have to use `'"'` with single quotes wrapping the double quote or R won't like it. 


> The make case is because the entire STANCFLAGS += string gets encapsulated by double quotes at some point, so it's necessary to use a single quote to encapsulate anything inside that string that needs encapsulating. If we can find where the double-quotes are added around the STANCFLAGS += string and change it to single quotes, we can get rid of this conditional.

Ok I think I see what you mean. I'm not sure where the double quotes are added at the moment, but we can probably track it down.

mike-lawrence:
Ah shoot, stanc doesn't like any kind of quotes around the value supplied to it's `--o` argument if-and-only-if there are quotes of any kind around the entire argument-value pair. So this works:
```
bin/stanc bernoulli.stan  --o="/tmp/RtmpcR8CG3/model-5a2cb26298ea1.hpp"
```
(works with single-quotes around the value too)

But this doesn't:
```
bin/stanc /home/mike/Code/cmdstanr/tests/testthat/resources/stan/bernoulli.stan  '--o="/tmp/RtmpcR8CG3/model-5a2cb26298ea1.hpp"'
```
Nor does this:
```
bin/stanc /home/mike/Code/cmdstanr/tests/testthat/resources/stan/bernoulli.stan  "--o='/tmp/RtmpcR8CG3/model-5a2cb26298ea1.hpp'"
```
Now, I'm not super familiar with the kinds of things that are likely to be passed as values to the various available arguments to make and stanc; presumably they should all be quote-encapsulated? I wonder if there's a way to tell `processx::run()` to not add it's own quotes? Else I'm having difficulty thinking of a way around this.

jgabry:
> Actually, do the failures above seem like bugs in stanc maybe?

Interesting. To be honest I have no clue if that’s the intended behavior or not. Maybe @rok-cesnovar will know.

mitzimorris:
> Actually, do the failures above seem like bugs in stanc maybe? I gather that most unix commands should be compatible with that pattern.

you need to put the flags before the filename.

```
bin/stanc '--o="foo.hpp"' bernoulli_ppc.stan 
```
produces output file which has quotes in the name:  file name isn't `foo.hpp`, it's `"foo.hpp"`

mike-lawrence:
>you need to put the flags before the filename.
Hm, I don't actually think that's true, if by flags you mean the `--o` bit. For example this works fine:

bin/stanc bernoulli.stan  --o="temp.hpp"

> produces output file which has quotes in the name: file name isn't foo.hpp, it's "foo.hpp"

Ah, this is the reason stanc is failing with the doubly-quote-encapsulated `--o` argument! Because the quotes are considered part of the *path*, which then doesn't exist. 

Is this perhaps still arguably unexpected behaviour on stanc's part? That is, I'd have thought most unix programs that have output-to-file arguments would treat quote-encapsulated string values as pointing to a location with possible spaces in the name, not pointing to locations with literal quotes in the name.

mitzimorris:
> Is this perhaps still arguably unexpected behaviour on stanc's part? That is, I'd have thought most unix programs that have output-to-file arguments would treat quote-encapsulated string values as pointing to a location with possible spaces in the name, not pointing to locations with literal quotes in the name.

not exactly - single and double quotes are used to assemble the list of argument strings available to the program's main.  what the program does with them is up to it.  in unix, in order to have a space in a path name, the space must be escaped via backslash - "ab c.hpp" is "ab\ c.hpp"

mike-lawrence:
Ah, ok, that makes sense. And I think that gives me an idea how to make things work with both stanc and processx as they are. When we're sanitizing the value strings in cmdstanr, I'll just look for spaces and replace with `\ `, which then obviates the need to add any quotes ourselves, leaving just the quotes around whole argument-value pairs that processx adds, which I think both make and stanc are fine with.

mike-lawrence:
Any idea why it's failing on just `windows-latest (devel)`? Looks like the failures are mostly in the test-fit-shared set, which I don't think this PR is likely to affect, but there is one warning (`cannot remove file 'resources/stan/folder spaces', reason 'Permission denied'`) in the test-model-compile set that I don't get elsewhere. Could this be an issue with the CI? Is it possible to selectively trigger a re-run of `windows-latest (devel)`?

rok-cesnovar:
> Could this be an issue with the CI? Is it possible to selectively trigger a re-run of windows-latest (devel)?

That is probably a one-off CI thing. Just rerun all jobs, its free compute cycles anyway.

jgabry:
Thanks @mitzimorris for helping sort out the quotes issue! 

> I'll just look for spaces and replace with `\ `, which then obviates the need to add any quotes ourselves, leaving just the quotes around whole argument-value pairs that processx adds, which I think both make and stanc are fine with.

@mike-lawrence Yeah that sounds good.

R/model.R
```diff
@@ -369,22 +374,152 @@ compile_method <- function(quiet = TRUE,
...
+  arg_strings <-
+    private$generate_arg_strings(
+      quiet = quiet,
+      make_or_stanc = "stanc",
+      include_paths = include_paths,
+      stanc_options = stanc_options
+    )
+
+  # if pedantic-mode warnings have been requested, first run without --auto-format & with echo=!quiet
+  if ("--warn-pedantic" %in% arg_strings) {
```
rok-cesnovar, line 452: I think it might be better to make a utility function (not part of the CmdStanModel) to run stanc3 with arbitrary arguments on a supplied .stan file. 

You can use that in check_syntax and autoformatter, etc.

There are other flags that won't work with the auto-format, like the standalone functions etc.

rok-cesnovar (commented):
Thanks for working on this!

I would redesign the functions to have a general stanc3 utility function that can be used everywhere. If you want I can help with that.

This also needs a merge with recent master. Sorry for the late review.

mike-lawrence:
> I would redesign the functions to have a general stanc3 utility function that can be used everywhere. If you want I can help with that.

Can do. 

> This also needs a merge with recent master. Sorry for the late review.

What's the proper sequence of gets commands to achieve this? I tried it a couple weeks ago and ended up completely mangling things￼

jgabry:
I was also talking with @mitzimorris and she had some opinions that we should take into consideration about how cmdstanr is doing compilation in general and also in relation to this hashing approach. @mitzimorris When you have a chance can you  summarize those thoughts here? Thanks!

mitzimorris:
the problems with this approach are:

- hash codes are not human readable; putting these into file names results in directories that are difficult to navigate

- the use case here isn't someone using CmdStanR directly; for CmdStanR users changes to the Stan program should trigger recompile; the timestamp would be all you need to detect this.

- this adds a lot of complexity to the code base.    there's a risk that this feature, if it gives the wrong answer, will compile either too often or not often enough. 

perhaps some parts of this PR should got into the code base, but not use of hashcodes in the filename.

mike-lawrence:
> hash codes are not human readable; putting these into file names results in directories that are difficult to navigate

The hashes are at the end of the filenames, so I don't see how they make anything more difficult.

> the use case here isn't someone using CmdStanR directly; for CmdStanR users changes to the Stan program should trigger recompile; the timestamp would be all you need to detect this.

I disagree. I've been using this myself and love that if I change a comment or whitespace, it doesn't trigger a recompile.

> This adds a lot of complexity to the code base. there's a risk that this feature, if it gives the wrong answer, will compile either too often or not often enough.

I disagree, I actually ended up simplifying some things, and the hashing part you are concerned about has tests that you're welcome to inspect and improve.


This feature was discussed previously and no issues were raised, so I put a bunch of work into first figuring out how to contribute then implementing things. Feels rather shitty to only now encounter pushback.

jgabry:
> This feature was discussed previously and no issues were raised, so I put a bunch of work into first figuring out how to contribute then implementing things. Feels rather shitty to only now encounter pushback.

I feel responsible for that given that I definitely encouraged you to work on this. I should have made sure to solicit this feedback sooner. I'm really sorry about that! 

As for the particular points: 

* The hashes in file names: I don't have a super strong opinion about this. I basically never look at the filenames when I'm using CmdStanR so it doesn't bother me personally if they have hashes in them. But @mitzimorris do you have an example where you think this would interfere with navigation or cause some other problem? You know about file systems and things like that better than I do. 

* Use case: this seems useful for packages like brms, but not necessary since brms could check if the formula and priors were changed and avoid recompiling. But it also seems useful for the case @mike-lawrence mentioned because we shouldn't have to recompile a Stan program if the comments or whitespace changes, only if the code changes. @mitzimorris Is there another way besides hashing that we can avoid having to recompile if only comments change? 

* Complexity: this definitely makes the code base more complicated (even just the reliance on regular expressions makes it more complicated since they make my head hurt ;) Good testing can mitigate that concern, but I do think this makes the compile() method more complicated. I may have not looked at the latest changes yet though, so I might be commenting with slightly outdated knowledge. I will double check.

rok-cesnovar:
> I basically never look at the filenames when I'm using CmdStanR so it doesn't bother me personally if they have hashes in them.

I agree with this point. I am not sure a use case of someone compiling in cmdstanr and then going in the folder and using the exe is a real one? At least not a common one.

> Complexity

I have to do a proper review, but I doubt reg exes are required here. I was juggling too many stuff this week and forgot to respond properly here.

I feel that compiling based on hashes is a valuable thing. I dont feel hashing should be considered dangerous or prone to errors. Its very common. If we cant agree in making it the default option we can have it as an option? My hunch is users will like it...

mike-lawrence:
I should also say that my initial reaction to Mitzi's feedback was a little overwrought; I hadn't considered that she hadn't been tagged in to give feedback until only recently, in which case my complaint over timing were not particularly valid. Let me see if I can rework things to create that public stanc method Rok suggests in a way that will simplify things even further. 

Also, certainly the regex bit is probably unnecesssarily complicated in retrospect; I was trying to use a standard tool that I thought everyone but me would be very familiar with, but now that I see that I'm not alone in my regex unease, there's probably a more legible way to combine the the known stanfile name, hash length and executable suffix to find any outdated exes. So I'll update that bit too.

mitzimorris:
> But @mitzimorris do you have an example where you think this would interfere with navigation or cause some other problem? You know about file systems and things like that better than I do.

hash codes in file names isn't going to cause system problems.  I do most of my work at the command line or in emacs or the MacOS file browser.  long names with hash codes in them are annoying clutter, but that's just me.

the only thing that would be problematic is writing scripts - it won't be possible to hardcode names - and maybe
one shouldn't hardcode names either.

jgabry:
> I see the problem with slow compilation times; I don't much like this solution - that's just my opinion.

Is there an alternative solution you can think of? 

> the main thing is that the code is correct, easy to understand and maintain, and the feature is properly documented.

Totally agree!

mitzimorris:
> Is there an alternative solution you can think of?

aside from C++ compilers getting faster, and BRMS doing better on model caching,
it would be much nicer to track file contents via hidden directories and keep the hashcodes there, à la git.
keep hashcodes out of filenames.  this might be more work, but if you could do it, that would be cool.

mike-lawrence:
Happy to do the hidden folder approach, I just personally found the use of the filename itself more elegant, but I can appreciate that others have different aesthetics. And the difficulty in hard-coding the executable names in scripts is certainly a non-aesthetic argument in favor of folders; indeed, I had some trouble in generating the tests that would have been avoided by folders. Does the idea of a hidden folder in the directory of the stanfile/executable affect the discussion elsewhere on compile directories? Would it make any sense to use the hidden folder for compile stuff too? (where now it uses a tmp folder I think?)

mitzimorris:
> Would it make any sense to use the hidden folder for compile stuff too? (where now it uses a tmp folder I think?)

no, because that's solving a different problem:  GNU make can't handle paths with spaces in the name.

jgabry:
> note that we're moving, step by step, from a lightweight wrapper interface to IDE levels of file mgmt.

This is definitely true, but for better or worse I've been learning over the past year that while the lightweight wrapper idea is really appealing to us developers, users don't want a lightweight wrapper interface (not just based on this example but based on feedback from many different users). They're using CmdStanX not because it is (or was) a lightweight wrapper but because it has the latest version of Stan and is easier (though not always easy!) to install and maybe a few other small reasons. 

On a separate note, @rok-cesnovar do you have any thoughts on this hidden folder approach? We should make sure to discuss this thoroughly before anyone invests too much time in it ;)

rok-cesnovar:
> On a separate note, @rok-cesnovar do you have any thoughts on this hidden folder approach?

I like it!

mike-lawrence:
Shoot. Since the last I looked at this I somehow convinced myself that we'd all decided that it would cause too much bloat, but on re-familiarizing myself with the most recent discussion I see that this wasn't the case. Now, a bunch of work has been done in the interim so I'll have to figure out how to merge those, but before I do: what about writing the hash of the auto-formatted code as a comment at the end of the stan file itself? I think most code editors these days take note when an open file has been changed and either auto-update or give the user a notice. Maybe said notices would be annoying?

martinmodrak:
I did read that only after putting some work into #483 which tries to address a similar issue in a simpler manner. I appreciate there is a lot of work in this! I do have one quick technical feedback:
I think not recompiling the model on comments/whitespace changes is potentially confusing as it can result in error messages signalling a mismatched line number. So if that's a feature I think it should not be turned on by default to avoid tripping novice users.

jgabry:
> I think not recompiling the model on comments/whitespace changes is potentially confusing as it can result in error messages signalling a mismatched line number. So if that's a feature I think it should not be turned on by default to avoid tripping novice users.

This is a really good point that I hadn't considered. I agree that if we do this it shouldn't be turned on by default for this reason. 



> Shoot. Since the last I looked at this I somehow convinced myself that we'd all decided that it would cause too much bloat, but on re-familiarizing myself with the most recent discussion I see that this wasn't the case.

I thought we also decided that, so maybe I just missed the more recent discussion.

mike-lawrence:
> I think not recompiling the model on comments/whitespace changes is potentially confusing as it can result in error messages signalling a mismatched line number.

FYI, the way I do this in [aria](https://github.com/mike-lawrence/aria) is that there's a syntax check before the auto-formatting, so any messages involving line numbers come out just fine.

martinmodrak:
> a syntax check before the auto-formatting

That helps with some use cases, but there are also runtime errors/warnings with line numbers. I don't want to push this too hard - I agree that actual problems with changing line numbers will likely be somewhat rare. But I am a bit conservative about this and value "no surprises" as a very important value for an interface, while others may have different value judgements on this.

mike-lawrence:
> That helps with some use cases, but there are also runtime errors/warnings with line numbers.

Point taken.

rok-cesnovar:
Closing this PR for now. It has some conflicts that need resolving and has become stale. There are also alternatives for users that need this feature.

PR #312, opened 2020-10-08
Replace output_dir with dir and deprecate output_dir

### LINKED ISSUE #297: Deprecate `output_dir` in favor of `dir`
Per the discussion over at https://github.com/stan-dev/cmdstanpy/issues/289, for consistency we should change the `output_dir` argument of the model fitting methods to `dir` to match the argument name used in `install_cmdstan()` and the `compile()` method.

rok-cesnovar:
#### Summary

Fixes #297 

#### Copyright and Licensing


rok-cesnovar:
I tried to also rebuild the .html files but `pkgdown::build_reference()` made a bunch of unrelated changes so I am assuming I should have called it differently.

jgabry:
> I tried to also rebuild the .html files but `pkgdown::build_reference()` made a bunch of unrelated changes so I am assuming I should have called it differently.

That might just be because we've merged other doc changes without regenerating the html, so they're getting included now.

R/args.R
```diff
@@ -34,9 +34,16 @@ CmdStanArgs <- R6::R6Class(
...
-
+                          dir = NULL,
+                          validate_csv = TRUE,
+                          # deprecated
+                          output_dir = NULL) {
+      # temporary deprecation warnings
+      if (!is.null(output_dir)) {
+        warning("'output_dir' is deprecated. Please use 'dir' instead.")
+        dir <- output_dir
+      }
```
jgabry, line 45: Do we need this in args.R if we also have it in each of the mode fitting methods? Is there a  case where we only throw this warning here without also throwing it from, e.g., `$sample()`?

jgabry:
@rok-cesnovar This looks good, just one question above. 

@mitzimorris Just wanted to double check that you're also planning on doing this in CmdStanPy. I think we agreed to just use `dir` everywhere, right?

R/args.R
rok-cesnovar: Oh right. Let me double check.

rok-cesnovar: Actually yeah, I just removed it. cmdstanArgs is internal. No need to keep it.

rok-cesnovar:
> That might just be because we've merged other doc changes without regenerating the html, so they're getting included now.

see commit https://github.com/stan-dev/cmdstanr/pull/312/commits/7bb23f72bd7e4cdf7b485c5fed548fef0619de81 for the changes after pkgdown::build_reference()

Thanks!

jgabry:
> see commit [7bb23f7](https://github.com/stan-dev/cmdstanr/commit/7bb23f72bd7e4cdf7b485c5fed548fef0619de81) for the changes after pkgdown::build_reference()

Yeah looks like stuff that hadn't been updated previously. Should be fine.

jgabry:
@mitzimorris Before I merge this, just want to confirm that this is the plan for CmdStanPy too. We decided to use `dir` everywhere right?

mitzimorris:
overall, I don't see a huge benefit from changing this now - perhaps we should have gone with the simplest possible argument name - we didn't.

reasons not to do this:

- the name `output_dir` comes from current CmdStan argument name `output file=...` and when we flatten CmdStan's arguments, it will be `output_file` and `data_file`

- keeping `output_dir` allows for the possibility of `data_dir` later.

jgabry:
Thanks for the comments: 

> * keeping `output_dir` allows for the possibility of `data_dir` later.

Yeah that's a good point that I hadn't considered before we talked about this. It does seem fine to have `output_dir` instead of `dir` if we're going to potentially have multiple `dir` related arguments.

rok-cesnovar:
Thanks! Closing this and the issue then.

PR #319, opened 2020-10-17
Expose chain_id

### LINKED ISSUE #266: Expose `chain_id` argument?
CmdStanPy exposes CmdStan's `chain_id` argument but we currently aren't exposing it in CmdStanR (we just always use 1:num_chains). 

@rok-cesnovar Do you think we need to expose `chain_id` for if people want to run the same call to `$sample()` on different machines and then combine the results? (If so, we'd also want a way to combine multiple CmdStanMCMC objects into a single one).

If we do implement this we should make sure to consider the issue in  https://github.com/stan-dev/cmdstanpy/pull/266.

rok-cesnovar:
#### Summary

Fixes #266 

#### Copyright and Licensing


NEWS.md
```diff
@@ -20,6 +20,9 @@ in fitting methods. (#281, #294)
 * `install_cmdstan()` gains argument `version` for specifying which version of 
 CmdStan to install. (#300, #308)
 
+* `$sample()` method for CmdStanModel objects gains argument `version` for specifying
```
jgabry, line 23:
```suggestion
* `$sample()` method for CmdStanModel objects gains argument `chain_id` for specifying
```

R/model.R
```diff
@@ -761,6 +761,9 @@ CmdStanModel$set("public", name = "check_syntax", value = check_syntax_method)
 #'   quantities block. If the parameters block is empty then using
 #'   `fixed_param=TRUE` is mandatory. When `fixed_param=TRUE` the `chains` and
 #'   `parallel_chains` arguments will be set to `1`.
+#'   * `chain_id`: (vector) A vector of chain IDs. Must contain `chains` unique
+#'   positive integers. If not set, the default chain IDs are used (integers
+#'   starting from `1`).
```
jgabry, line 766: Should we document this higher up next to the other arguments related to the chains? On the other hand, this argument won't be used much so I can also see a reason to keep it documented here at the bottom.

R/model.R
```diff
@@ -795,6 +798,7 @@ sample_method <- function(data = NULL,
                           fixed_param = FALSE,
                           validate_csv = TRUE,
                           show_messages = TRUE,
+                          chain_id = seq_len(chains),
```
jgabry, line 801: Same question as above about whether this should go at the bottom (because won't be used much) or higher up so it's next to the chain related arguments. I'm not sure, so either way is ok I guess.

R/model.R
```diff
@@ -848,6 +852,7 @@ sample_method <- function(data = NULL,
   checkmate::assert_integerish(chains, lower = 1, len = 1)
   checkmate::assert_integerish(parallel_chains, lower = 1, null.ok = TRUE)
   checkmate::assert_integerish(threads_per_chain, lower = 1, len = 1, null.ok = TRUE)
+  checkmate::assert_integerish(chain_id, lower = 1, len = chains, unique = TRUE, null.ok = FALSE)
```
jgabry, line 855: Can we just fall back on the check on `proc_ids` when validating the CmdStanArgs object? https://github.com/stan-dev/cmdstanr/blob/bc3c8a7824fcd13184fe447e73dfbbcc295eec83/R/args.R#L453

Or do we need to check it immediately here to avoid an issue before it's checked later?

jgabry (commented):
Looks good! I made a few tiny comments. The only other question I have is whether this should be called `chain_id` or `chain_ids`. Maybe the plural `chain_ids` makes the most sense?

R/model.R
rok-cesnovar: Moved up. I think that makes sense.

R/model.R
rok-cesnovar: Moved up. I think that makes sense.

R/model.R
rok-cesnovar: This does some additional checks. And I think the error mentioning `'chain_ids'` as opposed to `'sefl$procs_ids'` is more informative to the user.

jgabry: Ok yeah that makes sense.

jgabry (approved):
This looks good to go! Not merging in case there was anything else you wanted to do, but if not then go ahead and merge.

PR #336, opened 2020-11-03
Add procs$return_codes()

### LINKED ISSUE #332: Knowing if optimization (or variational inference) is successful or not by a value
When I use the optimization, I can see the logs in the screen, which tells me if the optimization faces some problem and fails, such as infinity log probability. 

Is that possible  I can get the status from a value returned by some function for optimization (or variational inference) ?  I face this issue when I want to use optimization for thousands of times.

rok-cesnovar:
#### Summary

Fixes #332 by adding adding `fit$runset$procs$return_codes()` to replace calls like `fit$runset$procs$get_proc(1)$get_exit_status()` (or X of them when using X chains).

#### Copyright and Licensing


jgabry:
Good idea. Should we also allow `fit$return_code()`? If this is just for internal use I’d say no, but if users are going to call
it then maybe we should add this so they don’t have to work with the runset object (which is undocumented).

rok-cesnovar:
Oh yeah, good call on runset being undocumented. Will add that tomorrow.

jgabry:
This looks good. I just pushed a commit adding documentation. @rok-cesnovar Does the doc look ok to you? If so then I think this can be merged.

jgabry (approved):

rok-cesnovar:
> I only mentioned in the doc that zero means success.

Yeah for other codes it's different for each method and I think we don't need to go into details there.

> If the doc looks good 

It does. Thanks!

Issue #351, opened 2020-11-17
Clarify pre-reqs for install


bob-carpenter:
@mitzimorris and I just tried to walk users through `cmdstanr` install from the instructions page and there are several key steps that are missing from the instructions page:

https://mc-stan.org/cmdstanr/#installation

* [ ] needs to point to `cmdstan` installation
* [ ] needs to say that the "beta" is 0.2.0 or whatever the recent number is
* [ ] needs to have `Rtools` installed before `devtools`
* [ ] for `Rtools`, we need to modify your path `PATH="${RTOOLS40_HOME}\usr\bin;${PATH}"`
* [ ] `https://cran.r-project.org/bin/windows/Rtools/`

* [ ] `devtools`: need to mention that needs to be installed
    - and if there are any pre-reqs to installing devtools, those also need to be mentioned
* [ ] we also need `dependencies = TRUE` to get `posterior`, etc., on which `cmdstanr` depends.  
`install.packages("cmdstanr", repos = c("https://mc-stan.org/r-packages/", getOption("repos")), dependencies = TRUE)` 
or we need to check the same in RStudio's installer
* [ ] need instructions to install `cmdstan`
`library(cmdstanr);`
`install_cmdstan(cores = 2)`
And we probably need to tell people that `cmdstan` is being installed in a hidden location they can find with `cmdstan_path()`

jgabry:
Thanks Bob, that's really helpful. A few comments: 

Yeah in the readme we just have installation instructions for the R package, not for CmdStan itself. Those instructions are in the Getting Started vignette. To put that in the readme we'll need to duplicate the content because it's still essential to have it in the vignette too (if the package makes it to CRAN then unfortunately the github readme won't be included but the vignettes will, so we need the content in the vignettes). So I think what makes the most sense is to make it more clear in the readme that the CmdStan instructions are in the vignette instead of duplicating the content. You're definitely right that we don't currently explicitly say this (and we should!). 

As for requirements, yeah you're right it looks like RTools isn't mentioned, so we should definitely add that. That's an oversight. As for devtools, I didn't think that's required. As far as I know RTools can be installed without devtools. 

> * [ ]  we also need `dependencies = TRUE` to get `posterior`, etc., on which `cmdstanr` depends.
>   `install.packages("cmdstanr", repos = c("https://mc-stan.org/r-packages/", getOption("repos")), dependencies = TRUE)`
>   or we need to check the same in RStudio's installer

[...]

PR #353, opened 2020-11-17
Update installation instructions

jgabry:
@rok-cesnovar @mitzimorris @bob-carpenter Based on feedback in #351 I've made the following edits: 

* Clearly mention in README that CmdStan has to be installed, not just the R package. A link is provided to the Getting Started vignette to avoid duplicating the installation content already in the vignette. 

* Add a link to the toolchain requirements in the installation section of the Getting Started vignette. These requirements change from time to time, so I think it's better to link to the CmdStan guide (https://mc-stan.org/docs/cmdstan-guide/cmdstan-installation.html). Otherwise we'll have to change the information in multiple places every time it gets updated. 

I think I addressed the other points in #351 in my response over there. But if I'm wrong in some of my comments over there we can add more to this PR to address the additional points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ analysis.
 
 ### Installation
 
+#### Installing the R package
+
 You can install the beta release of the **cmdstanr** R package with 
 
 ```r
@@ -44,6 +46,14 @@ or you can install the latest development version from GitHub:
 # install.packages("remotes")
 remotes::install_github("stan-dev/cmdstanr")
 ```
+
+#### Installing CmdStan
+
+If you don't already have CmdStan installed then, in addition to installing the
+R package, it is also necessary to install CmdStan. Instructions are provided in
+the [_Getting started with
+CmdStanR_](https://mc-stan.org/cmdstanr/articles/cmdstanr.html) vignette.
+
 
 ### Contributing 
 
--- a/README.md
+++ b/README.md
@@ -50,9 +50,10 @@ remotes::install_github("stan-dev/cmdstanr")
 #### Installing CmdStan
 
 If you don't already have CmdStan installed then, in addition to installing the
-R package, it is also necessary to install CmdStan. Instructions are provided in
-the [_Getting started with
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -45,6 +47,15 @@ or you can install the latest development version from GitHub:
 remotes::install_github("stan-dev/cmdstanr")
 ```
 
+#### Installing CmdStan
+
+If you don't already have CmdStan installed then, in addition to installing the
+R package, it is also necessary to install CmdStan using CmdStanR's 
+`install_cmdstan()` function. A suitable C++ toolchain is also required. 
+Instructions are provided in the [_Getting started with
+CmdStanR_](https://mc-stan.org/cmdstanr/articles/cmdstanr.html) vignette.
```
rok-cesnovar, line 56: Maybe mention that on Windows they can double check the toolchain is installed properly with check_cmdstan_toolchain()? I know this is a part of install_cmdstan so maybe redundant,

jgabry:
Note to self: also take Bob's suggestion and do this: 


> Rather than duplicating instructions, can the readme just point to the getting started doc with full installation instructions? Or could the install instructions be pulled out of the getting started doc into their own page. I agree that we don't want to maintain duplicate instructions---it never ends well.

jgabry:
> Note to self: also take Bob's suggestion and do this:
> 
> > Rather than duplicating instructions, can the readme just point to the getting started doc with full installation instructions? Or could the install instructions be pulled out of the getting started doc into their own page. I agree that we don't want to maintain duplicate instructions---it never ends well.

Actually we already have that in the getting started doc!

jgabry:
Ok I think this is all set. @rok-cesnovar I added a mention of `check_cmdstan_toolchain()` like you suggested, so I'm going to go ahead and merge this.

jgabry:
Actually I'll regenerate the online doc and then merge

rok-cesnovar:
Thanks!

PR #361, opened 2020-11-20
Argument to turn on pedantic mode 

### LINKED ISSUE #358: Making it easier to use pedantic mode
Should we add an argument to `$check_syntax()` (or even to `$compile()`) for turning on pedantic mode? I know we already can specify it in `stanc_options` but I think most users will stay away from the `stanc_options` argument. But we could add an argument like this 

```r
mod$check_syntax(pedantic=TRUE) # should probably be FALSE by default
```

which would make it much more likely for users to try it out and hopefully that would both benefit users and help get @rybern and others more feedback on the feature. 

Thoughts? 

This is also related to #298 where @mitzimorris rightfully pointed out that the `include_paths` argument is redundant given that it is part of `stanc_options`. I think users will be more likely to take advantage of certain features if we give them ON/OFF switches rather than making them specify compiler flags. But I also understand @mitzimorris's concern about redundancy and not offering too many ways to do the same thing if not necessary.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #358 

Adds `pedantic` argument to `$compile()` and `$check_syntax()`. 

#### Copyright and Licensing

**Columbia University**


R/model.R
```diff
@@ -520,23 +520,29 @@ CmdStanModel$set("public", name = "compile", value = compile_method)
 #' @section Usage:
 #'   ```
 #'   $check_syntax(
-#'     quiet = FALSE,
+#'     pedantic = FALSE,
 #'     include_paths = NULL,
-#'     stanc_options = list()
+#'     stanc_options = list(),
+#'     quiet = FALSE
```
jgabry, line 526: I rearranged these arguments a bit. I think for check_syntax `quiet` is probably only used by packages like brms so it doesn't need to be the first argument.

tests/testthat/test-model-compile.R
```diff
@@ -232,6 +233,55 @@ test_that("dir arg works for cmdstan_model and $compile()", {
   )
 })
 
+test_that("compiling stops on hyphens in stanc_options", {
```
jgabry, line 236: This isn't new I just moved it up in the file to keep the compilation tests separate from check_syntax tests

tests/testthat/test-model-compile.R
```diff
@@ -232,6 +233,55 @@ test_that("dir arg works for cmdstan_model and $compile()", {
...
+    fixed = TRUE
+  )
+  expect_error(
+    mod$compile(stanc_options = hyphens3),
+    "No leading hyphens allowed in stanc options (--o). Use options without leading hyphens, like for example `stanc_options = list('allow-undefined')`",
+    fixed = TRUE
+  )
+})
+
+test_that("compiling works with only names in list", {
```
jgabry, line 275: This isn't new I just moved it up in the file to keep the compilation tests separate from check_syntax tests

tests/testthat/test-model-compile.R
```diff
@@ -280,51 +331,9 @@ test_that("check_syntax() works", {
     mod_ok$check_syntax(stanc_options = list("allow-undefined", "warn-pedantic"), quiet = TRUE),
     regexp = NA
   )
-
-  pedantic_model <- write_stan_file("
```
jgabry, line 284: I removed the pedantic tests here since we have a separate test for pedantic mode below now

jgabry:
For now I just added a `pedantic` argument to `check_syntax()` but I actually think it could be worth adding it to `compile()` also. What do you think?

R/model.R
rok-cesnovar: Good call.

rok-cesnovar:
> For now I just added a pedantic argument to check_syntax() but I actually think it could be worth adding it to compile() also. What do you think?

I agree with adding it to compile()

rok-cesnovar (approved):
Looks good. I would add this to compile as well.

jgabry:
Ok thanks, I'll add it to compile today if I have time. If not then one of these next few days.

jgabry:
Ok I just added the argument also for the compile method, however it doesn't work if the executable is already up to date: 

```r
mod <- cmdstan_model(..., pedantic=TRUE) # will show pedantic warnings
mod <- cmdstan_model(..., pedantic=TRUE) # calling it again will just say up to date and not show warnings
```

@rok-cesnovar Do you think that's OK or should we maybe force recompile if pedantic is TRUE?

rok-cesnovar:
> Do you think that's OK or should we maybe force recompile if pedantic is TRUE?

I think that is fine. I don't like forcing recompile, The only other solution would be to call `check_syntax `with pedantic in that case. check syntax is instant.

jgabry:
> The only other solution would be to call `check_syntax `with pedantic in that case. check syntax is instant.

Yeah that's a possibility. We could do it automatically but maybe we can just recommend to the user to call check_syntax() if they want to use pedantic mode for a model that's already compiled? I could add something in the doc about that.

jgabry:
I just pushed a commit mentioning that in the doc, but let me know if you prefer something different and I can change it. Otherwise I think this PR is ready.

jgabry:
Actually this is going to conflict with @mike-lawrence's PR #306 so let's hold off on merging this until we sort things out.

rok-cesnovar:
> I just pushed a commit mentioning that in the doc, but let me know if you prefer something different and I can change it.

Cool with that.

> Actually this is going to conflict with @mike-lawrence's PR #306 so let's hold off on merging this until we sort things out.

That PR has other conflicts we need to sort out, so we can go ahead and merge this once tests pass.

mike-lawrence:
> That PR has other conflicts we need to sort out, so we can go ahead and merge this once tests pass.

I concur!

jgabry:
Ok thanks!

rok-cesnovar:
I now realized we could alao use this in the vignette: https://mc-stan.org/cmdstanr/articles/cmdstanr-internals.html

No rush though.

jgabry:
Good point. That's a quick change so I can do that now.

PR #369, opened 2020-11-25
Add threading for variational/optimize

### LINKED ISSUE #193: Threading interface for optimization/VI
Should be done similar to how we handle it for sampling after https://github.com/stan-dev/cmdstanr/pull/185 is merged.

rok-cesnovar:
#### Summary

Fixes #193 

#### Copyright and Licensing

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -865,7 +865,7 @@ sample_method <- function(data = NULL,
   }
   if (!is.null(num_cores)) {
     warning("'num_cores' is deprecated. Please use 'parallel_chains' instead.")
-    cores <- num_cores
+    parallel_chains <- num_cores
   }
   if (!is.null(num_chains)) {
     warning("'num_chains' is deprecated. Please use 'chains' instead.")
@@ -1024,7 +1024,23 @@ optimize_method <- function(data = NULL,
                             algorithm = NULL,
                             init_alpha = NULL,
                             iter = NULL,
-                            sig_figs = NULL) {
+                            sig_figs = NULL,
+                            threads = NULL) {
+  checkmate::assert_integerish(threads, lower = 1, len = 1, null.ok = TRUE)
+  if (is.null(self$cpp_options()[["stan_threads"]])) {
+    if (!is.null(threads)) {
+      warning("'threads' is set but the model was not compiled with ",
+              "'cpp_options = list(stan_threads = TRUE)' so 'threads' will have no effect!",
+              call. = FALSE)
+      threads <- NULL
+    }
+  } else {
+    if (is.null(threads)) {
+      stop("The model was compiled with 'cpp_options = list(stan_threads = TRUE)' ",
+           "but 'threads' was not set!",
+           call. = FALSE)
+    }
+  }
   optimize_args <- OptimizeArgs$new(
     algorithm = algorithm,
     init_alpha = init_alpha,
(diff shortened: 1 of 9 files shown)


R/model.R
```diff
@@ -1024,7 +1029,23 @@ optimize_method <- function(data = NULL,
                             algorithm = NULL,
                             init_alpha = NULL,
                             iter = NULL,
-                            sig_figs = NULL) {
+                            sig_figs = NULL,
+                            threads = NULL) {
```
rok-cesnovar, line 1033: Not so sure about the name here. but `threads_per_chain` does not work for optimize/variational.

R/model.R
```diff
@@ -865,7 +865,7 @@ sample_method <- function(data = NULL,
   }
   if (!is.null(num_cores)) {
     warning("'num_cores' is deprecated. Please use 'parallel_chains' instead.")
-    cores <- num_cores
+    parallel_chains <- num_cores
```
rok-cesnovar, line 868: Unrelated to the PR, just a bug I found by chance.

R/model.R
jgabry: Yeah that's a good question. Ideally we would use the same argument name but you're right that there are no chains here. I guess `threads` is fine. @mitzimorris or @bbbales2 any thoughts on what name we should use here?

R/model.R
```diff
@@ -1024,7 +1029,23 @@ optimize_method <- function(data = NULL,
...
                             init_alpha = NULL,
                             iter = NULL,
-                            sig_figs = NULL) {
+                            sig_figs = NULL,
+                            threads = NULL) {
+  checkmate::assert_integerish(threads, lower = 1, len = 1, null.ok = TRUE)
+  if (is.null(self$cpp_options()[["stan_threads"]])) {
+    if (!is.null(threads)) {
+      warning("'threads' is set but the model was not compiled with ",
+              "'cpp_options = list(stan_threads = TRUE)' so 'threads' will have no effect!",
```
jgabry, line 1038: This doesn't have to be part of this PR, but I wonder if we should do something similar to the `pedantic` and `include_paths` arguments and just provide an argument to `compile()` to turn on threading instead of forcing the use of `cpp_options()`. Not totally sure, but something to ponder.

jgabry:
I put a few comments, but this looks good to me, thanks!

jgabry (approved):

R/model.R
rok-cesnovar: Yeah, we had that originally and it's now deprecated. Let's tackle that in a separate issue/PR

rok-cesnovar:
Thanks. I guess the only thing to discuss is whether `threads = X` is a good argument for `optimization`/`variational`. It's a different name than sampling, but `threads_per_chain `does not work for `optimization`.

R/model.R
jgabry: Oh yeah I forgot about that. Maybe we were right originally! But yeah we can deal with that later.

jgabry:
> Thanks. I guess the only thing to discuss is whether `threads = X` is a good argument for `optimization`/`variational`. It's a different name than sampling, but `threads_per_chain `does not work for `optimization`.

I think `threads` is good but we can wait a little while to see if anyone has a different suggestion.

bbbales2:
I'm happy with threads. I hope in the future we do a multiple-runs sorta VI (everything should be checked -- https://arxiv.org/abs/2009.00666), but that isn't here now.

jgabry:
Yeah I think multiple-run VI would be a good thing in the future. I guess we can adjust as necessary when that time comes.

jgabry:
Seems like we're going with `threads` at least until there's multi-run VI. That seems fine so I'm going to go ahead merge this because the only thing holding it up was that argument name.

PR #386, opened 2020-12-04
Loo method for CmdStanMCMC objects

### LINKED ISSUE #366: Add loo method
Implement loo so that the following works for cmdstanr fit object similarly as for rstan object
```
loo(fit)
```

https://discourse.mc-stan.org/t/using-loo-with-cmdstanr/17230/2 suggest to use
```
loo(fit$draws("log_lik"))
```
but this misses `r_eff` and should be
```
loo(fit$draws("log_lik"), r_eff=relative_eff(fit$draws("log_lik")))
```
which is much more difficult to remember than `loo(fit)`

Seems like an easy feature to add?

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #366 

Adds `fit$loo()` method for CmdStanMCMC objects. Will add some review comments to point out a few decisions we need to make. 

#### Copyright and Licensing

**Columbia University**


R/fit.R
```diff
@@ -953,7 +998,24 @@ CmdStanMCMC <- R6::R6Class(
...
+        stop("Please install the loo package to use this method.", call. = FALSE)
+      }
+      LLarray <- self$draws(variables)
+      if (is.logical(r_eff)) {
+        if (isTRUE(r_eff)) {
+          r_eff_cores <- list(...)[["cores"]] %||% getOption("mc.cores", 1)
+          r_eff <- loo::relative_eff(exp(LLarray), cores = r_eff_cores)
+        } else {
+          r_eff <- NULL
+        }
```
jgabry, line 1014: This gives the user the choice of computing `r_eff` automatically (the default), turning off `r_eff` computation (by setting `r_eff=FALSE`) or supplying their own `r_eff` values that will be passed to loo. @avehtari This seems like the most flexible approach and if you like it I can also add the same thing in the method in RStan.

R/fit.R
```diff
@@ -953,7 +998,24 @@ CmdStanMCMC <- R6::R6Class(
...
+      }
+      loo::loo.array(LLarray, r_eff = r_eff, ...)
```
jgabry, line 1016: @rok-cesnovar @avehtari Right now the loo method just returns the loo object, but we could also store the results in the fit object so that repeated calls to loo don't need to do the computation again (perhaps unless the user selects an option to force it). That could look roughly like this: 

```r
loo = function(..., force = FALSE) {
  ...
  if (is.null(private$loo_object_) || force) {
    private$loo_object_ <- loo::loo.array(LLarray, r_eff = r_eff, ...)
  }
  return(private$loo_object_)
}
```

I'm not sure if we want to do this or not. On the one hand it's pretty convenient, on the other hand users can just assign the result of `loo$fit()` to some object to avoid recomputing so maybe we don't need to do this. What do you think?

rok-cesnovar: I think storing it separately is the way to go.

jgabry: Ok I agree, let's not try to get too fancy here.

rok-cesnovar (approved):
This looks good to me. Will also let @avehtari comment if he has time.

R/fit.R
avehtari: This is good!

R/fit.R
avehtari: btw. brms has add_criterion function for adding to the fit object, but we don't need to do that here or now

avehtari (approved):
I checked the edits and all look good.

PR #392, opened 2020-12-08
verbose mode and replace regexpr

### LINKED ISSUE #385: Consistent usage of grepl and regexpr
Sometimes we use `grepl(...)` and sometimes we use `regexpr(...) > 0`. Not a big deal but would be preferable to unify.

### LINKED ISSUE #391: Verbose mode
I think it would make sense to have something like a verbose mode that could be used for troubleshooting, debugging, and developing with cmdstanr. I think an option would be just fine. This is not something a regular user would need.
Like 
```R
options("cmdstanr_verbose"= TRUE)
```
would print all commands and all output from processx. This came up when debugging a problem with @yizhang-yiz [here](https://github.com/stan-dev/cmdstanr/pull/350#issuecomment-740786862)

rok-cesnovar:
#### Summary

Fixes #385 
Fixes #391 

I was in a slam dunk issue squashing mood so decided to squash these two.

#### Copyright and Licensing

Rok Češnovar


jgabry:
> I was in a slam dunk issue squashing mood so decided to squash these two.

Nice, thanks!

This looks great except that the uses of `is_verbose_mode` all need to be `is_verbose_mode()` so that it actually calls the function. That's probably why there are errors. I can go ahead and make that change now.

R/utils.R
```diff
@@ -14,6 +14,10 @@ os_is_macos <- function() {
   isTRUE(Sys.info()[["sysname"]] == "Darwin")
 }
 
+is_verbose_mode <- function() {
+  getOption("cmdstanr_verbose", default = FALSE)
```
jgabry, line 18: The way you had this function was totally fine, but to simplify things I changed it to use `getOption()`, which conveniently handles the `NULL` case with the `default` argument.

jgabry:
Ok I changed `is_verbose_mode` to `is_verbose_mode()` so we'll see if all the tests pass now.

R/utils.R
rok-cesnovar: Ah didnt know that. Thanks!

rok-cesnovar:
Maybe we add it to the internals vignette? Do we want to do it here or add to the vignette list?

jgabry:
> Maybe we add it to the internals vignette? Do we want to do it here or add to the vignette list?

Sounds good. Either way is fine by me.

rok-cesnovar:
Will add it here tomorrow and then merge.

jgabry:
Ok cool, thanks!

rok-cesnovar:
@jgabry I updated the vignette, check whenever you have a few moments.

jgabry:
Looks great! I'm going to quickly add an example to the section you created so people can see what the verbose mode output looks like and then I'll merge this.

PR #406, opened 2020-12-15
only look for "PCH file" not full message

### LINKED ISSUE #400: Change how we detect PCH file problems?
Currently we check for errors related to the precompiled header file like this: 

https://github.com/stan-dev/cmdstanr/blob/d0c0f7797f3e8715bd85cafc4c31d73b0884ef95/R/model.R#L508-L511

But the following error messages that are different were [just reported on the forums](https://discourse.mc-stan.org/t/struggling-to-get-cmdstan-working-on-mac-os-big-sur/19774/4?u=jonah): 

```
error: stack protector mode differs in PCH file vs. current file
error: PCH file was compiled for the target 'x86_64-apple-macosx10.9.0' but the current translation unit is being compiled for target 'x86_64-apple-macosx11.0.0'
2 errors generated.
```

Maybe it makes sense for us to only grep for `"PCH file"` so we have a better chance of catching these errors. I think that would be fine unless it's possible for `"PCH file"` to also appear in a message that we shouldn't be warning the user about.

jgabry:
fixes #400

#### Submission Checklist

- [x] Run unit tests

#### Summary

When detecting problems with the precompiled header, instead of looking for the full message just look for "PCH file" to hopefully catch more cases. We saw an [example on the forum](https://discourse.mc-stan.org/t/struggling-to-get-cmdstan-working-on-mac-os-big-sur/19774/4?u=jonah) where the error message contained "PCH file" but wasn't the same as the full messages we were looking for. 

@rok-cesnovar Is it possible for "PCH file" to appear in a different message that we wouldn't want to catch? If so then maybe we don't want to make this switch, but I'm thinking that's very unlikely right? Just in case, I made the warning message slightly more speculative. 

#### Copyright and Licensing

**Columbia University**

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -505,10 +505,13 @@ compile_method <- function(quiet = TRUE,
     spinner = quiet && interactive(),
     stderr_line_callback = function(x,p) {
       if (!startsWith(x, paste0(make_cmd(), ": *** No rule to make target"))) message(x)
-      if (grepl("PCH file uses an older PCH format that is no longer supported", x, fixed = TRUE)
-          || grepl("PCH file built from a different branch", x, fixed = TRUE)) {
-        warning("Cmdstan encountered an issue with an outdated precompiled header (PCH). Run rebuild_cmdstan() to rebuild the PCH files.\n",
-        "If the issue persists please open a bug report.")
+      if (grepl("PCH file", x)) {
+        warning(
+          "CmdStan encountered an issue with an outdated precompiled header (PCH). ",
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
> Is it possible for "PCH file" to appear in a different message that we wouldn't want to catch?

I think if PCH file occurs in stderr, its most likely a sign that a rebuild of cmdstan is either necessary to make it work or at least suggested.

rok-cesnovar (approved):

Issue #81, opened 2019-11-13
Use new R6 roxygen support


jgabry:
In https://github.com/stan-dev/cmdstanr/issues/78#issuecomment-553179118 @stevebronder linked to the announcement of support for R6 documentation in roxygen2. 
This is good news! We should update. 

https://www.tidyverse.org/blog/2019/11/roxygen2-7-0-0/#r6-documentation

jgabry:
Unfortunately it doesn't yet support documentation for methods added dynamically via `$set()` (which we use a bunch for longer method definitions). The new release of roxygen2 only supports inline doc for methods defined when the class is initially defined.

We could either change our code to define all methods initially with the class (suboptimal since doing what we currently do saves a lot of code duplication when defining shared methods) or wait for a roxygen2 that supports what we need (it's on their to-do list: https://github.com/r-lib/roxygen2/issues/931).

PR #408, opened 2020-12-16
Use `@param` when documenting arguments to R6 methods

### LINKED ISSUE #367: Formal @param entries for method arguments
It would be nice if methods like `$sample()` had formal `@param` entries for the arguments. Currently, they are documented in an informal list:

https://github.com/stan-dev/cmdstanr/blob/3b7fbe551617c5c6c30883df501cbb34ea07b865/R/model.R#L670-L772

Formal `@params` would allow packages built on top of `cmdstanr` to use `@inheritParams`. 

The reason I bring this up is that I am currently building a Bayesian data analysis framework on top of `cmdstanr` and [`targets`](https://github.com/wlandau/targets). The package is called `stantargets`, and I hope to make it open source within the next week or two (currently waiting for the required approvals).

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #367. 

Uses `@param` in empty docstrings (no function) when documenting CmdStanModel and CmdStanFit methods. This will allow other developers like @wlandau to use roxygen2's `@inheritParams` to import our doc into packages that use cmdstanr. This PR is unrelated to #81 and doesn't use roxygen2's limited R6 support.

I think this is a good idea but there are some downsides to doing this that we should also consider: 

* For regular functions in an R package (e.g. functions like `write_stan_json()` or `cmdstan_model()`) the Usage section in the doc (showing the function signature) comes before the Arguments section. This is done automatically by roxygen2. However, when using empty docstrings with `@param` to document the R6 methods the Arguments section is automatically generated but we still have to manually generate the Usage section. This results in the Usage section coming *after* the Arguments section. I really don't like this but I don't know if it's a dealbreaker. (I'm also not sure if there's a way around the ordering that roxygen2 is generating. Maybe we can modify it.)

* As far as I know, we can't have subsections in the Arguments section in the doc. When using `@param` the Arguments section is automatically generated. Previously we were breaking it up into subsections to separate args shared by all model fitting methods and method specific args (and in the latter we were splitting again into args in CmdStanR but not CmdStan vs args in both CmdStanR and CmdStan). When using `@param` it all becomes one long list of arguments with no breaking it up with subsection descriptions. 

#### Copyright and Licensing

**Columbia University**


jgabry:
@rok-cesnovar I'll generate the website docs once we're ready to merge this. Or if you want to review what that looks like I can also generate them now so you can just use `pkgdown::preview_site()` to view the changes.

man-roxygen/model-sample-args.R
```diff
@@ -0,0 +1,96 @@
+#' @param chains (positive integer) The number of Markov chains to run. The
+#'   default is 4.
+#' @param parallel_chains (positive integer) The _maximum_ number of MCMC chains
+#'   to run in parallel. If `parallel_chains` is not specified then the default
+#'   is to look for the option `"mc.cores"`, which can be set for an entire \R
+#'   session by `options(mc.cores=value)`. If the `"mc.cores"` option has not
+#'   been set then the default is `1`. This argument is available for
+#'   the `$sample()` method but not for the `$sample_mpi()` method.
```
jgabry, line 8: since both sample and sample_mpi now use this same file I just added a sentence to the arguments that sample_mpi doesn't support

man-roxygen/model-sample-args.R
```diff
@@ -0,0 +1,96 @@
...
+#'   [compiled][model-method-compile] with threading support, the number of
+#'   threads to use in parallelized sections _within_ an MCMC chain (e.g., when
+#'   using the Stan functions `reduce_sum()` or `map_rect()`). This is in
+#'   contrast with `parallel_chains`, which specifies the number of chains to
+#'   run in parallel. The actual number of CPU cores used use is
+#'   `parallel_chains*threads_per_chain`. For an example of using threading see
+#'   the Stan case study [Reduce Sum: A Minimal
+#'   Example](https://mc-stan.org/users/documentation/case-studies/reduce_sum_tutorial.html).
+#'   This argument is available for the `$sample()` method but not for the
+#'   `$sample_mpi()` method.
```
jgabry, line 22: same as above

R/model.R
```diff
@@ -722,113 +722,9 @@ CmdStanModel$set("public", name = "check_syntax", value = check_syntax_method)
...
-#'
-#'   * `chains`: (positive integer) The number of Markov chains to run. The
-#'   default is 4.
-#'
-#'   * `parallel_chains`: (positive integer) The _maximum_ number of MCMC chains
-#'   to run in parallel. If `parallel_chains` is not specified then the default
-#'   is to look for the option `"mc.cores"`, which can be set for an entire \R
-#'   session by `options(mc.cores=value)`. If the `"mc.cores"` option has not
-#'   been set then the default is `1`.
+#' @template model-sample-args
```
jgabry, line 725: I moved all the arguments unique to the sample method into a separate roxygen2 template. That way sample and sample_mpi can share them.

R/model.R
```diff
@@ -1044,19 +940,16 @@ CmdStanModel$set("public", name = "sample", value = sample_method)
...
-#'   * `mpi_args`: (list) A list of arguments to use when launching MPI processes.
-#'     For example, `mpi_args = list("n" = 4)` launches the executable as
-#'     `mpiexec -n 4 model_executable`, followed by CmdStan arguments
-#'     for the model executable.
-#'
-#'    All other arguments are the same as for [`$sample()`][model-method-sample]
-#'    except `$sample_mpi()` does not have arguments `threads_per_chain` or
-#'    `parallel_chains`.
+#' @template model-common-args
+#' @template model-sample-args
```
jgabry: sample_mpi and sample share same args template

wlandau:
Thank you SO much for this, @jgabry! It will be a whole lot easier to keep up with changes in `cmdstanr` for [`stantargets`](https://wlandau.github.io/stantargets). I am using your PR in [this `stantargets` branch](https://github.com/wlandau/stantargets/tree/inheritParams), and `R CMD check` seems satisfied with the docs.

jgabry:
> Thank you SO much for this, @jgabry! It will be a whole lot easier to keep up with changes in `cmdstanr` for [`stantargets`](https://wlandau.github.io/stantargets). I am using your PR in [this `stantargets` branch](https://github.com/wlandau/stantargets/tree/inheritParams), and `R CMD check` seems totally satisfied with the docs.

Awesome! And yeah R CMD check seems to be fine. At this point the only thing I'm worried about is this: 

> For regular functions in an R package (e.g. functions like write_stan_json() or cmdstan_model()) the Usage section in the doc (showing the function signature) comes before the Arguments section. This is done automatically by roxygen2. However, when using empty docstrings with @param to document the R6 methods the Arguments section is automatically generated but we still have to manually generate the Usage section. This results in the Usage section coming after the Arguments section. I really don't like this but I don't know if it's a dealbreaker. (I'm also not sure if there's a way around the ordering that roxygen2 is generating. Maybe we can modify it.)

Even if I put the manually created Usage section before the `@param` entries roxygen2 will put the automatically generated Arguments section before the manually created Usage section. I suppose we could write some function to parse the generated Rd files and change the order of the sections after roxygen2 generates them, but that could be error prone and doesn't seem like a great solution. I know there are ways to extend roxygen2 but I haven't figured out yet if that would allow something like changing the order of the sections.

rok-cesnovar (approved):
Nice! Thanks!

jgabry:
Actually, I have an idea about how to deal with the usage sections. Not sure if it will work but I will try now and if it does I'll explain.

R/model.R
```diff
@@ -357,17 +343,15 @@ CmdStanModel <- R6::R6Class(
...
-compile_method <- function(quiet = TRUE,
-                           dir = NULL,
-                           pedantic = FALSE,
-                           include_paths = NULL,
-                           cpp_options = list(),
-                           stanc_options = list(),
-                           force_recompile = FALSE,
-                           #deprecated
-                           threads = FALSE) {
+compile <- function(quiet = TRUE,
```
jgabry, line 346: Taking the compile method as an example (but this works with all the other methods too), to fix the Usage section problem we can switch to defining a function `compile` (instead of `compile_method`) and we create doc for that `compile` function (instead of using `NULL` at the end of the docstring). But then we do _not_ export the `compile` function and we do _not_ use `@noRd`. This results in roxygen2 creating the Rd file as if the function were exported but it's not actually exported. We then use `CmdStanModel$set("public", name = "compile", value = compile)`. 

This also has the benefit that `sample_mpi` can use `inheritParams` to inherit from `sample` instead of having to move to roxygen templates.

jgabry: I still need to do this for all the CmdStanFit methods, but it seems to be working for all the CmdStanModel methods!

R/model.R
jgabry: This isn't needed anymore. We can use `@inheritParams` after the change I made to fix the usage sections. (see comments below)

jgabry:
Ok I was able to fix the Usage section problem the way I mentioned for the compile method above. I think this is now ready. 

@wlandau Can you double check that this still works well for you after the changes I made to fix the Usage section issue? I don't think it should change anything for you but it would be good to know if it still works and if R cmd check still is fine for you. Thank you!

wlandau:
Just ran `R CMD check` again on https://github.com/stan-dev/cmdstanr/commit/8810cd7117f9b7010b71d7ab78d877cb952b2add and found no issues with the docs.

jgabry:
Thanks! R cmd check runs fine for me too. Thanks again for suggesting this.

jgabry:
@rok-cesnovar Are you ok if I make this (and everything else that's been merged recently) into a 0.2.3 release?

PR #412, opened 2020-12-18
CmdStanMCMC/MLE/VB from read_cmdstan_csv ouput

### LINKED ISSUE #411: Create CmdStanMCMC R6 object from CSV
**Is your feature request related to a problem? Please describe.**
Currently no sane way to create `CmdStanMCMC` object as returned by `$sample` command by reading from CSV. 

**Describe the solution you'd like**
Something like
```r
fit <- to.cmdstan.mcmc(read_cmdstan_csv(c(file1, file2, file3, file4)))
```
so that later `fit` can be used for various actions such as
```r
draws <- fit$draws()
```

**Describe alternatives you've considered**
None so far.

**Additional context**
One motivation is to help user share/debug outputs, as indicated [here](https://discourse.mc-stan.org/t/draws-method-in-cmdstanr-is-still-slow/19490/5?u=yizhang). Another one is to port past cmdstan output to be postprocessed.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #411 

Allows creating CmdStanMCMC, CmdStanVB, and CmdStanMLE objects from the list returned by `read_cmdstan_csv()`. 
In reality these objects are not exactly the same because they don't have an associated CmdStanRun object. This means that methods like `cmdstan_diagnose()` and `save_output_files()` are unavailable, but the most important methods (e.g., `draws()`, `summary()`) work fine. 

#### Copyright and Licensing

**Columbia University**


R/fit.R
```diff
@@ -15,9 +15,6 @@ CmdStanFit <- R6::R6Class(
       self$runset$num_procs()
     },
     print = function(variables = NULL, ..., digits = 2, max_rows = 10) {
-      if (!length(self$output_files(include_failed = FALSE))) {
-        stop("Fitting failed. Unable to print.", call. = FALSE)
-      }
```
jgabry, line 20: Need to remove this so it doesn't error if output files aren't available. Otherwise we can't use this with the output from read_cmdstan_csv(). 

However, this doesn't risk `print()` failing to catch an important error because the same error will be caught by the `metadata()` method that is called internally by `print()`. `metadata()` is smarter about throwing the error because it only throws the error if the files are missing AND `metadata()` hasn't already been read in.

R/fit.R
```diff
@@ -330,7 +327,7 @@ CmdStanFit$set("public", name = "lp", value = lp)
 #'
 summary <- function(variables = NULL, ...) {
   draws <- self$draws(variables)
-  if (self$runset$method() == "sample") {
+  if (self$metadata()$method == "sample") {
```
jgabry, line 330: We need to check the method via metadata()$method and not runset$method() or it won't work with the objects created after read_cmdstan_csv

R/fit.R
```diff
@@ -599,10 +596,10 @@ CmdStanFit$set("public", name = "output", value = output)
 #' }
 #'
 metadata <- function() {
-  if (!length(self$output_files(include_failed = FALSE))) {
-    stop("Fitting failed. Unable to retrieve the metadata.", call. = FALSE)
-  }
   if (is.null(private$metadata_)) {
+    if (!length(self$output_files(include_failed = FALSE))) {
+      stop("Fitting failed. Unable to retrieve the metadata.", call. = FALSE)
+    }
```
jgabry, line 602: Move this error after the `is.null` check so that it's only thrown if metadata hasn't already been read in.

R/fit.R
```diff
@@ -705,17 +702,17 @@ CmdStanMCMC <- R6::R6Class(
       } else {
         if (self$runset$args$validate_csv) {
           fixed_param <- runset$args$method_args$fixed_param
-          data_csv <- read_cmdstan_csv(
+          csv_contents <- read_cmdstan_csv(
```
jgabry, line 705: This and every other change to `csv_contents` isn't important, it just makes naming consistent throughout our code

R/fit.R
```diff
@@ -730,9 +727,6 @@ CmdStanMCMC <- R6::R6Class(
 
     # override the CmdStanFit draws method
     draws = function(variables = NULL, inc_warmup = FALSE) {
-      if (!length(self$output_files(include_failed = FALSE))) {
```
jgabry, line 733: This error will also be thrown by `private$read_csv_` so it is unnecessary here. We also need to remove it or it will result in errors after creating the objects from read_cmdstan_csv.

R/fit.R
```diff
@@ -773,45 +767,45 @@ CmdStanMCMC <- R6::R6Class(
       if (!length(self$output_files(include_failed = FALSE))) {
         stop("No chains finished successfully. Unable to retrieve the draws.", call. = FALSE)
       }
-      data_csv <- read_cmdstan_csv(
+      csv_contents <- read_cmdstan_csv(
```
jgabry, line 770: this all and cases below are just name changes

R/read_csv.R
```diff
@@ -370,9 +373,112 @@ read_sample_csv <- function(files,
...
+#' @param check_diagnostics Should diagnostic checks be performed? The default
+#'   is `TRUE` but set to `FALSE` to avoid checking diagnostics like
+#'   divergences.
+#' @examples
+#' \dontrun{
+#' # create a CmdStanMCMC object
+#' x <- read_cmdstan_csv(fit$output_files())
+#' fit <- as_cmdstan_mcmc(x)
+#' }
+as_cmdstan_mcmc <- function(csv_contents, check_diagnostics = TRUE) {
```
jgabry, line 388: What do you think of the names `as_cmdstan_mcmc()`, `as_cmdstan_mle()`, `as_cmdstan_vb()`?

R/read_csv.R
```diff
@@ -370,9 +373,112 @@ read_sample_csv <- function(files,
...
+#' instead of from a CmdStanRun object
+#'
+#' The resulting object has fewer methods than a CmdStanMCMC/MLE/VB object
+#' because it doesn't have access to the CSV files directly.
+#'
+#' @noRd
+#'
+CmdStanMCMC2 <- R6::R6Class(
+  classname = "CmdStanMCMC2",
+  inherit = CmdStanMCMC,
```
jgabry, line 418: Don't actually create CmdStanMCMC objects but CmdStanMCMC2 objects that inherit from CmdStanMCMC. Same approach with MLE and VB below. This avoids having to create the runset object because we can override the `initialize` method.  

Suggestions for better class name than `CmdStanMCMC2`?

R/read_csv.R
```diff
@@ -370,9 +373,112 @@ read_sample_csv <- function(files,
...
+    "init",
+    "output",
+    "return_codes",
+    "time"
+  )
+for (method in unavailable_methods_CmdStanFit2) {
+  CmdStanMCMC2$set("public", name = method, value = error_unavailable_CmdStanFit2)
+  CmdStanMLE2$set("public", name = method, value = error_unavailable_CmdStanFit2)
+  CmdStanVB2$set("public", name = method, value = error_unavailable_CmdStanFit2)
+}
```
jgabry, line 479: This makes sure that the methods that won't work have informative error messages.

jgabry (commented):
@rok-cesnovar Just added a bunch of review comments that hopefully help explain the changes.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: cmdstanr
 Title: R Interface to 'CmdStan'
-Version: 0.3.0
+Version: 0.3.0.9000
```
jgabry, line 3: bumping to dev version

R/read_csv.R
```diff
@@ -370,9 +388,124 @@ read_sample_csv <- function(files,
...
+# CmdStanFit2 -------------------------------------------------------------
+#' Create a CmdStanMCMC/MLE/VB-ish objects from `read_cmdstan_csv()` output
+#' instead of from a CmdStanRun object
+#'
+#' The resulting object has fewer methods than a CmdStanMCMC/MLE/VB object
+#' because it doesn't have access to the CSV files directly.
+#'
+#' @noRd
+#'
+CmdStanMCMC2 <- R6::R6Class(
```
jgabry, line 429: I put these in the read_csv.R file because of the use case but we could also but these in the fit.R file with the definition of CmdStanMCMC, CmdStanMLE, etc. Preference?

jgabry:
@yizhang-yiz I think this is now ready to play around with if you want to try it out.

yizhang-yiz:
```r
> fit <- cmdstanr::as_cmdstan_mcmc(dir(pattern=".csv", full.name=TRUE))
```
This works
```r
bayesplot::mcmc_dens_overlay(fit$draws(), pars = c("CL", "Q", "V1", "V2", "ka", "sigma"), facet_args = list(nrow = 2))
```
but this doesn't
```r
> print(fit$time(), digits = 3)
Error: This method is not available for objects created using as_cmdstan_mcmc().
```
I wonder what the rationale is.

jgabry:
> I wonder what the rationale is.

No good one other than it would take more effort to get that to work. `read_cmdstan_csv()` doesn't currently give you the timing information. When model fitting happens we create a CmdStanRun object

https://github.com/stan-dev/cmdstanr/blob/46ae80f12881356e9152edd3881457b6da40016f/R/model.R#L804-L806

which that handles preparing the call to cmdstan, setting up the external processes, and then running cmdstan. It's also the object that report the run times (we don't wait for reading in the csv to report the time). The way I created `as_cmdstan_mcmc()` it avoids creating this object entirely and just uses the output from `read_cmdstan_csv()`. So to get time information with `as_cmdstan_mcmc()` we'd just need to add it to `read_cmdstan_csv()` (probably a good idea anyway).

rok-cesnovar:
> we'd just need to add it to read_cmdstan_csv() (probably a good idea anyway). 

Agree. Will do that separately.

R/read_csv.R
rok-cesnovar: I like them, but have another idea. What about `as_cmdstan_fit()` that does all three?

R/read_csv.R
rok-cesnovar: What aobut CmdStanMCMCCSV, CmdStanMCMC_CSV? I am bad for names.

R/read_csv.R
rok-cesnovar: I think having them here is fine.

R/read_csv.R
jgabry: Good idea. I just changed it.

jgabry:
I changed to `as_cmdstan_fit()`, good idea. I haven't changed the class names yet but I'm leaning towards `CmdStanMCMC_CSV`. That's definitely better than `CmdStanMCMC2`. 

Let's wait until  #414 is merged and then we can update this PR to include `time` before merging it.

jgabry:
Ok I changed the class names to `CmdStanMCMC_CSV`, `CmdStanMLE_CSV`, `CmdStanVB_CSV` and I added the `time()` method for `CmdStanMCMC_CSV` after pulling in the changes from #414.

rok-cesnovar (approved):
Looks good to me. Thanks.

PR #470, opened 2021-03-16
global max_rows option for print method

### LINKED ISSUE #469: global max_rows option for print method
Suggested by @bob-carpenter. We should allow setting a global option in addition to the `max_rows` argument. We could use `cmdstanr_max_rows` for the name. To do this just requires changing the default value of the `max_rows` argument from `10` to `getOption("cmdstanr_max_rows", default = 10)`.  The user could then do `options(cmdstanr_max_rows = whatever)` to set it for an entire session.

jgabry:
closes #469

#### Submission Checklist

- [x] Run unit tests

#### Summary

Allows `options(cmdstanr_max_rows = value)` to set `max_rows` argument of print method for an entire R session (or until changing the option). 

#### Copyright and Licensing

**Columbia University**

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -17,7 +17,7 @@ CmdStanFit <- R6::R6Class(
     num_procs = function() {
       self$runset$num_procs()
     },
-    print = function(variables = NULL, ..., digits = 2, max_rows = 10) {
+    print = function(variables = NULL, ..., digits = 2, max_rows = getOption("cmdstanr_max_rows", 10)) {
       if (is.null(private$draws_) &&
           !length(self$output_files(include_failed = FALSE))) {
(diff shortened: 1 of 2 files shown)


rok-cesnovar (approved):
Looks good. One suggestion or rather an idea. Maybe we go with the name

`cmdstanr.max_rows `

I know brms uses options like `brms.normalize` and such. You are obviously the R packages expert here so you would know better than me what is more common.

jgabry:
Yeah I've seen both conventions used. I don't have a strong preference, I just used `cmdstanr_max_rows` to be consistent with the `cmdstanr_verbose` option that we already have.

rok-cesnovar:
Cool, if both are common then lets merge this.

PR #477, opened 2021-03-28
Concise progress

mike-lawrence:
#### Submission Checklist

- [X] Run unit tests

#### Summary

No more console spam! Tried to make this as minimal as possible, both in terms of code changes and visual output of the progress information. I'll be following this with a PR that adds an estimated-time-remaining, and then another PR with prettier-but-trickier-to-implement-cross-consoles multi-line progress indicators.

#### Copyright and Licensing

Mike Lawrence

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -816,6 +816,8 @@ sample <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
+    iter = ifelse(is.null(sample_args$iter_warmup),1e3,sample_args$iter_warmup) +
+      ifelse(is.null(sample_args$iter_sampling),1e3,sample_args$iter_sampling),
     parallel_procs = parallel_chains,
     threads_per_proc = threads_per_chain,
     show_stderr_messages = show_messages
@@ -954,6 +956,8 @@ sample_mpi <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
+    iter = ifelse(is.null(sample_args$iter_warmup),1e3,sample_args$iter_warmup) +
+      ifelse(is.null(sample_args$iter_sampling),1e3,sample_args$iter_sampling),
     parallel_procs = 1,
     show_stderr_messages = show_messages
   )
@@ -1073,6 +1077,7 @@ optimize <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
+    iter = ifelse(is.null(optimize_args$iter),1e3,optimize_args$iter),
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
@@ -1197,6 +1202,7 @@ variational <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
+    iter = ifelse(is.null(variational_args$iter),1e3,variational_args$iter),
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
--- a/R/model.R
+++ b/R/model.R
@@ -816,7 +816,7 @@ sample <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
-iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
+    iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
     parallel_procs = parallel_chains,
     threads_per_proc = threads_per_chain,
     show_stderr_messages = show_messages
@@ -955,7 +955,7 @@ sample_mpi <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
-iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
+    iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
     parallel_procs = 1,
     show_stderr_messages = show_messages
   )
@@ -1075,7 +1075,7 @@ optimize <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
-    iter = optimize_args$iter %||% 1e3,
+    iter = optimize_args$iter %||% 2e3,
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
@@ -1200,7 +1200,7 @@ variational <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
-    iter = variational_args$iter %||% 1e3
+    iter = variational_args$iter %||% 1e4
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
--- a/R/model.R
+++ b/R/model.R
@@ -816,8 +816,7 @@ sample <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
-    iter = ifelse(is.null(sample_args$iter_warmup),1e3,sample_args$iter_warmup) +
-      ifelse(is.null(sample_args$iter_sampling),1e3,sample_args$iter_sampling),
+iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
     parallel_procs = parallel_chains,
     threads_per_proc = threads_per_chain,
     show_stderr_messages = show_messages
@@ -956,8 +955,7 @@ sample_mpi <- function(data = NULL,
   )
   cmdstan_procs <- CmdStanMCMCProcs$new(
     num_procs = chains,
-    iter = ifelse(is.null(sample_args$iter_warmup),1e3,sample_args$iter_warmup) +
-      ifelse(is.null(sample_args$iter_sampling),1e3,sample_args$iter_sampling),
+iter = sample_args$iter_warmup %||% 1e3 + sample_args$iter_sampling %||% 1e3,
     parallel_procs = 1,
     show_stderr_messages = show_messages
   )
@@ -1077,7 +1075,7 @@ optimize <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
-    iter = ifelse(is.null(optimize_args$iter),1e3,optimize_args$iter),
+    iter = optimize_args$iter %||% 1e3,
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
@@ -1202,7 +1200,7 @@ variational <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
-    iter = ifelse(is.null(variational_args$iter),1e3,variational_args$iter),
+    iter = variational_args$iter %||% 1e3
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
--- a/R/model.R
+++ b/R/model.R
@@ -1200,7 +1200,7 @@ variational <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
-    iter = variational_args$iter %||% 1e4
+    iter = variational_args$iter %||% 1e4,
     show_stdout_messages = (is.null(refresh) || refresh != 0),
     threads_per_proc = threads
   )
--- a/R/run.R
+++ b/R/run.R
@@ -533,17 +533,17 @@ CmdStanProcs <- R6::R6Class(
   classname = "CmdStanProcs",
   public = list(
     initialize = function(num_procs,
-                          iter,
+                          iter = NULL,
                           parallel_procs = NULL,
                           threads_per_proc = NULL,
                           show_stderr_messages = TRUE,
                           show_stdout_messages = TRUE) {
       checkmate::assert_integerish(num_procs, lower = 1, len = 1, any.missing = FALSE)
-      checkmate::assert_integerish(iter, lower = 1, len = 1, any.missing = FALSE)
+      checkmate::assert_integerish(iter, lower = 1, len = 1, null.ok = TRUE)
       checkmate::assert_integerish(parallel_procs, lower = 1, len = 1, any.missing = FALSE, null.ok = TRUE)
       checkmate::assert_integerish(threads_per_proc, lower = 1, len = 1, null.ok = TRUE)
       private$num_procs_ <- as.integer(num_procs)
-      private$iter_ <- as.integer(iter)
+      private$iter_ <- iter
       if (is.null(parallel_procs)) {
         private$parallel_procs_ <- private$num_procs_
       } else {
--- a/R/run.R
+++ b/R/run.R
@@ -772,7 +772,8 @@ CmdStanProcs <- R6::R6Class(
           }
           if (grepl("Iteration:", line, perl = TRUE)) {
             regex_match <- regexpr('[0-9]+\\s+/',line)
-            private$iter_done_[[id]] <- as.numeric(substr(line, regex_match, regex_match + attr(regex_match, 'match.length') - 3))
+            iter_done_string <- substr(line, regex_match, regex_match + attr(regex_match, 'match.length') - 3)
+            private$iter_done_[[id]] <- as.numeric(iter_done_string)
             ignore_line <- TRUE
           }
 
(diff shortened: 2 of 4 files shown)


mike-lawrence:
What it looks like: 
[asciicast](https://asciinema.org/a/JHFiWmUKA0hDDQOxdk5JD0701)

rok-cesnovar:
The example looks really nice!! Thanks for working on this Mike! Will take a look at the code tomorrow. 

Couple quick questions: 
- what happens if a informational message pops up (like a nan warning or something like that)
- what happens if we have a print in our model (I understand those will mess with any progress reporting)

mike-lawrence:
>     * what happens if a informational message pops up (like a nan warning or something like that)
>     * what happens if we have a print in our model (I understand those will mess with any progress reporting)

Assuming other functions/programs printing lines to the console will end in a newline character (or at least do not include a carriage-return or other control sequences) they should remain on screen amid continued updating of the progress indicator. (Technically any non-progress line will overwrite the last progress info, as I put a carriage-return at the end of the progress print, but the next progress print will appear at the bottom again.) The only lines that are skipped are those that grep-match the string `Iteration:`.

mike-lawrence:
Oh, though note that I also took off the call to `report_time()` in `check_finished()` as I felt that it was an unnecessary output.

rok-cesnovar (commented):
A couple questions/comments. Will review the behaviour on Windows/Linux later today.

R/run.R
```diff
@@ -742,16 +770,23 @@ CmdStanProcs <- R6::R6Class(
...
-          } else if ((private$show_stdout_messages_ && private$proc_state_[[id]] >= 3) || is_verbose_mode()) {
+          } else if ((private$show_stdout_messages_ && private$proc_state_[[id]] >= 3 && !ignore_line) || is_verbose_mode()) {
             cat(line, collapse = "\n")
           }
         } else {
           # after the metadata is printed and we found a blank line
           # this represents the start of fitting
           if (self$proc_state(id) == 2.5) {
-              self$set_proc_state(id, new_state = 3)
+            private$iter_done_[[id]] = 0
```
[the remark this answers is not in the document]
mike-lawrence: Correct, unneeded.

R/run.R
```diff
@@ -834,9 +871,15 @@ CmdStanMCMCProcs <- R6::R6Class(
           if (state < 3 && grepl("Iteration:", line, perl = TRUE)) {
             state <- 3 # 3 =  warmup
             next_state <- 3
+            private$iter_done_[[id]] = 0
```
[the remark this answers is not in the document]
mike-lawrence: Correct, unneeded.

R/run.R
```diff
@@ -531,20 +533,23 @@ CmdStanProcs <- R6::R6Class(
...
       checkmate::assert_integerish(threads_per_proc, lower = 1, len = 1, null.ok = TRUE)
       private$num_procs_ <- as.integer(num_procs)
+      private$iter_ <- iter
       if (is.null(parallel_procs)) {
         private$parallel_procs_ <- private$num_procs_
       } else {
         private$parallel_procs_ <- as.integer(parallel_procs)
       }
-      private$threads_per_proc_ <- as.integer(threads_per_proc)
+      # private$threads_per_proc_ <- as.integer(threads_per_proc) # spandrel?
```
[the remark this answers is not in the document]
mike-lawrence: That line is redundant to the that follows, and I observed that things broke when removing the following line, so commented this over out with intent to bring it up but then forgot. Should be outright deleted if you agree it's unneeded.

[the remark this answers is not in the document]
rok-cesnovar: Yeah, I  agree.

mike-lawrence:
@rok-cesnovar Is the protocol for PRs that I now push a commit that implements the changes you've recommended?

rok-cesnovar:
Typically yes. I way busy today, so have a more comprehensive review of this penciled in for tomorrow. So may as well wait if there will be other comments.

R/model.R
```diff
@@ -1073,6 +1077,7 @@ optimize <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
+    iter = ifelse(is.null(optimize_args$iter),1e3,optimize_args$iter),
```
jgabry, line 1080:
```suggestion
    iter = optimize_args$iter %||% 1e3,
```

jgabry: Also is 1e3 the right value to use for optimize?

R/model.R
```diff
@@ -1197,6 +1202,7 @@ variational <- function(data = NULL,
 
   cmdstan_procs <- CmdStanProcs$new(
     num_procs = 1,
+    iter = ifelse(is.null(variational_args$iter),1e3,variational_args$iter),
```
jgabry, line 1205:
```suggestion
    iter = variational_args$iter %||% 1e3
```

jgabry: Also is 1e3 the right value for variational?

R/run.R
```diff
@@ -555,6 +560,7 @@ CmdStanProcs <- R6::R6Class(
       private$proc_total_time_ = zeros
       private$show_stderr_messages_ = show_stderr_messages
       private$show_stdout_messages_ = show_stdout_messages
+      private$iter_done_ = as.list(rep(0,length(private$proc_ids_)))
```
jgabry, line 563:
```suggestion
      private$iter_done_ <- as.list(rep(0,length(private$proc_ids_)))
```

For code consistency we should use `<-`, although it looks like the few lines above this also use `=` (an oversight, we should fix those too)

R/run.R
```diff
@@ -723,11 +728,34 @@ CmdStanProcs <- R6::R6Class(
...
+        cat( format(id, width=id_width), ': ', sep='' )
+        # set %-compute text
+        if(private$iter_done_[[id]]==0){
+          # waiting-to-start text is ...
+          cat(' ... | ') # same # of chars as '100% '
+        }else{
+          # if any done, show %
+          cat( format(floor(100*private$iter_done_[[id]]/private$iter_),width=3), '% | ', sep='' )
+        }
+      }
```
jgabry, line 747:
```suggestion
      for (id in self$proc_ids()) {
        cat(format(id, width = id_width), ': ', sep = '')
        # set %-compute text
        if (private$iter_done_[[id]] == 0) {
          # waiting-to-start text is ...
          cat(' ... | ') # same # of chars as '100% '
        } else {
          # if any done, show %
          cat(format(floor(100*private$iter_done_[[id]]/private$iter_), width = 3), '% | ', sep = '' )
        }
      }
```

Just spacing changes to follow coding style

R/run.R
```diff
@@ -742,16 +770,23 @@ CmdStanProcs <- R6::R6Class(
           if (self$proc_state(id) == 2.5 && grepl("Exception:", line, fixed = TRUE)) {
             self$set_proc_state(id, new_state = 3)
           }
+          if (grepl("Iteration:", line, perl = TRUE)) {
+            regex_match = regexpr('[0-9]+\\s+/',line)
+            private$iter_done_[[id]] = as.numeric(substr(line,regex_match,regex_match+attr(regex_match,'match.length')-3))
+            ignore_line = TRUE
```
jgabry, line 776:
```suggestion
            regex_match <- regexpr('[0-9]+\\s+/',line)
            private$iter_done_[[id]] <- as.numeric(substr(line, regex_match, regex_match + attr(regex_match, 'match.length') - 3))
            ignore_line <- TRUE
```

Also, the line starting `as.numeric` is a bit long and hard to read. Maybe break it up into two steps, e.g.

```r
iter_done <- substr(line, regex_match, regex_match + attr(regex_match, 'match.length') - 3)
private$iter_done_[[id]] <- as.numeric(iter_done)
```

R/run.R
```diff
@@ -834,9 +871,15 @@ CmdStanMCMCProcs <- R6::R6Class(
           if (state < 3 && grepl("Iteration:", line, perl = TRUE)) {
             state <- 3 # 3 =  warmup
             next_state <- 3
+            private$iter_done_[[id]] = 0
+          }
+          if (grepl("Iteration:", line, perl = TRUE)) {
+            regex_match = regexpr('[0-9]+\\s+/',line)
+            private$iter_done_[[id]] = as.numeric(substr(line,regex_match,regex_match+attr(regex_match,'match.length')-3))
+            ignore_line = TRUE
```
jgabry, line 879:
```suggestion
            regex_match <- regexpr('[0-9]+\\s+/',line)
            private$iter_done_[[id]] <- as.numeric(substr(line, regex_match, regex_match + attr(regex_match, 'match.length') - 3))
            ignore_line <- TRUE
```

Also same comment as above about splitting up the long line

R/run.R
```diff
@@ -723,11 +728,34 @@ CmdStanProcs <- R6::R6Class(
         }
       }
     },
+    print_progress = function(){
+      # get the width of the widest id (sets # leading zeros on all)
+      id_width = nchar(as.character(length(self$proc_ids())))
```
jgabry, line 733:
```suggestion
      id_width <- nchar(as.character(length(self$proc_ids())))
```

jgabry:
Just tried it out and it works really great on Mac! And I really like how lightweight this implementation is.

rok-cesnovar:
I like this a lot! It works great with even a huge number of chains.

A few edge cases to polish:

- Running with one chain does not print the execution time. It finishes with:

```
Running MCMC with 1 chain...

1:  78% | 
```
I would expect 100% regardless of the refresh value and the execution time printed.

- if `refresh = 0`, we should not use the progress
```
Running MCMC with 4 parallel chains...

1:  ... | 2:  ... | 3:  ... | 4:  ... | 
All 4 chains finished successfully.
Mean chain execution time: 0.0 seconds.
Total execution time: 0.2 seconds.
```
- when all chains finish successfully the last print should be all chains at 100%. With
```
fit <- mod$sample(
  data = data_list,  ​seed = 123, ​chains = 4, parallel_chains = 2, ​refresh = 1,
 ​iter_warmup = 1000, ​iter_sampling = 1000
)
```
it finishes as 
```
Running MCMC with 4 chains, at most 2 in parallel...

1: 100% | 2: 100% | 3:  95% | 4:  95% | 
All 4 chains finished successfully.
Mean chain execution time: 1.1 seconds.
Total execution time: 2.4 seconds.
```
- if I run with fixed_param = TRUE, `iter_warmup = 0`, `iter_sampling = 0 ` the error I get is  
```
Error in initialize(...) : 
  Assertion on 'iter' failed: Element 1 is not >= 1.
```
This case does not error on master. I understand it is a silly case, but it should either error with a more informative message (iter is not a user-facin arg so its confusing) or not error.

- if I run with fixed_param = TRUE, iter_sampling = 0 and iter_warmup = 100 I get
```
Running MCMC with 1 chain...

1:  ... | 
```
I am not sure what would we should print here though.

jgabry:
Following up on Rok's comments, I'm noticing that anytime I set `refresh` to any value <= 5 it messes up the percentages.

R/model.R
mike-lawrence: Good catch, it's actually 1e4

mike-lawrence: oops, that's variational. optimize is 2e3

R/model.R
mike-lawrence: good catch, should be 1e4

R/run.R
mike-lawrence: gotcha, will do the as.numeric in a separate commit.

mike-lawrence:
### Simple stuff fixed
* single chain completion time: prior code seems to assume report_time is called in check_finished and therefore has logic to omit the time with a single chain. Fixed this. 

* I also changed the wording of the "... finished successfully" message to simply be a uniform "All chains finished successfully" regardless of chain count (previously had logic to say "Both chains..." with 2 chains, and implicitly had "Chain 1..." in the case of 1 chain). 

* I realized that the warning message when a chain has finished unexpectedly isn't printed now that I've taken `report_time()` out of `check_finished()`, so I put the logic to print that warning in `check_finished()` directly. Arguably, the `report_time()` functions now could have the section relating to the printing of timing for a single chain removed, but I left it in for now just in case there's side effects of that I'm not considering.

### Points for feedback
* @rok I've been running with `refresh=1` a bunch and don't notice any weirdness in the %s; can you elaborate on what you observe?

* Proposal: Always set `refresh=1` internally & replace both the user-facing `refresh` argument and the internal `show_stdout_messages` variable with a user-facing `show_progress` argument that's a binary toggle 

* it occurs to me that by issuing a carriage-return command at the end of the progress print, any messages printed by other functions will overwrite the progress, as I intended, but if they are very short messages, some progress will remain on screen, seemingly concatenated to the short message. A solution to this is to have all messages we let print anywhere be prepended by a newline character (as we do in a few places but not outright consistently). What do you guys think?

rok:
@rok-cesnovar

mike-lawrence:
Oops! Sorry @rok ! I even meant to tag Jonah on that one, but I'll be sure to take care in the future when I do want to tag Rok Češnovar

rok:
No worries @mike-lawrence. I was just making sure @rok-cesnovar got pinged :)

jgabry:
> Oops! Sorry @rok ! I even meant to tag Jonah on that one, but I'll be sure to take care in the future when I do want to tag Rok Češnovar

The more Rok's the merrier!

jgabry:
> * @jgabry  I've been running with `refresh=1` a bunch and don't notice any weirdness in the %s; can you elaborate on what you observe?

Here's what I get with one of the example models with refresh=1 (chain 4 finishes at 98%): 

```
> fit <- cmdstanr_example(quiet = FALSE, refresh = 1)
Compiling Stan program...
Running MCMC with 4 sequential chains...

1: 100% | 2: 100% | 3: 100% | 4:  98% | 
All 4 chains finished successfully.
Mean chain execution time: 1.4 seconds.
Total execution time: 6.3 seconds.
```

> * Proposal: Always set `refresh=1` internally & replace both the user-facing `refresh` argument and the internal `show_stdout_messages` variable with a user-facing `show_progress` argument that's a binary toggle

I think there's a decent bit of overhead with refresh=1. For example, here's running with the default

```
> system.time(fit <- cmdstanr_example(quiet = FALSE, seed = 123))
Model executable is up to date!
Running MCMC with 4 sequential chains...

1: 100% | 2: 100% | 3: 100% | 4: 100% | 
All 4 chains finished successfully.
Mean chain execution time: 0.2 seconds.
Total execution time: 1.0 seconds.

   user  system elapsed 
  0.947   0.234   1.285 
```

and now with refresh = 1

```
> system.time(fit <- cmdstanr_example(quiet = FALSE, refresh = 1, seed = 123))
Model executable is up to date!
Running MCMC with 4 sequential chains...

1: 100% | 2: 100% | 3: 100% | 4:  98% | 
All 4 chains finished successfully.
Mean chain execution time: 1.5 seconds.
Total execution time: 6.1 seconds.

   user  system elapsed 
  2.783   0.338   6.392 
```

> * it occurs to me that by issuing a carriage-return command at the end of the progress print, any messages printed by other functions will overwrite the progress, as I intended, but if they are very short messages, some progress will remain on screen, seemingly concatenated to the short message. A solution to this is to have all messages we let print anywhere be prepended by a newline character (as we do in a few places but not outright consistently). What do you guys think?

mike-lawrence:
> Here's what I get with one of the example models with refresh=1 (chain 4 finishes at 98%):
Ok, fixed that with a final update of the progress info when it's finished

> I think there's a decent bit of overhead with refresh=1. For example, here's running with the default
This overhead should be fairly constant (associated with cmdstan issuing a more frequent fixed-size print statement to stdout) regardless of model size/compute time, so while there's a substantial seeming % slow down for fast models as in your example, it shouldn't be substantially irksome to users. For cases (like sbc) where the user is running many small models, they should be turning off progress for individual runs anyway, so I can't think of any cases where this overhead concern outweighs the benefits of using `refresh=1`.

jgabry:
> > Here's what I get with one of the example models with refresh=1 (chain 4 finishes at 98%):
> 
> Ok, fixed that with a final update of the progress info when it's finished

Great, thanks! I'll test it out soon. 

> This overhead should be fairly constant (associated with cmdstan issuing a more frequent fixed-size print statement to stdout) regardless of model size/compute time, so while there's a substantial seeming % slow down for fast models as in your example, it shouldn't be substantially irksome to users. For cases (like sbc) where the user is running many small models, they should be turning off progress for individual runs anyway, so I can't think of any cases where this overhead concern outweighs the benefits of using refresh=1.

I agree it will be fairly constant, but 6 seconds for a 1 second model (with default settings) will definitely be irksome to some (if not many) users. That said, I do see the benefits of forcing refresh=1 and just offering a show_progress option, so I'm torn on this. @rok-cesnovar What do you think?

jgabry:
Just checked and I can confirm that `refresh=1` now works fine with the example I posted before. Thanks!

rok-cesnovar:
> I agree it will be fairly constant, but I do think 6 seconds for a 1 second model (with default settings) will be irksome to some (if not many) users. That said, I do see the benefits of forcing refresh=1 and just offering a show_progress option, so I'm torn on this. @rok-cesnovar What do you think?

Going from 1 seconds to 6 seconds is too much. My suggestions would be to maybe decrease the default from 100 to 40 and that means a 2% step for the typical 1000/1000 sampling case.

bwiernik:
Any chance this might be finalized at some point?

rok-cesnovar:
@bwiernik This is still our priority feature, I hope to work on this in the near future. 

For the time being, I am going to close this PR as it has gotten stale and has some conflicts to resolve. Feel free to reopen once those are resolved.

Issue #445, opened 2021-02-04
Investigate memory consumption on reading in CSV


rok-cesnovar:
We have trimmed down the reading-in wall time quite a bit in the last few months, now we also need to investigate and profile the memory footprint of both `read_cmdstan_csv` and the `$draws()` function.

It seems that we are currently not near the optimal there, as can be seen in [this report](https://discourse.mc-stan.org/t/cmdstanr-cannot-allocate-vector-of-size-2-0-gb/20592/4?u=rok_cesnovar) by @yizhang-yiz.

The baseline footprint is 
```
fread_cmd <- paste0("grep -v '^#' --color=never ", output_file)
 draws <- data.table::fread(
          cmd = fread_cmd,
          data.table = FALSE
 )
```

yizhang-yiz:
FWIW this is the output with `verbose` on:
```r
> fread_cmd <- paste0("grep -v '^#' --color=never ", "*.csv")
> draws <- data.table::fread(cmd=fread_cmd, data.table=FALSE, verbose=TRUE)
  OpenMP version (_OPENMP)       201511
  omp_get_num_procs()            4
  R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
  R_DATATABLE_NUM_THREADS        unset
  R_DATATABLE_THROTTLE           unset (default 1024)
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          4
  OMP_THREAD_LIMIT               unset
  OMP_NUM_THREADS                unset
  RestoreAfterFork               true
  data.table is using 2 threads with throttle==1024. See ?setDTthreads.
			Input contains no \n. Taking this to be a filename to open
[01] Check arguments
  Using 2 threads (omp_get_max_threads()=4, nth=2)
  NAstrings = [<<NA>>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as integer
[02] Opening the file
  Opening file /tmp/RtmpeZdCIY/file3bed6548c9ef
  File opened, size = 1.945GB (2088953363 bytes).
  Memory mapped ok
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
[05] Skipping initial rows if needed
  Positioned on line 1 starting: <<output.1.csv:lp__,accept_stat_>>
[06] Detect separator, quoting rule, and ncolumns
[...]

PR #479, opened 2021-04-05
Refactor csv.R

rok-cesnovar:
#### Summary

This PR is part 1 of addressing #445. It refactors the csv.R to make the subsequent changes in the next PR easier to integrate. I split this up to 

The ideas behind the refactor: 
- read in all metadata first and check it before reading in samples
- do all the column selection logic outside the loop before reading in samples
- read in sampler diagnostics and the rest in separate reads (makes it more RAM friendly when subsetting later - avoids copies)
- do most of the method-type logic near the return not in the CSV reading loop 

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/R/csv.R
+++ b/R/csv.R
@@ -136,34 +136,38 @@ read_cmdstan_csv <- function(files,
   if (file_idx > 1) {
     check_csv_metadata_matches(csv_metadata)
   }
-  metadata <- csv_metadata[[1]]
-  inv_metric[[as.character(metadata$id)]] <- metadata$inv_metric  
-  step_size[[as.character(metadata$id)]] <- metadata$step_size_adaptation  
-  time <- NULL
-  metadata$id <- NULL
-  metadata$seed <- NULL
-  metadata$init <- NULL
-  metadata$step_size <- NULL
-  metadata$step_size_adaptation <- NULL
-  metadata$fitted_params <- NULL
-  for (file_metadata in csv_metadata) {
-    id <- file_metadata$id
-    metadata$id <- c(metadata$id, id)
-    metadata$seed <- c(metadata$seed, file_metadata$seed)
-    metadata$init <- c(metadata$init, file_metadata$init)
-    metadata$step_size <- c(metadata$step_size, file_metadata$step_size)
-    metadata$step_size_adaptation <- c(metadata$step_size_adaptation, file_metadata$step_size_adaptation)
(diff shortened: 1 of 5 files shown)


jgabry:
Thanks @rok-cesnovar, will start looking at this soon.

R/csv.R
```diff
@@ -297,41 +270,74 @@ read_cmdstan_csv <- function(files,
...
       posterior::variables(warmup_draws) <- repaired_variables
+      if (posterior::niterations(warmup_draws) == 0) {
+        warmup_draws <- NULL
+      }
+    } else {
+      warmup_draws <- NULL
+    }
+    if (length(draws) > 0) {
+      draws <-  posterior::as_draws_array(draws)
+      posterior::variables(draws) <- repaired_variables
```
jgabry, line 289: This line will error if draws are read in for particular variables instead of all variables:

```r
fit <- cmdstanr_example()
fit$draws("alpha")
```

gives this error: 

```
Error in dimnames(x) <- `*vtmp*` : 
  length of 'dimnames' [3] not equal to array extent
Called from: `variables<-.draws_array`(`*tmp*`, value = "alpha")
```

When the error happens `posterior::variables(draws)` has both `"lp__"` and `"alpha"` whereas `repaired_variables` is just `"alpha"`. I guess this is because `"lp__"` has already been read in immediately after fitting? 

This same error occurs a few lines earlier at the analogous line for warmup draws if warmup is included.

jgabry (commented):
> The ideas behind the refactor:
> 
> * read in all metadata first and check it before reading in samples
> * do all the column selection logic outside the loop before reading in samples
> * read in sampler diagnostics and the rest in separate reads (makes it more RAM friendly when subsetting later - avoids copies)
> * do most of the method-type logic near the return not in the CSV reading loop
> 

This sounds great. Will look more at the code soon but for now I tried it out and found that it errors if draws are first read in with only some variables (may be related to the failing tests). My comment above has details.

R/csv.R
rok-cesnovar: Thanks, the issue was the missing , drop = FALSE, which caused names to drop. Should be good now and tests seem to be passing.

jgabry (approved):
Cool, I don't see any issues with the code now and if all tests are passing then hopefully I didn't miss anything. Let's go ahead and merge this.

PR #480, opened 2021-04-06
Fix the use of seed for multi chain sampling

rok-cesnovar:
#### Summary

Right now, cmdstanr sets different seeds for each of the cmdstan process call even if only one seed is given.
```
fit <- mod$sample(..., seed = 2)
```
will set seeds 2,3,4,5 for the 4 chains. This is not ok in the case where transformed uses any _rng functions as
that means that the 4 chains would not have the same data. Even if we set the same seed for all chains, the 
seed used in the rest of the model is actually the seed with a stride defined by the id. See https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/create_rng.hpp#L27
So what we need to do is set the same seeds for all subprocesses.

Test model:
  
```stan
transformed data {
  int N = 100;
  vector[N] a;
  for (i in 1:N) {
    a[i] = uniform_rng(0, 1);
  }
}
parameters {
  real y;
}
model {
  y ~ std_normal();
}
generated quantities {
  vector[N] tdata = a;
  vector[N] gq;
  for (i in 1:N) {
    gq[i] = uniform_rng(0, 4);
  }
}
```

```R
mod <- cmdstan_model("model.stan")
fit <- mod$sample(iter_sampling = 1)
> fit$draws("tdata")
# A draws_array: 1 iterations, 4 chains, and 100 variables
, , variable = tdata[1]

chain
iteration    1    2    3   4
1 0.19 0.62 0.42 0.4

, , variable = tdata[2]

chain
iteration    1    2    3    4
1 0.74 0.67 0.49 0.31

, , variable = tdata[3]

chain
iteration    1    2    3     4
1 0.66 0.95 0.12 0.073

, , variable = tdata[4]

chain
iteration     1    2 3    4
1 0.011 0.82 1 0.45

# ... with 96 more variables

```

This PR fixes this so that these are the same across chains. A test was added to make sure chains actually have a different seed outside of data.

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/R/args.R
+++ b/R/args.R
@@ -837,10 +837,9 @@ validate_seed <- function(seed, num_procs) {
 #' @return An integer vector of length `num_procs`.
 maybe_generate_seed <- function(seed, num_procs) {
   if (is.null(seed)) {
-    seed <- base::sample(.Machine$integer.max, num_procs)
+    seed <- rep(base::sample(.Machine$integer.max, 1), num_procs)
   } else if (length(seed) == 1 && num_procs > 1) {
-    seed <- as.integer(seed)
-    seed <- c(seed, seed + 1:(num_procs -1))
+    seed <- rep(as.integer(seed), num_procs)
   }
   seed
 }
--- a/R/args.R
+++ b/R/args.R
@@ -837,9 +837,9 @@ validate_seed <- function(seed, num_procs) {
 #' @return An integer vector of length `num_procs`.
 maybe_generate_seed <- function(seed, num_procs) {
   if (is.null(seed)) {
-    seed <- rep(base::sample(.Machine$integer.max, 1), num_procs)
+    seed <- base::rep(base::sample(.Machine$integer.max, 1), num_procs)
   } else if (length(seed) == 1 && num_procs > 1) {
-    seed <- rep(as.integer(seed), num_procs)
+    seed <- base::rep(as.integer(seed), num_procs)
   }
   seed
 }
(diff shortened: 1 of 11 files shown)


rok-cesnovar:
If someone wants explicitly different seed they can still set `seed = c(1,2,3,4)`. As far as I understand, this is required for something like SBC.

rok-cesnovar:
This also matches the behavior of rstan and CmdStanPy.

jgabry:
This makes sense. I think we should also update the doc for `seed` to mention what you said above about cmdstan using id to augment the single seed and that it's possible to specify a vector of seeds if the user wants to force different rng results in transformed data.

rok-cesnovar:
Thanks @jgabry, added an attempt at the seed docs. Ready for a look.

jgabry (approved):
I made a few small edits to the `seed` doc, what do you think? If you're ok with the edits then I think this is ready to merge. 

Optionally, one other thing to consider: 

With this change `fit$metadata()$seed` will return a vector that looks like all chains used the same seed: 

```r
> fit <- cmdstanr_example(seed = 123)
> fit$metadata()$seed
[1] 123 123 123 123
```

Unless `seed` is specified as a vector with different elements should we change it so that `fit$metadata()$seed` just gives the single seed instead of the vector of replicated seeds? 

Not really a big deal, but perhaps worth considering.

rok-cesnovar:
The doc changes look great! Thanks for the fixes!

> just gives the single seed instead of the vector of replicated seeds?

Good call. Will fix immediately.

PR #482, opened 2021-04-08
Custom draws_format output

### LINKED ISSUE #445: Investigate memory consumption on reading in CSV

rok-cesnovar:
#### Summary

I am opening this to start a discussion. This is not ready yet, but easier to show with code and a thread to discuss. 
Primarily tagging @jgabry  and @yizhang-yiz for feedback, but anyone feel free to chime in.

The idea is to add a `draws_format` argument to `fit$draws()` and `read_cmdstan_csv()`.

`csv <- read_cmdstan_csv(files, draws_format = "draws_list")`

It could also be set as an option `cmdstanr_draws_format`. The very nice consequence in addition to being more user-friendlier is speed/RAM use, so this would close #445 (or at least cmdstanr side of it).

Example with 4 500MB files:

`f <- read_cmdstan_csv(files, draws_format = "draws_list")` runs for ~20 seconds and the RAM use peaks at ~2.3 GB.
`f <- read_cmdstan_csv(files)` runs for ~45 seconds and the RAM usage peaks at ~15GB

The difference is running `as_draws_array` or `as_draws_list`. The former seems to not be very efficient (~15GB RAM usage for 2GB of data seems excessive). We might be able to make it more efficient but a draws list will always be easier to create from X CSV files.

So we would advise users to use this format when dealing with large number of parameters.

#### Copyright and Licensing

Rok Češnovar


rok-cesnovar:
Can also be used with 
```r
f <- fit$draws(draws_format = "draws_list")
```
or
```r
f <- as_cmdstan_fit(files, draws_format = "draws_list" )
```

jgabry:
Yeah I think adding a `draws_format` argument is a good idea. 

A few minor questions/ideas: 

* Should the argument name be `draws_format` or just `format`? If the name of the method is already `draws` we can probably just use `format`. 
* I think it would be nice if both `"draws_list"` and `"list"` were valid value (with the same behavior). Same for `"draws_array"` and `"array"` and all the others. 
* How would this work with `as_cmdstan_fit()`? Would it change how the draws are stored internally in the resulting fit object?

rok-cesnovar:
Agree with the shorter names and list/array as argument values.

> Would it change how the draws are stored internally in the resulting fit object?

Yes, exactly. We use posterior function to subset/bind the draws, so this should be fine.

jgabry:
Ok cool, that sounds good.

rok-cesnovar:
This is also ready for review.

jgabry:
@rok-cesnovar This looks great. I just pushed a bunch of doc edits. Do they look ok to you? 

Also, do you want to mention in the doc the issue with the speed and memory usage of the different formats?

rok-cesnovar:
Thanks! The doc changes look great. 

> Also, do you want to mention in the doc the issue with the speed and memory usage of the different formats so the use has an idea why the format argument might be useful?

Good call. Added a comment on that.

jgabry (approved):
Ok great, I think this is ready to go now. The tests were passing before and the only recent changes are the doc edits, so let's go ahead and merge this! I'm going to add a NEWS entry and then merge.

rok-cesnovar:
Fantastic!

rok-cesnovar:
I put an item in https://github.com/stan-dev/cmdstanr/issues/200 to maybe make a vignette to showcase this. Might be useful together with the rvar stuff once that is polished.

jgabry:
> I put an item in #200 to maybe make a vignette to showcase this. Might be useful together with the rvar stuff once that is polished.

Ok great, that sounds good.

Issue #483, opened 2021-04-13
Make write_stan_file choose deterministic names (as a caching mechanism)


martinmodrak:
**Is your feature request related to a problem? Please describe.**

When running generated models from `brms`, the `cmdstanr` model is often recompiled even when the Stan code stayed the same. Similarly, when rerunning an analysis that has Stan code in an R variable (in my case usually copied from Stan Discourse), running twice means that the Stan model is always recompiled.

**Describe the solution you'd like**

`write_stan_csv` could hash the Stan code and use it to generate a name that will be the same when the same arguments are passed. It would then check if indeed the file contains the expected Stan code and resolve collisions by either a) deterministically modifying the filename again b) overwrite the file (not sure which is better, but collisions should be rare anyway). This - in combination with the default behaviour where the model is not recompiled when a corresponding binary file exists - would behave as a very simple cache where models don't recompile unless the temp directory is cleared.

This would however potentially introduce issues when multiple processes try to write and compile the same model at the same time. I don't think that's a frequent use case (optionally, there could be an argument to have the tempfile behaviour).

**Describe alternatives you've considered**
Have some sort of in-memory cache of already compiled models.

**Additional context**
If you like it, I am open to (at some point) implement this and file a PR. Any feedback welcome.

rok-cesnovar:
I like this idea! 

We have been thinking (weel @mike-lawrence mostly honestly :)) of using hashing to optimize when compiling models (https://github.com/stan-dev/cmdstanr/issues/423 and https://github.com/stan-dev/cmdstanr/pull/306) though there are much more possible gotchas there compared to this case.

> when rerunning an analysis that has Stan code in an R variable (in my case usually copied from Stan Discourse), running twice means that the Stan model is always recompiled.

+1. I also do this a ton when debugging reported issues or do cmdstanr development and use some toy model and have to comment out the compile stages or something the like. Models in strings are not something I would encourage for serious work but its obviously useful in some cases.

> It would then check if indeed the file contains the expected Stan code and resolve collisions by either a) deterministically modifying the filename again b) overwrite the file (not sure which is better, but collisions should be rare anyway)

I would go for b) here. With sufficient length of hash, chance of collision is negligible.

> This would however potentially introduce issues when multiple processes try to write and compile the same model at the same time. I don't think that's a frequent use case (optionally, there could be an argument to have the tempfile behaviour).

Indeed. Options to circumvent this: 

a) use process ID as hash salt

[...]

PR #494, opened 2021-05-12
Deterministic filenames for `write_stan_file`

martinmodrak:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Makes `write_stan_file` choose filenames deterministically based on the code, so that models do not get unnecessarily recompiled  when calling `write_stan_file` multiple times with the same code, as discussed in #483.

To make this work I needed a hashing algorithm. I chose `rlang::hash`, because we already implicitly depend on `rlang` via `posterior`. The dependency for `rlang` has now been made explicit. I think that's not a big burden, but if needed, the dependency can probably be removed and we can have hashing function implemented locally (it doesn't seem that base R has a hashing function exposed).

#### Copyright and Licensing

Institute of Microbiology of the Czech Academy of Sciences

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,8 @@ Imports:
     jsonlite (>= 1.2.0),
     posterior (>= 0.1.3),
     processx (>= 3.5.0),
-    R6 (>= 2.4.0)
+    R6 (>= 2.4.0),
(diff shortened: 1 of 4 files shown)


martinmodrak:
Bad branch naming, creating a correct version at #495

Issue #205, opened 2020-06-16
Expose diagnostic check functions


rok-cesnovar:
`check_sampler_transitions_treedepth` and `check_divergences` should be accessible to the user in some way. 

Either separately or as part of `fit$diagnose()` or something with possibly other checks.

Cmdstan's diagnostics checks E-BMFI which is simple (only requires `energy__` I think) and two that require reading all data (ESS and RHAT checks).

jgabry:
Yeah I agree we should expose these. I think we should also change the name of `check_sampler_transitions_treedepth` to just `check_treedepth`. 

Maybe a method `fit$diagnose(diagnostic = NULL)`, where if `diagnostic=NULL` it checks all of them but the user can also specify which ones, e.g. `fit$diagnose(c("divergence", "ebfmi"))`?

Also I think it would be useful if these methods returned the value of diagnostic to the user instead of only printing messages (it can also print messages).

rok-cesnovar:
If posterior will have these checks (see https://github.com/stan-dev/posterior/issues/77), we should just use those.

The divergences, treedepth and ebfmi use sampler diagnostics so those are Stan specific and will be done by our helper functions.

cc @paul-buerkner

jgabry:
Yeah I agree

rok-cesnovar:
What should we do here for now: 
- add fit$diagnose() that runs treedepth, divergence and ebmfi without rhat/ess
- add fit$diagnose() that runs treedepth, divergence and ebmfi and also implement warnings for rhat/ess (at least the latter would get replaced by posterior)

This one should be simple.

jgabry:
I think posterior is going to end up including warnings also for HMC/NUTS, so I think all of them could end up being replace by posterior. Do you think we should just wait for posterior or implement them and then replace them with posterior's versions?

rok-cesnovar:
Lets just wait for posterior then and move the milestone if you agree.

jgabry:
Sounds good

rok-cesnovar:
So what I think the API for this should be: 
- add a `diagnostics = c("divergences", "treedepth", "bfmi")` argument to `$sample()`
- deprecate `validate_csv`. I think the name was my idea. In hindsight its a bad name, not informative at all. `validate_csv = FALSE` is the same as `diagnostics = NULL` or `""`
- add `fit$diagnose(diagnostics = c("divergences", "treedepth", "bfmi"))`

[...]

PR #500, opened 2021-05-15
Add check_bfmi function

jsocolar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Added function `check_bfmi`, which takes as input the output of `sampler_diagnostics`, computes the estimated Bayesian fraction of missing information (E-BFMI) for each chain, and prints a message if any chain has E-BFMI less than 0.3.  Note that this uses the threshold of 0.3, consistent with cmdstan's [diagnose.cpp](https://github.com/stan-dev/cmdstan/blob/4f569096ab15676c8050b9cff83de6b17541b609/src/cmdstan/diagnose.cpp) (see line 154), but different from the 0.2 threshold used in [rstan's `check_energy` function](https://github.com/stan-dev/rstan/blob/da2fc9c079534a82d3d26adda51ad17bf22f5e2b/rstan/rstan/R/check_hmc_diagnostics.R) (see line 259).

#### Copyright and Licensing

Jacob B. Socolar


jsocolar:
See also https://discourse.mc-stan.org/t/ebfmi-in-cmdstanr/22509

R/utils.R
```diff
@@ -291,6 +291,20 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
   }
 }
 
+check_bfmi <- function(post_warmup_sampler_diagnostics) {
+  if (!is.null(post_warmup_sampler_diagnostics)) {
+    energy <- posterior::extract_variable_matrix(post_warmup_sampler_diagnostics, "energy__")
+    ebfmi <- apply(energy, 2, function(x) {
+      (sum(diff(x)^2)/length(x))/var(x)
```
jgabry, line 298:
```suggestion
      (sum(diff(x)^2)/length(x))/stats::var(x)
```

That will fix the R cmd check NOTE here: 
https://github.com/stan-dev/cmdstanr/pull/500/checks?check_run_id=2589978139#step:16:63

R/utils.R
```diff
@@ -291,6 +291,20 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+    })
+    if (any(ebfmi < .3)) {
+      message(sum(ebfmi < .3), " of ", length(ebfmi) , " chains had estimated Bayesian fraction
+      of missing information(E-BFMI) less than 0.3, which may indicate poor exploration of the
```
jgabry, line 302: How about a `threshold` argument with a default of 0.3 instead of fixing it at 0.3?

jgabry (changes requested):
Thanks @jsocolar! I made a few small suggestions in review comments, but this looks good. Besides those suggestions we just need a test in https://github.com/stan-dev/cmdstanr/blob/master/tests/testthat/test-utils.R. 

We also need to decide how to use this function (but that doesn't necessarily have to be decided before merging this). Currently this is implemented in utils.R but then not used anywhere so there won't actually be any warnings thrown. One option is to add this check to all the places we do the divergence and treedepth checks (that probably makes sense). We could also combine the divergence, treedepth and ebfmi checks into a `fit$diagnose()` method. Thoughts? Also tagging @rok-cesnovar to get his opinion.

rok-cesnovar:
> One option is to add this check to all the places we do the divergence and treedepth checks (that probably makes sense).

For now, I think we should add it to the two places where other checks are used. In fact I just pushed a commit that does that.

But then we need to do the fit$diagnose() yes, but that should be done once this is in. See my comment in the issue: https://github.com/stan-dev/cmdstanr/issues/205#issuecomment-841694931

rok-cesnovar:
And thanks Jacob for working on this!

jgabry:
@rok-cesnovar Any idea why tests are failing? Currently I don't think anything in this PR should affect the tests that used to be passing. I can look into it if it's mysterious to you too.

R/utils.R
jgabry, line 302:
```suggestion
      message(sum(ebfmi < .3), " of ", length(ebfmi) , " chains had estimated Bayesian fraction
      of missing information (E-BFMI) less than 0.3, which may indicate poor exploration of the 
```

R/utils.R
```diff
@@ -291,6 +291,20 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+check_bfmi <- function(post_warmup_sampler_diagnostics) {
+  if (!is.null(post_warmup_sampler_diagnostics)) {
+    energy <- posterior::extract_variable_matrix(post_warmup_sampler_diagnostics, "energy__")
+    ebfmi <- apply(energy, 2, function(x) {
+      (sum(diff(x)^2)/length(x))/var(x)
+    })
+    if (any(ebfmi < .3)) {
+      message(sum(ebfmi < .3), " of ", length(ebfmi) , " chains had estimated Bayesian fraction
+      of missing information(E-BFMI) less than 0.3, which may indicate poor exploration of the 
+      posterior. Try to reparameterize the model.")
```
jgabry, line 303: Thinking about this a bit more, I think we should remove "Try to reparameterize the model". At least for divergences and max treedepth we have a few concrete recommendations, but I think this is too vague of a suggestion to be helpful to a user. And even for divergences and treedepth I think we should move in the direction of removing the recommended solutions from the warning messages themselves and having the warning messages just be warnings with a pointer to resources that provide recommendations (that way we can also have the recommendations evolve without having to keep changing the warning messages).

R/utils.R
jgabry: Also maybe "energy-based Bayesian fraction of missing information"? Or some other name that includes "energy". It's weird to me to call it E-BFMI but ignore the E when writing out the name (I know it's like that in other places too but I think it's confusing in those places too).

rok-cesnovar:
> Any idea why tests are failing? 

One cause of failures is the edge case of iter_sampling = 1

```
cmdstanr_example("logistic", iter_sampling = 1)
```
I think one issue is var(x) in that case? This is a weird edge case and I am not sure what ebfmi should be in that case.

> Currently I don't think anything in this PR should affect the tests that used to be passing. 

It will run after sampling in all tests.

R/utils.R
jgabry: I'm not sure but maybe we should use the RStan threshold of 0.2. It sounds like the CmdStan threshold is maybe too conservative: 

> We set the E-BFMI warning threshold at 0.3 based on some empirical heuristics but that may have been a bit conservative given how robustly dynamic HMC handles even extremely heavy-tailed distributions. 

from https://discourse.mc-stan.org/t/fixing-bfmi-was-low-problems/658/2. And I'm not sure we have sufficiently good recommendations to justify throwing the warning more often. 

That said, when we add `fit$diagnose()` we can let the user pick the thresholds and be as conservative as they want. Here I'm just thinking about when the warning should automatically happen.

jgabry:
> This is a weird edge case and I am not sure what ebfmi should be in that case.

Good point. I don't think we should compute it for 1 iteration.

R/utils.R
jgabry: Unless the "E" refers to "estimated", but I always thought it referred to "energy"

jsocolar:
Thanks a bunch guys, especially for spotting the iter_sampling = 1 issue (I was mystified about the checks!). I may not have time to take care of this tomorrow (Sunday), but Monday should be doable.

rok-cesnovar:
No rush at all! Thanks!

jgabry:
Just a heads up that I merged in master and fixed a conflict introduced by a different PR that was merged.

R/utils.R
jsocolar: This all makes sense. On the other hand, it could be confusing for the user to get different default behavior from however we implement this diagnostic on the R6 object and `$cmdstan_diagnose()`.  Thoughts @jgabry?

jgabry: Hmm, that's a good point. I would normally lean towards matching CmdStan in that case, however pretty soon nobody will need `cmdstan_diagnose()` anymore (and maybe we should consider deprecating when that's the case) because all of the same checks will be available directly in R (and the R implementations will be preferred since they'll be accessible even after `as_cmdstan_fit()` and even if the CSV files aren't present). So maybe in anticipation of that we should go with the threshold we prefer and not worry about matching cmdstan_diagnose? 

@jsocolar and @rok-cesnovar Do you have strong preferences here?

jsocolar: I definitely don't have preferences. Additional food for thought regarding deprecation of `cmdstan_diagnose()`:
* It's currently the only way to get check the sampler diagnostics and the chain diagnostics (rhat, etc) all at once.
* Is `cmdstan_diagnose()` nontrivially faster than `posterior::summarise_draws` for getting rhat and ess?
* Does it make sense for a wrapper to cmdstan to implement a different check from cmdstan by default?

jsocolar: I always assumed e was for estimated, but p44 of the conceptual intro to HMC paper suggests that it refers to "energy", so I agree with you here.  

TBH I don't understand the first thing about the name of this quantity.  I have no idea why this quantity is a "fraction of missing information", not do I know why we want the fraction of missing information to be large (seems counterintuitive).  Maybe BFMI isn't the fraction of information that is missing, but rather the fraction of missing information that is Bayesian (like due to posterior uncertainty rather than sampler misbehavior)?

jgabry: Yeah those are all good questions. Initial thoughts (may change upon further reflection): 

> * It's currently the only way to get check the sampler diagnostics and the chain diagnostics (rhat, etc) all at once.

Yeah, I don't think we should deprecate it (if we even do deprecate it even) unless we also have the rhat and ess checks in `fit$diagnose()` too. But that should be easy to do. 

> * Is `cmdstan_diagnose()` nontrivially faster than `posterior::summarise_draws` for getting rhat and ess?

Not sure, but that's definitely something to check before deciding to deprecate it. If it's much faster then maybe we don't deprecate it we just mention in the doc that we recommend the R version unless speed is a concern and the CSV files are also available. 

> * Does it make sense for a wrapper to cmdstan to implement a different check from cmdstan by default?

I think it's ok but not ideal. Ideally we'll change it in CmdStan too if the e-bfmi check is too conservative.

jgabry: Yeah I think it's a suboptimal name and it's confusing precisely for the reasons you mentioned. In fact I think I'm going to post on the forum about this and open an issue in the docs repo. I don't think there's anything in the Stan manual about BFMI but there should be an explanation that's clearer than the one in the conceptual intro, which is a great paper, but doesn't clarify this at all.

jgabry: Just posted: https://discourse.mc-stan.org/t/issues-with-e-bfmi-missing-doc-confusing-name-etc/22553. Hope it's ok that I mentioned I was motivated by chatting with you about this stuff.

jsocolar:
This now passes checks (again) locally

jsocolar:
Now feeling silly--ran checks on the check_ebfmi function, but didn't run all tests locally.  The failed tests are because I renamed check_bfmi to check_ebfmi to be more consistent with the output and with @jgabry and my discussions, but forgot that @rok-cesnovar had already included the function elsewhere.  Re-running *all* tests locally, will push once complete.

rok-cesnovar:
Dont feel silly Jacob :) Been there done that countless times haha

No need to wait, just push the changes.

jgabry:
> Dont feel silly Jacob :) Been there done that countless times haha
> 
> No need to wait, just push the changes.

Indeed no worries at all!

jgabry:
Now that there are more changes I'll try to give this another look later today or this weekend.

R/utils.R
```diff
@@ -280,6 +280,44 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+  } else if (dim(pwsd)[1] <= 2) {
+    if (! return_ebfmi) {
+      warning("E-BFMI is undefined for posterior chains of length less than 3")
+    } else {
+      stop("E-BFMI is undefined for posterior chains of length less than 3")
+    }
+  } else {
+    energy <- posterior::extract_variable_matrix(pwsd, "energy__")
+    if (any(is.na(energy))) {
+      if (! return_ebfmi) {
```
jgabry, line 300:
```suggestion
      if (!return_ebfmi) {
```

R/utils.R
```diff
@@ -280,6 +280,44 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+    }
+  } else if (dim(pwsd)[1] <= 2) {
+    if (! return_ebfmi) {
+      warning("E-BFMI is undefined for posterior chains of length less than 3")
+    } else {
+      stop("E-BFMI is undefined for posterior chains of length less than 3")
+    }
+  } else {
+    energy <- posterior::extract_variable_matrix(pwsd, "energy__")
+    if (any(is.na(energy))) {
```
jgabry, line 299:
```suggestion
    if (anyNA(energy)) {
```

jsocolar:
The first three test failures (actually errors outside of testthat) have to do with running `check_ebfmi` on a model with no post-warmup draws.

@rok-cesnovar would you prefer a fix so that `check_ebfmi` doesn't error in this case, or a fix so that `diagnose` doesn't run if there are no iterations to diagnose?

The fourth test failure is mysterious to me as I cannot reproduce locally.

jgabry (changes requested):
I made some minor suggestions, but before addressing those we should probably decide about adding the `return_ebmi` argument. The addition of that argument makes this check function behave differently than the others, which I think we should avoid. I'd also prefer not changing what a function returns based on an argument (that's something it's generally good practice to avoid if possible, although it's less of a concern in internal functions like this). To avoid that we could do one of these options (perhaps there are other options too): 

* have all the check functions always return the diagnostic value (seems like the simplest solution). we don't need to return the value to have the warnings but it doesn't hurt anything and avoids the problem I mentioned
* have a function to compute the diagnostic and another function that inputs the result of that and throws warnings if necessary
 
I think either of those options would be preferable to introducing the `return_ebfmi` argument. @rok-cesnovar @jsocolar what do you think?

R/utils.R
```diff
@@ -280,6 +280,44 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+check_ebfmi <- function(post_warmup_sampler_diagnostics, ebfmi_threshold = .2, return_ebfmi = F) {
+  pwsd <- posterior::as_draws_array(post_warmup_sampler_diagnostics)
+  if (!("energy__" %in% dimnames(pwsd)$variable)) {
+    if (! return_ebfmi) {
+      warning("E-BFMI not computed as the 'energy__' diagnostic could not be located")
+    } else {
+      stop("E-BFMI not computed as the 'energy__' diagnostic could not be located")
+    }
+  } else if (dim(pwsd)[1] <= 2) {
+    if (! return_ebfmi) {
```
jgabry, line 292:
```suggestion
    if (!return_ebfmi) {
```

rok-cesnovar:
> have a function to compute the diagnostic and another function that inputs the result of that and throws warnings if necessary


I like this one.

jgabry:
> would you prefer a fix so that `check_ebfmi` doesn't error in this case, or a fix so that `diagnose` doesn't run if there are no iterations to diagnose?

I think it's fine to just not run any of the diagnostics if there are no iterations (we can't diverge or hit max treedepth either in that case). @rok-cesnovar?

rok-cesnovar:
Agreed. No sampler diagnostics => no checks.

R/utils.R
```diff
@@ -280,6 +280,44 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+        stop("E-BFMI not computed because 'energy__' contains NAs")
+      }
+    }
+    ebfmi <- apply(energy, 2, function(x) {
+      (sum(diff(x)^2) / length(x)) / stats::var(x)
+    }
+    )
+    if (any(ebfmi < ebfmi_threshold)) {
+      message(paste0(sum(ebfmi < ebfmi_threshold), " of ", length(ebfmi), " chains had energy-based Bayesian fraction ",
+      "of missing information (E-BFMI) less than ", ebfmi_threshold, ", which may indicate poor exploration of the ",
```
jgabry, line 312: Since we last discussed this @martinmodrak made a suggestion in 

https://discourse.mc-stan.org/t/issues-with-e-bfmi-missing-doc-confusing-name-etc/22553/17

that we just use the name E-BFMI and don't bother with the full "energy-based Bayesian fraction of missing information", which we all agree is confusing. I'm fine either way, but just wanted to mention that in case we want to change it.

jgabry:
> The fourth test failure is mysterious to me as I cannot reproduce locally.

Agree it's mysterious. I have a theory about it that I'm looking into. Will report back shortly.

R/utils.R
```diff
@@ -280,6 +280,44 @@ check_sampler_transitions_treedepth <- function(post_warmup_sampler_diagnostics,
...
+      }
+    }
+    ebfmi <- apply(energy, 2, function(x) {
+      (sum(diff(x)^2) / length(x)) / stats::var(x)
+    }
+    )
+    if (any(ebfmi < ebfmi_threshold)) {
+      message(paste0(sum(ebfmi < ebfmi_threshold), " of ", length(ebfmi), " chains had energy-based Bayesian fraction ",
+      "of missing information (E-BFMI) less than ", ebfmi_threshold, ", which may indicate poor exploration of the ", 
+      "posterior"))
```
jgabry, line 313:
```suggestion
      "posterior."))
```

jgabry:
> Agree it's mysterious. I have a theory about it that I'm looking into. Will report back shortly.

Still mysterious. I thought perhaps it was due to the recently released R 4.1.0, but I just installed it and still can't reproduce the error locally.

rok-cesnovar:
The tests should be fixed now. The issues were handling the check when no samples were output and a test that had hard coded order of the sampler diagnostics names (the order has now changed because we retrieve `energy__` earlier if we run the checks (previously we only read in `treedepth__` and `divergences__`). Sorry for the force-push, botched the first commit.

rok-cesnovar:
I want to work on the `fit$diagnose()` stuff in the not to distant future and decided to finish the last few things missing here. Also did a bit of a cleanup. @jsocolar hopefully you dont mind.

Changes: 
- pulled out the actual computation. That function will warn if ebfmi can not be computed for any of the reasons (no `energy__` column, less than 3 values, NAs in the column)

- `check_ebfmi` additionally messages if the computed values are below the provided threshold. 

- the `check_ebfmi` and `ebfmi` functions require that the input is in one of the posterior  draws formats (if used with sampler diagnostics it always will be). That way we avoid the convert that was there now. 

This is ready for another look. No rush, I know its vacation season. I can work on the fit$diagnose() on top of this branch anyways now that its close to the finish line.

jgabry:
Sorry for the delay on this. I just pushed a few minor edits and I think this is ready to go. 

But I'm wondering if we should wait to merge it until we have https://github.com/stan-dev/cmdstanr/issues/505 so that we have a resource for users when they see these warnings. At the moment I think only a tiny subset of users will have any idea what to do when they see an E-BFMI warning. Not saying we should definitely wait, just wondering out loud.

jgabry:
Going to leave this open for now, but we probably won't need to actually merge this since all of @jsocolar's commits are included in #585. 

Also, @jsocolar in the branch for PR #585 I'm going to add you to the DESCRIPTION file as a contributor. I realized you're not listed but I think we should list you, both for code contributions (ebfmi , future #565 code, probably other stuff you've done that I don't recall at the moment) and especially for contributing to discussions about the development of the package, which can be just as important as actually contributing code. (If you'd rather not be included let me know and I can undo it.)

Issue #513, opened 2021-06-17
Length-one vectors are incorrectly read as single values


CharnelMouse:
My model takes a few vector inputs, where the length is given as an earlier input.
However, if that length happens to be one, the input is, I think, erroneously taken as a single integer, causing an error.

Simple data block example:

```{r}
data {
  int<lower=0> len;
  int vals[len];
}
```

Error:

`Chain 1 Exception: mismatch in number dimensions declared and found in context; processing stage=data initialization; variable name=vals; dims declared=(1); dims found=() (in 'path', line 3, ...`

This occurs using both JSON and a `stan_rdump`. I'm using the same model on multiple sets of data, where only some have length-one lengths, so I'd rather not do a model rewrite if I can avoid it.

Operating system: Windows 10
Version: 2.27.0

mike-lawrence:
Encountered the same thing myself just recently. Thanks for taking the time to submit as an issue.

jgabry:
Thanks for the bug report. Do you have a small self contained example I can use to trigger the error? Thanks!

CharnelMouse:
Sure!

```{r}
library(cmdstanr)

writeLines(
"data {
  int<lower=0> len;
  int<lower=1,upper=2> vals[len];
}
",
"tmp.stan"
)

len <- 1L
vals <- 2L
rstan::stan_rdump(c("len", "vals"), "tmp.rdump")
dat <- list(
  len = len,
  vals = vals
)
write_stan_json(dat, "tmp.json")

model <- cmdstan_model("tmp.stan")
model$sample(data = "tmp.rdump")
model$sample(data = "tmp.json")
```

jgabry:
Thanks. I think the solution (at least for now) is to use `as.array(vals)`, which will work regardless of the number of elements: 

```r
dat <- list(
  len = len,
  vals = as.array(vals)
)
model$sample(data = dat)  # or using json or rdump
```

Also, totally unrelated to the issue, but you can use the convenience function `write_stan_file()`, which defaults to using a temporary file, and returns the path to the file:

```r
file <- write_stan_file("
data {
  int<lower=0> len;
  int vals[len];
}
")
model <- cmdstan_model(file)
```

What you used is totally fine, just wanted to let you know about `write_stan_file()` in case you hadn't seen it.

CharnelMouse:
Using arrays makes it work for now, thanks!

jgabry:
[...]

PR #519, opened 2021-06-18
Add `$variables()`

rok-cesnovar:
#### Summary

This PR will attempt to add `$variables()`. First we need to settle what will be the args (if any) and what it will return.

Current suggestion is:
```r
library(cmdstanr)

code <- "
data {
  int<lower=0> N;
  int<lower=0> K;
  int<lower=0,upper=1> y[N];
  matrix[N, K] X;
}
parameters {
  real alpha;
  vector[K] beta;
}
model {
  target += normal_lpdf(alpha | 0, 1);
  target += normal_lpdf(beta | 0, 1);
  target += bernoulli_logit_glm_lpmf(y | X, alpha, beta);
}
generated quantities {
  vector[N] log_lik;
  for (n in 1:N) log_lik[n] = bernoulli_logit_lpmf(y[n] | alpha + X[n] * beta);
}
"

mod <- cmdstan_model(write_stan_file(code))
```
```r
mod$variables()
$data
$data$N
$data$N$type
[1] "int"

$data$N$dimensions
[1] 0


$data$K
$data$K$type
[1] "int"

$data$K$dimensions
[1] 0


$data$y
$data$y$type
[1] "int"

$data$y$dimensions
[1] 1


$data$X
$data$X$type
[1] "real"

$data$X$dimensions
[1] 2



$parameters
$parameters$alpha
$parameters$alpha$type
[1] "real"

$parameters$alpha$dimensions
[1] 0


$parameters$beta
$parameters$beta$type
[1] "real"

$parameters$beta$dimensions
[1] 1



$transformed_parameters
NULL

$generated_quantities
$generated_quantities$log_lik
$generated_quantities$log_lik$type
[1] "real"

$generated_quantities$log_lik$dimensions
[1] 1
```

```
> names(mod$variables())
[1] "data"                   "parameters"             "transformed_parameters" "generated_quantities"  
> names(mod$variables()$data)
[1] "N" "K" "y" "X"
```

#### Copyright and Licensing

Rok Češnovar


jgabry:
> First we need to settle what will be the args (if any) and what it will return.

I like your current suggestion. Just a few thoughts: 

* Will the use of the name "dimensions" here be confusing?  For example if we have `vector[100] x` then `fit$metadata()$stan_variable_dims$x` would say `100` but  `variables()` would say "dimensions" is 1. Both make sense, but they're using "dimensions" (or "dims" in the case of metadata()) to mean two different things. 

* I don't think we absolutely need any arguments but we could add an argument `block`. That way we could do `mod$variables("parameters")` instead of `mod$variables()$parameters`. But that doesn't really matter too much. 

And a few questions more related to CmdStan and stanc3 than CmdStanR: 

* Is the "--info" option documented anywhere for stanc3? I couldn't find it so I opened an issue: https://github.com/stan-dev/docs/issues/372

* Is there any way to get more detailed type information from stanc3? That is, if I have a simplex or correlation matrix, is it possible to get it to give me those types instead of just "real"?

rok-cesnovar:
> Will the use of the name "dimensions" here be confusing

could be yes.m, good call. Any other suggestions? Dimension_length? dims_length? Any other?

> we could add an argument block

I like it!

> Is the "--info" option documented anywhere for stanc3?

That is missing yes. Thanks for the issue.

> Is there any way to get more detailed type information from stanc3?

not at the moment, but we can definitely add stuff to what —info returns.

jgabry:
> could be yes.m, good call. Any other suggestions? Dimension_length? dims_length? Any other?
> 

Maybe `n_dims` or `num_dims`? 

Alternatively, we could keep `dimensions` here and instead change `stan_variable_dims` to `stan_variable_sizes` or `stan_variable_elements`?

rok-cesnovar:
This is now ready for review finally. Changes: 
- removed the `block` arg
- added basic docs
- changed stan_variable_dims to stan_variable_sizes

R/model.R
```diff
@@ -1387,4 +1430,23 @@ include_paths_stanc3_args <- function(include_paths = NULL) {
...
+    args = c(stan_file, "--info"),
+    wd = cmdstan_path(),
+    echo = FALSE,
+    echo_cmd = FALSE,
+    stdout = out_file,
+    error_on_status = TRUE
+  )
+  variables <- jsonlite::read_json(out_file, na = "null")
+  variables$data <- variables$inputs
+  variables$inputs <- NULL
```
rok-cesnovar, line 1448: I think `data` is a better name here than `inputs`.

tests/testthat/test-model-variables.R
```diff
@@ -0,0 +1,77 @@
...
+  expect_equal(names(mod$variables()$data), c("N", "y"))
+  expect_equal(names(mod$variables()$parameters), c("theta"))
+  expect_equal(mod$variables()$data$N$type, "int")
+  expect_equal(mod$variables()$data$N$dimensions, 0)
+  expect_equal(mod$variables()$data$y$type, "int")
+  expect_equal(mod$variables()$data$y$dimensions, 1)
+  expect_equal(mod$variables()$parameters$theta$type, "real")
+  expect_equal(mod$variables()$parameters$theta$dimensions, 0)
+  expect_equal(length(mod$variables()[["transformed parameters"]]), 0)
+  expect_equal(length(mod$variables()[["generated quantities"]]), 0)
```
rok-cesnovar, line 19: Should we rename these two to a name with an underscore?

mitzimorris:
regarding names - CmdStanPy and CmdStanR have somewhat diverged on the methods for the CmdStanMCMC object
w/r/t the functions which return the draws for the sampler output variables.

I would love to come up with a good and consistent set of names for these functions on both the CmdStanModel and CmdStanMCMC object.  the difference in dimensionality may be confusing to users - a Stan variable as declared in
the model has N dimensions, in the sampler output, it has N+1 dimensions because is an array of length draws where each
array element has N dimensions.   that's as concisely as I can put it, but is it concise or confusing?

Also, CmdStanMCMC object in CmdStanPy makes a distinction between `sampler_vars` corresponding to `lp__` and friends, and `stan_vars` which are only the model variables which are output by the `write_array` function, i.e., variables in the parameters, transformed parameters, and generated quantities block.   OTOH, from the model `stan_variables` returns data and transformed data variables as well.  Getting the names and dimensions of the data variables is useful.  Getting the names and dimensions of the transformed data variables is not particularly useful.

rok-cesnovar:
>  that's as concisely as I can put it, but is it concise or confusing?

I think its concise.

> Also, CmdStanMCMC object in CmdStanPy makes a distinction between sampler_vars corresponding to lp__ and friends, and stan_vars which are only the model variables which are output by the write_array function, i.e., variables in the parameters, transformed parameters, and generated quantities block.

Yes, CmdStanMCMC does this similarly, just with `$draws()` that returns draws for all the parameters/trans. parameters/GQ but also include `lp__`. The rest of the MCMC diagnostics are returned by sampler_diagnostics. At the time `lp__` was considered to be more of a regular variable then a sampler diagnostic, which I tend to agree.

> Should we make a distinction between data variables and output variables?

I agree that transformed data names are not relevant as they do not occur in the input nor the output, they are not present here.

The purpose of `$variables()` in CmdstanModel is to return the information on all the input or output model variables. So this does not return any values, that would be part of CmdstanMCMC, this is just to return the names, dimenionalities and types of the variables (whether they require integers or reals).

mitzimorris:
> The purpose of $variables() in CmdstanModel is to return the information on all the input or output model variables. So this does not return any values, that would be part of CmdstanMCMC, this is just to return the names, dimenionalities and types of the variables (whether they require integers or reals).

right, this is good and useful.  as it's a method on the CmdStanModel object, the name `model_variables` would be both long and redundant, at the same time, but what about splitting this into `data_vars` and `output_vars` or something like that?

rok-cesnovar:
Sure, no issue in splitting, we discussed this in the issue I think as well.

Options:
a) `$data_variables()`, `$output_variables()`
b) `$input_variables()`, `$output_variables()`
c) `$data_vars()`, `$output_vars()`
d) `$input_vars()`, `$output_vars()`

Not a huge fan of the vars shorthand but seems that is where cmdstanpy is headed, so maybe we should as well. I dont have a strong preference though. @jgabry might have one?

mitzimorris:
I'm happy to go with what whatever consensus we can find.
I have no objection to changing `vars` to `variables` and actually prefer the latter.
similarly, can see arguments for either `data` or `input`.
the inference methods all have argument `data` and the block is named `data`, but inasmuch as everything's data, calling it `input` data also makes sense.

discuss on forums?

jgabry:
> Sure, no issue in splitting, we discussed this in the issue I think as well.
> 
> Options:
> a) `$data_variables()`, `$output_variables()`
> b) `$input_variables()`, `$output_variables()`
> c) `$data_vars()`, `$output_vars()`
> d) `$input_vars()`, `$output_vars()`
> 
> Not a huge fan of the vars shorthand but seems that is where cmdstanpy is headed, so maybe we should as well. I dont have a strong preference though. @jgabry might have one?

Sorry for the delay on this! I totally missed these comments earlier.  

I think I prefer `variables` over `vars`.  And I slightly favor `data` instead of `inputs`.

jgabry:
> I think I prefer `variables` over `vars`. And I slightly favor `data` instead of `inputs`.

But neither preference is super super strong

mitzimorris:
cmdstanpy has `stan_variables` and `method_variables`
the latter is for things like `lp__`
and `stan_variables` are the model variables that are written to the Stan CSV file - i.e.,
parameters, transformed parameters, and generated quantities.

mitzimorris:
are `data_variables` always available?   in CmdStanPy, if you reconstitute an object from a Stan CSV file, you don't have access to the input data.

jgabry:
> are `data_variables` always available? in CmdStanPy, if you reconstitute an object from a Stan CSV file, you don't have access to the input data.

Good point. I think this is true for CmdStanR too

rok-cesnovar:
> are data_variables always available?

`mod$data_variables()` and `mod$output_variables()` will only return the information on all the input and output model variables, not the actual values. So this will return names, dimensionalities and types. Given that its part of the model class, its not possible to return anything else. 

Not sure why availability is the question here? Maybe I am misunderstanding, which is always an option :)

mitzimorris:
OK, now I understand - these are methods on the model class.

I think this is potentially confusing.  it conflates the Stan language with
the current behavior of the Stan platform I/O.
also, given that the model code is available, why are these methods needed?
but if we report Stan program variables, it should be in terms of the Stan
block names - not `input` and `output`.

jgabry:
> Not sure why availability is the question here? Maybe I am misunderstanding, which is always an option :)

No you're right. In my haste I got confused, which tends to happen :)

jgabry:
> given that the model code is available, why are these methods needed?

I'm not sure it's absolutely needed, but it can be helpful. For example, it would allow us to solve this:  https://github.com/stan-dev/cmdstanr/issues/513

mitzimorris:
> I'm not sure it's absolutely needed, but it can be helpful. For example, it would allow us to solve this: #513

if that's the case, then you should add methods. `data_variables`, `parameter_variables`, and `generated_quantities` variables.  and maybe `transformed_data` and `transformed_params` too.

rok-cesnovar:
> how about variables as function name, and arg block allowing folks to examine just the data block, etc.

Yes, this is basically what is currently implemented. I had the block arg initially, but I felt its redundant as one can simply do `mod$variables()$data` or `mod$variables()[["data"]]`.

vignettes/cmdstanr-internals.Rmd
```diff
@@ -163,6 +163,62 @@ mod_pedantic <- cmdstan_model(stan_file_pedantic, compile = FALSE)
...
+names(variables$generated_quantities)
+```
+
+Each variable is represented as a list containing the type
+information (`real` or `int`) and the number of dimensions.
+
+```{r variable-type-dims}
+variables$data$J
+variables$data$sigma
+variables$transformed_parameters$theta
```
jgabry, line 219: @rok-cesnovar When I run this I get `NULL` for the transformed parameters even though it should have `theta`. Do you get that too?

jgabry (changes requested):
Looks good, just the one comment about the missing `transformed parameters`.

vignettes/cmdstanr-internals.Rmd
rok-cesnovar: Yeah, sorry, should have used 
```
variables$`transformed parameters`
variables$`generated quantities`
variables[["generated quantities"]]
```

rok-cesnovar: This was leftover from an earlier version when I replaced the space with underscore. I am not sure which one is better. Space is more similar to the actual block name in the Stan model, underscore is easier to use with `$`. I am fine with either.

jgabry: > underscore is easier to use with `$`. I am fine with either.

I think I prefer the underscore if that's ok with you. Sorry for the extra work!

rok-cesnovar: Not a problem at all. Done.

jgabry (approved):
Thanks! Changes look good. I'll go ahead and merge now.

PR #538, opened 2021-08-22
`process_data()` improvements

### LINKED ISSUE #513: Length-one vectors are incorrectly read as single values

### LINKED ISSUE #533: Error reading matrix with large integral numbers
**Describe the bug**
Stan model fails to run with errors like
```
Chain 1 Exception: msglm_normalize_model_namespace::msglm_normalize_model: qData[sym1__, sym2__] is -1.26487e+09, but must be greater than or equal to 0 (in '/tmp/Rtmpf0rDdB/model-5e8fb26d65883.stan', line 17, column 2 to column 47)
```

The line in question in the Stan model is
```
matrix<lower=0>[Nobjects, Nmschannels] qData;
```

The input data that leads to the failure is [temp.json.zip](https://github.com/stan-dev/cmdstanr/files/6942289/temp.zip) (generated with `write_stan_json())`.
In R the original *qData* matrix has type double.

**To Reproduce**
Define the Stan model with *qData* as specified above, run it with given *qData* from the attachment.

**Expected behavior**
The given qData is correctly read by cmdstan, no range error occurs.

**Operating system**
ArchLinux

**CmdStanR version number**
CmdStanR is 0.4.0.9000, Stan is 2.27

**Additional context**
My current workaround is to modify the *qData* to contain fractional parts (e.g by adding very small number).
It seems in that case the numbers would be parsed as float and no integral overflow will occur.
So one potential fix would be to somehow force the decimal dot in the *.json* output of the double numbers.

There were no problems running the model with *rstan* (Stan 2.19).
Probably because with *rstan* the data is fed to the *Stan* model directly without saving and loading the *.json* file.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Improves process_data in two ways:
 - errors on missing data variables prior to starting the executable
 - handles generating JSON for scalars or arrays/vectors of length 1 differently (fixes #513)
 - fixes #533 by forcing decimal points when outputing floating point values and applying `as.integer()` to R objects in the supplied named list that are supposed to be integers (thanks to @alyst)

#### Copyright and Licensing

Rok Češnovar, Alexey Stukalov


jgabry (approved):
Looks good to me. Thanks!

PR #544, opened 2021-09-02
Add checks for NA in return status

### LINKED ISSUE #543: Model operations fail to handle external process death gracefully
#### Summary:
Under certain circumstances, we encounter a crash at initialization of `brm()`.

#### Description:
In `R/model.R`, the `compile` and `check_syntax` functions use `processx::run()` to launch external processes, and then check the `status` component of the return value against 0 to detect certain failures. Unfortunately, this component gets set to `NA` if the external process is killed, per [https://processx.r-lib.org/reference/run.html](the processx docs), and so the check can generate the error below.

#### Reproducible Steps:
Unfortunately, we have not yet figured out what is causing the external processes to die, and are seeing this in only about 10 percent of runs.

#### Current Output:
When the crash occurs, we see
```
Error in if (run_log$status != 0) { :
  missing value where TRUE/FALSE needed
Calls: brm ... eval -> .fun -> <Anonymous> -> initialize -> <Anonymous>
```

#### Expected Output:
I assume the module should check for NA and report accordingly.

#### Current Version:
Observed in v2.26.1, but the if statement in question remains in the v2.27.0 code.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #543

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@
 * Fixed bug that caused stdour/stderr not being read at the end of
 optimization. (#522)
 
+* Fixed issue with handling `NA` as the reported external process
+status. (#544, @myshkin)
 ### New features
 
 * Default directory changed to `.cmdstan` instead of `.cmdstanr` so that
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ optimization. (#522)
 
 * Fixed issue with handling `NA` as the reported external process
 status. (#544, @myshkin)
+
 ### New features
 
 * Default directory changed to `.cmdstan` instead of `.cmdstanr` so that
--- a/R/model.R
+++ b/R/model.R
@@ -513,7 +513,7 @@ compile <- function(quiet = TRUE,
     },
     error_on_status = FALSE
   )
-  if (run_log$status != 0) {
+  if (is.na(run_log$status) || run_log$status != 0) {
     stop("An error occured during compilation! See the message above for more information.",
          call. = FALSE)
   }
@@ -676,7 +676,7 @@ check_syntax <- function(pedantic = FALSE,
     },
     error_on_status = FALSE
   )
-  if (run_log$status != 0) {
+  if (is.na(run_log$status) || run_log$status != 0) {
     stop("Syntax error found! See the message above for more information.",
          call. = FALSE)
   }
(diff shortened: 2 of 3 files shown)


rok-cesnovar:
This is an easy fix so I think there is no need for reviews, will merge once tests pass. 

Thanks @myshkin and @bachlaw for reporting and suggesting the fix.

jgabry:
> This is an easy fix so I think there is no need for reviews, will merge once tests pass.
> 
> Thanks @myshkin and @bachlaw for reporting and suggesting the fix.

Thanks!

rok-cesnovar:
Tests passed, except for the Mac R-devel tests, but that issue is because the R-devel binary isnt available.

rok-cesnovar:
@myshkin you can install the latest development version now and the fix should be included.

```
remotes::install_github("stan-dev/cmdstanr")
```

myshkin:
Thank you for the quick turnaround! This should help with debugging other issues.

myshkin:
I will note that our output on erroring jobs now looks like
```
Loading required package: cmdstanr
This is cmdstanr version 0.4.0.9000
- Online documentation and vignettes at mc-stan.org/cmdstanr
- CmdStan path set to: /bp/home/bpcompute/.cmdstanr/cmdstan-2.27.0-linux-arm64
- Use set_cmdstan_path() to change the path
The C++ toolchain required for CmdStan is setup properly!
Loading required package: pacman
Error: An error occured during compilation! See the message above for more information.
```
which certainly helps me narrow down where the problem is, but looks misleading, since no relevant message appears above.

rok-cesnovar:
I would advise setting

```
options("cmdstanr_verbose" = TRUE)
```

while debugging stuff like this.

myshkin:
Thanks for the help. I hope to have another bug report for you soon. In the mean time, when I do get the new 
```
Error: An error occured during compilation! See the message above for more information.
```
message, the exit code is still 0. Is that working as expected, or should there be a nonzero exit code to indicate an error?

rok-cesnovar:
> the exit code is still 0

Do you mean the exit code (`$status`) in R or the exit code of the R script you are running?

myshkin:
> Do you mean the exit code ($status) in R or the exit code of the R script you are running?

I meant the exit code of the script; I'm using that to detect whether to rerun the job. Having updated cmdstanr and brms to HEAD of github master, though, I am now getting exit code 1 as expected.

PR #555, opened 2021-09-17
On install, report if the previous installation had a non-empty make/local file

### LINKED ISSUE #455: print reminder to update make/local after install_cmdstan()
I just updated to cmdstan 2.26.1 and very nearly forgot to update make/local. It would be great if users could get reminded about this.  I think the ideal behavior for `install_cmdstan` might be to look for the most recent version of cmdstan (if it exists) in the same directory where the new cmdstan was installed. If there is a previous installation of cmdstan in the same directory, and if make/local exists and contains any non-whitespace characters, print to the console something like:

```
Your previous cmdstan installation ([most recent version number]) may contain persistent compiler flags at [filepath to old make/local]. To continue using these, add them at [filepath to new make/local]. Contents of make/local from version [most recent version number]:
[print(readLines(filepath to old make/local))]
See ?cmdstan_make_local for further information.
```

Users who aren't using make/local don't see any message. Users who do use make/local see a message with instructions for what to do, but still (desirably IMO) have to engage a little bit with what their compiler flags are (rather than just automatically porting across the old make/local, which I think would be potentially undesirable).

rok-cesnovar:
#### Summary

Fixes #455 

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -85,6 +85,14 @@ install_cmdstan <- function(dir = NULL,
   if (check_toolchain) {
     check_cmdstan_toolchain(fix = FALSE, quiet = quiet)
   }
+  make_local_msg <- NULL
+  if (!is.null(cmdstan_version(error_on_NA = FALSE))) {
+    current_make_local_contents <- cmdstan_make_local()
+    if (length(current_make_local_contents) > 0) {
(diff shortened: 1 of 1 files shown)


jgabry (approved):
Looks good, thanks! Merging now.

PR #583, opened 2021-11-02
Fix issue with `stan_threads=FALSE` or `stan_threads=something`

### LINKED ISSUE #577: assert_valid_threads doesn't check value of `cpp_options$stan_threads`
**Describe the bug**
If a model is compiled with `cpp_options = list(stan_threads = X)`, where `X` is *anything at all* (including `FALSE`), and `sample()` is called without setting `threads_per_chain`, then assert_valid_threads will error with `Error: The model was compiled with 'cpp_options = list(stan_threads = TRUE)' but 'threads_per_chain' was not set!`.

**To Reproduce**
file <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")
mod <- cmdstan_model(file, cpp_options = list(stan_threads = FALSE), force_recompile = T)
mod$sample()

Under the hood, what happens is that `assert_valid_threads` is set to error whenever `threads_per_chain` is not set and `is.null(cpp_options[["stan_threads"]]` is `FALSE`.

**What to do**
I'm happy to submit a patch, except that I don't really have a handle on what the behavior should be for values of `cpp_options$stan_threads` that are neither `TRUE` nor `FALSE`.  Depending on the answer to this question, it might be necessary to modify `cmdstan_model` itself in order to implement this behavior.

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #577 

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -1472,7 +1472,7 @@ assert_valid_threads <- function(threads, cpp_options, multiple_chains = FALSE) 
   threads_arg <- if (multiple_chains) "threads_per_chain" else "threads"
   checkmate::assert_integerish(threads, .var.name = threads_arg,
                                null.ok = TRUE, lower = 1, len = 1)
-  if (is.null(cpp_options[["stan_threads"]])) {
+  if (is.null(cpp_options[["stan_threads"]]) || !isTRUE(cpp_options[["stan_threads"]])) {
     if (!is.null(threads)) {
       warning(
         "'", threads_arg, "' is set but the model was not compiled with ",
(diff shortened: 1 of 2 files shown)


jgabry (approved):
Looks good. One question / optional suggestion: what do you think about adding a warning if `stan_threads` is set to some value  that makes no sense (e.g., "dummy string" from the test you added)? If you don't want to add that then it's fine to just go ahead and merge.

R/model.R
```diff
@@ -417,6 +417,10 @@ compile <- function(quiet = TRUE,
   if (length(self$stan_file()) == 0) {
     stop("'$compile()' cannot be used because the 'CmdStanModel' was not created with a Stan file.", call. = FALSE)
   }
+  if (!is.null(cpp_options[["stan_threads"]]) && !is.logical(cpp_options[["stan_threads"]]) ||
+      !is.null(private$precompile_cpp_options_[["stan_threads"]]) && !is.logical(private$precompile_cpp_options_[["stan_threads"]])) {
```
jgabry, line 421: I just realized that this is going to prevent `stan_threads = "true"` or `stan_threads = "TRUE"` and require the user to use `stan_threads = TRUE` only. That's ok with me, but if we want to allow the strings too I guess we'd need to make this check more complicated. What do you think? We can also just skip this warning if you prefer.

rok-cesnovar: Oh right. Lets just skip it then. Will fix a bit later.

tests/testthat/test-threads.R
```diff
@@ -8,159 +8,197 @@ if (not_on_cran()) {
   data_file_json <- test_path("resources", "data", "bernoulli.data.json")
 }
 
-test_that("using threads_per_chain without stan_threads set in compile() warns", {
+# test_that("using threads_per_chain without stan_threads set in compile() warns", {
```
jgabry, line 11: Just adding a reminder for us to turn these tests back on before we merge. I often forget that I've commented out tests.

R/model.R
jgabry: Ok sounds good

Issue #505, opened 2021-05-20
Replace recommendations in warning messages with link to website


jgabry:
@martinmodrak @andrewgelman @avehtari and I are working on the content for a webpage that will have recommendations for how to proceed after seeing Stan's various warning messages (basically it's an improved (and non rstan-specific) version of http://mc-stan.org/misc/warnings). This will allow us to remove the recommendations from the warning messages themselves and instead point to the website. Three of the advantages of this are: 

* we can keep improving the recommendations without having to keep changing the warning messages
* we can put more thorough recommendations than we can fit in warning messages
* multiple interfaces can point to the same webpage

For now there's no action necessary for CmdStanR, but as soon as the website is ready we can address this issue.

jgabry:
@avehtari @martinmodrak @andrewgelman Any updates on this website? Is there anything I can do to help get us closer? I ask because we're about to add E-BFMI warnings to CmdStanR (see #500) to go along with the current checks for divergences and treedepth problems, but I'm mindful of the fact that we still don't have anywhere to direct the users when they see these warnings.

martinmodrak:
My conclusion from the last discussion with @avehtari is that the current draft is satsifactory except for the "What are divergences" appendix that needs a bit more work. So I think we can definitely make the draft into a public page without the appendix almost immediately (it might be useful to ask for additional feedback on Discourse before that).

jgabry:
Ok great, thanks Martin! 

> (it might be useful to ask for additional feedback on Discourse before that).

By "before that", do you mean before making it public or before adding the divergences appendix?

martinmodrak:
I meant before making the document public and start linking to it from error messages.

jgabry:
Ok cool, that makes sense. @avehtari What do you think? Are you ok with soliciting comments about the current draft on discourse at this point?

PR #585, opened 2021-11-04
New method summarizing sampler diagnostics and warnings

### LINKED ISSUE #205: Expose diagnostic check functions

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

* Closes #205
* Builds off of PR #500 from @jsocolar (so merging this one would in effect merge that one too). 

Opening a draft PR to discuss introducing a new method to summarize the sampler diagnostics and regenerate the warning messages. The returned values are vectors of diagnostics per chain (e.g. divergences per chain). 

```r
# note: the warning messages will change once we have the website to point people to
> fit$diagnose_sampler()

Warning: 89 of 4000 (2.0%) transitions ended with a divergence.
This may indicate insufficient exploration of the posterior distribution.
Possible remedies include: 
  * Increasing adapt_delta closer to 1 (default is 0.8) 
  * Reparameterizing the model (e.g. using a non-centered parameterization)
  * Using informative or weakly informative prior distributions 

$num_divergent
[1] 34 24 11 20

$num_max_treedepth
[1] 0 0 0 0

$ebfmi
[1] 0.4277230 0.3856897 0.3388729 0.3541327
```

### Things I'd like feedback on

Tagging some people who have expressed interest in this: @rok-cesnovar @jsocolar @martinmodrak @avehtari 

#### The method name 

Options: 

* `diagnose_sampler()`
* `diagnostic_summary()`
* `check_diagnostics()`
* other suggestions? 

#### Should it include R-hat, ESS?

Does it make sense to include both together here? On the one hand it's nice to have all diagnostics together. On the other hand there are important differences:  

* The HMC/NUTS diagnostics are specific to the Markov chains not the individual parameters. 
* R-hat and ESS are parameter-specific diagnostics. 
* Calculating all of the R-hats and ESS values therefore takes a lot longer than calculating these HMC/NUTS diagnostics. 

Right now they're separate: the HMC/NUTS diagnostics are handled by this method and the R-hat and ESS diagnostics are handled by `fit$summary()`. 

Also note: eventually the posterior package may provide functionality for diagnostics, but this method will be useful until then (or potentially could make use of what posterior provides eventually). 

#### Copyright and Licensing

**Columbia University**


avehtari:
I like `diagnostic_summary()`, then the same name could be used for other inference methods, too.

> Calculating all of the R-hats and ESS values therefore takes a lot longer than calculating these HMC/NUTS diagnostics.

I think this is a sufficient reason to not make it default, Currently we don't know which variables are relevant for computing R-hat and ESS. If the generated quantities computes, e.g. log_lik and many predictions, the number of variables can be very large. It's better to let the user have more control for computing those, e.g., I use `summarise_draws()` with `variable` argument to keep the summary output reasonable length.  

We could check by default Rhat for `lp__`. It's not going to catch all problems, but it would give at least somewhat better chance to recognise well separated multimodality case.

jsocolar:
I think the name depends crucially on whether R-hat and ESS get included here.  I see pros and cons to both options, as you've outlined @jgabry .  I just want to draw out a couple of specific points in the con column.  
1. We don't want to burden the user with doing those computations just to see info about divergences.
2. I imagine we definitely don't want to take info about R-hat and ESS out of the `$summary` output, but we don't want to force the user to compute it twice.
3. Adding enough flexibility to the diagnose method to address the above points would take a very streamlined method and make it a bit unwieldy. I think at a minimum there would be a need for a logical argument for whether to compute R-hat and ESS and a `.cores` argument for computing those in parallel, and perhaps also an argument for which parameters to use.
4. The division between R-hat/ESS on one side and divergences/treedepth/ebfmi on the other is deeper than just a distinction between parameter-level and chain-level diagnostics.  As @jgabry has pointed out elsewhere, it's also a distinction between HMC-specific diagnostics and MCMC-general diagnostics.  As such, it makes sense for R-hat/ESS to live in `posterior`, but for now there is no great way to put the others in `posterior`.  

Overall, my feeling is that these cons outweigh the pros of outputting everything in one place.  If others agree, then my suggestion for a name depends crucially on another thing:  which if any of these diagnostics is applicable to ADVI (or for that matter to potential future plans to return draws as a result of optimizing)?  If none of these diagnostics matters/exists for ADVI, then we should call this method `HMC_diagnostics()` or similar.  If some of these are applicable to ADVI, then I guess `sampler_diagnostics()` or similar might be ok.  Of course if R-hat/ESS get include here then neither of those names works.

jsocolar:
By the way, are all of these diagnostics defined on ADVI? And if not, does this function currently do the right thing in terms of outputting only the relevant diagnostics and not erroring on ADVI fits?

jgabry:
Thanks @avehtari and @jsocolar. 

I agree that the cons outweigh the pros, so we shouldn't include Rhat and ESS here (although @avehtari's idea of maybe just doing it for `lp__` is something to consider). 

> By the way, are all of these diagnostics defined on ADVI? And if not, does this function currently do the right thing in terms of outputting only the relevant diagnostics and not erroring on ADVI fits?

Right now we don't get any information like this after running variational inference. There maybe be ADVI diagnostics in the future but they're likely to be pretty different than these HMC diagnostics.
The way I've coded it this new method is currently only defined for CmdStanMCMC objects and doesn't exist for CmdStanVB objects.

jgabry:
So regarding the method name, @avehtari liked `diagnostic_summary` and @jsocolar liked `hmc_diagnostics` (especially since we don't have these for ADVI and optimization). 

If we use `diagnostic_summary` then it could be defined for all fitted model objects (CmdStanMCMC, CmdStanVB, CmdStanMLE objects) but currently it would only have content for CmdStanMCMC. Later we could add things for other inference methods as they become available.

If we use `hmc_diagnostics` then we would only define this method for CmdStanMCMC objects. If we later want to have diagnostics for ADVI or optimization we would define separate methods (e.g. `fit$advi_diagnostics()`) for CmdStanVB and CmdStanMLE objects. 

I think either of these options could work.

martinmodrak:
> Currently we don't know which variables are relevant for computing R-hat and ESS.

Second this. As an additional problem, for some types of generated quantities, you can have NA Rhats and ESS  even for a model that is well behaved, while NAs would need to be signalled for other types of parameters. The problematic cases are when all the posterior samples are exactly equal - which can easily happen for discrete parameters and is guaranteed to happen for stuff like diagonals of correlation matrices.

jgabry:
Ok based on the feedback so far we won't include R-hat and ESS (I like @avehtari's suggestion to include them for `lp__`, but let's consider adding that as a potential future improvement). 

I think the main decision now is between these two options: 

* A single method `diagnostic_summary()` that does different things for the different fitting methods, (it returns a different named list of diagnostics for MCMC than for ADVI).

* Each class of fitted model object gets its own diagnostic method, e.g., `hmc_diagnostics` for CmdStanMCMC objects, `advi_diagnostics` for `CmdStanVB` objects, `optim_diagnostics` for CmdStanMLE objects (don't necessarily have to use these names). Right now only `hmc_diagnostics` would exist but others could be defined in the future. 

I'm not yet sure which of these I prefer, but I'm hoping to make a decision soon because we'd like to include this in the upcoming v1.0 release of CmdStanR. Does anyone have a strong preference on this or a different option?

jsocolar:
FWIW I do not feel strongly about `diagnostic_summary()` versus `hmc_diagnostics()`.  The only advantage I see in the latter is that it clarifies that the user shouldn't expect to see r-hat and ESS here.  We have the general MCMC diagnostics in posterior, and then the HMC diagnostics in cmdstanr. `hmc_diagnostics()` seems like a good way to get this separation across, but I also see the argument for avoiding a proliferation of different diagnostic methods that must be matched to the fitting method.

jgabry:
> The only advantage I see in the latter is that it clarifies that the user shouldn't expect to see r-hat and ESS here.

That's a really good point and I think it's a pretty strong argument for using a name like `hmc_diagnostics` instead of `diagnostic_summary`. 

> but I also see the argument for avoiding a proliferation of different diagnostic methods that must be matched to the fitting method.

Yeah it's not ideal, but since we only have a small number of different fitting methods it won't be too bad.

jgabry:
> a name like `hmc_diagnostics`

Actually it would be a bit weird to have `hmc_diagnostics` and the existing `sampler_diagnostics` do different things. Maybe `hmc_diagnostic_summary` for the summary/warnings method? If we use that name then I guess that would also suggest renaming `sampler_diagnostics` to `hmc_diagnostics` (`sampler_diagnostics` works better with the `diagnostic_summary` name). If we did want to rename some methods now would be a good time because we'd like to release version 1.0 very soon.

avehtari:
I guess it would be good to ask cmdstanpy people, too? It would be nice to have similarly named methods

rok-cesnovar:
Tagging @mitzimorris @wardbrian @ahartikainen

Though I think cmdstanpy only offers this through the diagnose cmdstan executable right? Or maybe through arviz?

My vote would go to `diagnostic_summary()` FWIW.

ahartikainen:
We should implement these hmc checks into cmdstanpy

jgabry:
> My vote would go to `diagnostic_summary()` FWIW.

I'd be ok with that. That would mean we could keep the name `sampler_diagnostics()` for the method that returns the per iter/chain values and the name `diagnostic_summary` also works well with the existing `summary()` method. The only downside is that it's not obvious that R-hat and ESS wouldn't be included but that's something we can document.

jgabry (commented):
Ok I think this is ready for review. Here is a summary of the changes in this PR: 

#### `sample()` gains new argument `diagnostics` 

The `sample()` method gets a new argument `diagnostics` (replacing `validate_csv`) to control which diagnostics get warnings after sampling. The default is `diagnostics = c("divergences", "treedepth", "ebfmi")`:

```
fit <- cmdstanr_example(
  "schools", 
  max_treedepth = 4, 
  diagnostics = c("divergences", "treedepth", "ebfmi")
)

Warning: 93 of 4000 (2.0%) transitions ended with a divergence.
This may indicate insufficient exploration of the posterior distribution.
Possible remedies include: 
  * Increasing adapt_delta closer to 1 (default is 0.8) 
  * Reparameterizing the model (e.g. using a non-centered parameterization)
  * Using informative or weakly informative prior distributions 

Warning: 2860 of 4000 (72.0%) transitions hit the maximum treedepth limit of 4 or 2^4-1 leapfrog steps.
Trajectories that are prematurely terminated due to this limit will result in slow exploration.
Increasing the max_treedepth limit can avoid this at the expense of more computation.
If increasing max_treedepth does not remove warnings, try to reparameterize the model.
```

(note: these warnings messages will be simplified when we have #505 soon). 

#### new method `diagnostic_summary()`

The new method `diagnostics_summary()` can regenerate these warnings messages and also returns the underlying diagnostic summaries used for the warnings. 

```
fit$diagnostic_summary()
Warning: 93 of 4000 (2.0%) transitions ended with a divergence.
This may indicate insufficient exploration of the posterior distribution.
Possible remedies include: 
  * Increasing adapt_delta closer to 1 (default is 0.8) 
  * Reparameterizing the model (e.g. using a non-centered parameterization)
  * Using informative or weakly informative prior distributions 

Warning: 2860 of 4000 (72.0%) transitions hit the maximum treedepth limit of 4 or 2^4-1 leapfrog steps.
Trajectories that are prematurely terminated due to this limit will result in slow exploration.
Increasing the max_treedepth limit can avoid this at the expense of more computation.
If increasing max_treedepth does not remove warnings, try to reparameterize the model.

$num_divergent
[1] 15  3 37 38

$num_max_treedepth
[1] 784 889 567 620

$ebfmi
[1] 0.3260298 0.2498915 0.4044333 0.2606575
```

It takes two arguments, `diagnostics` (to specify the diagnostics) and `quiet` (to suppress the warnings). For example:

```
fit$diagnostic_summary(diagnostics = "divergences", quiet = TRUE)
$num_divergent
[1] 15  3 37 38
```

#### differences compared to CmdStan's diagnose utility

This is pretty similar to `fit$cmdstan_diagnose()` (which runs `bin/diagnose`), however there are two major differences: 

* `bin/diagnose` diagnoses problems with R-hat and ESS whereas this new method doesn't include those. Those are handled separately by the `summary()` method for reasons discussed in previous comments above. 

* `bin/diagnose` only prints whereas this new method prints the warnings _and_ returns the computed diagnostics summaries

DESCRIPTION
```diff
@@ -12,7 +12,8 @@ Authors@R:
       person(given = "Mikhail", family = "Popov", role = "ctb"), 
       person(given = "Mike", family = "Lawrence", role = "ctb"), 
       person(given = c("William", "Michael"), family = "Landau", role = "ctb",
-           email = "will.landau@gmail.com", comment = c(ORCID = "0000-0003-1878-3253")))
+           email = "will.landau@gmail.com", comment = c(ORCID = "0000-0003-1878-3253")), 
+      person(given = "Jacob", family = "Socolar", role = "ctb"))
```
rok-cesnovar, line 16: Yay! 🎉 More than deserved @jsocolar

rok-cesnovar (approved):
Looks great to me!

jgabry:
> Looks great to me!

Ok awesome. @rok-cesnovar I guess we can go ahead and merge this, but we could also wait a few more days in case anyone else tagged above has any additional feedback. Preference?

rok-cesnovar:
We can definitely wait.

jgabry:
@rok-cesnovar I just fixed the merge conflicts on this PR. I think we've waited long enough now for feedback, so i think we should go ahead and merge this once the checks pass. Are you ok with that?

jgabry:
Actually we finally got the new warnings webpage online so I'm going to take care of #505 as part of this PR too and then we can merge.

jgabry:
Ok @rok-cesnovar (and others), after addressing #505 our new warning messages look like this: 

```
Warning: 95 of 4000 (2.0%) transitions ended with a divergence.
See https://mc-stan.org/misc/warnings for details.

Warning: 3584 of 4000 (90.0%) transitions hit the maximum treedepth limit of 10.
See https://mc-stan.org/misc/warnings for details.

Warning: 2 of 4 chains had an E-BFMI less than 0.2. 
See https://mc-stan.org/misc/warnings for details.
```

If those look ok then I think this is ready to merge (if the tests pass).

rok-cesnovar:
Looks good to me! Feel free to merge.

Issue #589, opened 2021-11-06
Investigate two issues related to unit tests


jgabry:
@rok-cesnovar Opening this just to remind us to investigate the two issues we noticed when merging PR #587. 

_Originally posted by @jgabry in https://github.com/stan-dev/cmdstanr/issues/587#issuecomment-962476914_: 

* After your `Update test-fit-shared.R` commit (which looks good) everything passed except for one of the test coverage runs, which failed with ["Error in `not_on_cran()`: could not find function "not_on_cran"](https://github.com/stan-dev/cmdstanr/runs/4120745980?check_suite_focus=true#step:12:24) (I had removed that function and forgot it was used in that test). I'm not sure how this only affected the test coverage on linux and not the other runs. I think the others should have failed too. It makes sense now that everything is passing after my commit removing `not_on_cran()`, but I'm not sure how they passed before I did that. Do you have any ideas? 

* It looks like we're still using CmdStan 2.27 in the unit tests: https://github.com/stan-dev/cmdstanr/runs/4122519354?check_suite_focus=true#step:17:34, which comes from https://github.com/stan-dev/cmdstanr/blob/21966ac3d60ce6995b6e59f1f5483c2671546fa3/.github/workflows/R-CMD-check.yaml#L34-L38 and https://github.com/stan-dev/cmdstanr/blob/21966ac3d60ce6995b6e59f1f5483c2671546fa3/.github/workflows/Test-coverage.yaml#L20-L23 Do you remember if we hardcoded the CmdStan version for a reason instead of just installing the latest version automatically?

jgabry:
Going to close this since the main issue is resolved now in #592. I'll open a separate issue so we remember to think about whether to hard code the CmdStan version in the GHA workflow

PR #587, opened 2021-11-05
skip all tests on CRAN

jgabry:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and agree to license (see below)

#### Summary

Closes #574. 

CRAN won't have CmdStan installed so were setting `skip_on_cran()` within individual tests. This PR instead just conditions running `test_check()` on the `NOT_CRAN=TRUE` environment variable, which will be set on GHA. 

This shouldn't affect GHA, but to be 100% sure we should probably look at the log of at least one of the runs and make sure the tests are run. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting 
(this will be you or your assignee, such as a university or company): 
**Columbia University**

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)    
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #592, opened 2021-11-09
try to get test to error in order to investigate #589

jgabry:
This PR should _not_ be merged, just using it to trigger the CI.  I'm trying to investigate the question in #589 as to why a test that should have failed in all runs only failed in test coverage on linux. This attempts to recreate the problem.

### DIFF
--- a/tests/testthat/test-install.R
+++ b/tests/testthat/test-install.R
@@ -95,7 +95,7 @@ test_that("install_cmdstan() works with version and release_url", {
   expect_message(
     expect_output(
       install_cmdstan(dir = dir, overwrite = TRUE, cores = 4,
-                      release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.24.0/cmdstan-2.24.0.tar.gz"),
+                      release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.25.0/cmdstan-2.25.0.tar.gz"),
       "Compiling, linking C++ code",
       fixed = TRUE
     ),
@@ -106,9 +106,9 @@ test_that("install_cmdstan() works with version and release_url", {
     expect_message(
       expect_output(
         install_cmdstan(dir = dir, overwrite = TRUE, cores = 4,
-                        version = "2.23.0",
+                        version = "2.26.1",
                         # the URL is intentionally invalid to test that the version has higher priority
-                        release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.23.2/cmdstan-2.23.2.tar.gz"),
+                        release_url = "https://github.com/stan-dev/cmdstan/releases/download/v2.26.3/cmdstan-2.26.3.tar.gz"),
         "Compiling, linking C++ code",
         fixed = TRUE
       ),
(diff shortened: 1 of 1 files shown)


jgabry:
Ok @rok-cesnovar this is pretty strange. I just intentionally inserted a function that doesn't exist into a test and yet all runs except linux test coverage pass. This replicates the issue described in issue #589 that happened in PR #587

tests/testthat/test-install.R
```diff
@@ -26,6 +26,9 @@ test_that("install_cmdstan() successfully installs cmdstan", {
 
 test_that("install_cmdstan() errors if installation already exists", {
   skip_if_offline()
+  if (not_on_cran()) {
```
jgabry, line 29: this function doesn't exist anymore and so should cause an error!

jgabry: the only thing I can think of is maybe `skip_if_offline()` is being triggered? but that doesn't really make any sense if cmdstan was successfully installed from the internet in each of these CI runs. I will try removing `skip_if_offline()` just to see what happens.

jgabry:
Ok very strange. It looks like the test was indeed being skipped because of `skip_if_offline`: once I commented out `skip_if_offline` the test started failing on other runs (as of this comment it's now failing on mac, which it wasn't before. We'll see about the others when they finish). What I don't understand is 

* why does it think it's offline? `skip_if_offline` checks `has_internet <- !is.null(curl::nslookup(host, error = FALSE))` where `host="r-project.org"`. 
* why did this previously only fail on the linux test coverage run?

I think it's probably safe to remove all uses of `skip_if_offline`. At least I'll do that on this branch so we can see if the `install_cmdstan` tests pass (`skip_if_offline` was only being used in those tests).

jgabry:
Without `skip_if_offline` we now have tests failing just on windows in r cmd check but apparently not failing on windows in the separate test coverage run and not failing with the other OSes

jgabry:
@rok-cesnovar No rush, but at some point can you take a look at the windows runs that are now failing. My lack of windows knowledge is preventing me from figuring out what's going on here.

jgabry:
Thanks! I think after my debugging it's pretty clear we need to remove all uses of `skip_if_offline()` (still a mystery as to why that wasn't working properly), but I couldn't quite figure out why that was now causing the windows failures. Thanks for looking into it.

jgabry:
Just noticed this from one of the earlier failed mac runs: 

https://github.com/stan-dev/cmdstanr/runs/4157825884?check_suite_focus=true#step:17:189

It says 

```
══ Skipped tests ═══════════════════════════════════════════════════════════════
• !mpi_toolchain_present() is TRUE (1)
• curl cannot be loaded (5)
```

so I guess it was failing to load curl and that's why `skip_if_offline()` was triggered. Still not sure why it was failing to load curl. Anyway, now that we're removing `skip_if_offline()` this is kind of irrelevant.

rok-cesnovar:
So the issue was that older CmdStan versions (pre-2.26) had an issue with one of the dependency libraries (Boost Program Options) and the GHA setup is done in such a way that building failed. That issue went away with 2.26 so I just bumped the versions used in the tests, so should be good now.

jgabry:
Ok great, thanks for sorting that out. I guess we should we should merge this then. I'll convert it from a draft and then merge.

Issue #590, opened 2021-11-09
Update vignettes and readme for 1.0


jgabry:
One of the last things we should do before the 1.0 release is to update the main vignette, which currently refers to CmdStanR being in beta. 

Also the readme needs to be updated

jgabry:
Also, remember that the opencl vignette doesn't run unless the `CMDSTANR_OPENCL_TESTS` environment variable is set. And when publishing the OPENCL vignette to the website with v1.0 (and eventually to CRAN) we need to decide if we want it to be run in linux or if it's ok if I generate it on mac.

rok-cesnovar:
Yeah, I will run it with my GPU whenever we will do that. Not a problem.

jgabry:
The only issue is that CRAN allows each package to have only a single maintainer at a time and that person has to submit the releases. So I can't be the maintainer but have you generate one of the vignettes (as far as I know, but maybe there's a workaround). So I think that gives us a few alternatives: 

(a) I'm the CRAN maintainer and the vignettes are all generated on my mac
(b) You're the CRAN maintainer and the vignettes are all generated on whatever linux machine you're using 
(c) We don't include the vignettes with the CRAN package, just links to the ones on the CmdStanR website. This allows either of us to be CRAN maintainer and either of us to upload any of the vignettes to the CmdStanR website. 

I don't really care who the maintainer is (I'm happy to do it or not do it), so I guess if we want to make sure the opencl vignette is generated by you then we should do option (b) or (c). I'm ok with either of those options.

rok-cesnovar:
We could also just not include the OpenCL vignette right?

jgabry:
Yeah that's true. We could have the OpenCL vignette only be on the website and not on CRAN. That may be the best option.

jgabry:
I will add this to #595. I think we can put the OpenCL vignette in a subdirectory and have CRAN ignore it but pkgdown put it online.

danielinteractive:
@jgabry Hi Jonah, quick question from the sidelines - do you plan to put `cmdstanr` on CRAN soon?

jgabry:
> do you plan to put `cmdstanr` on CRAN soon?

Yes! 

There are only a few more issues to address before we do a v1.0 release: 

https://github.com/stan-dev/cmdstanr/milestone/2

And really only two of those issues involve much work (#582 and #534).

danielinteractive:
[...]

PR #595, opened 2021-11-10
don't exclude vignettes on GHA (except opencl vignette)

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

advances #590 

* Gets the vignettes running on GHA (except the opencl vignette). Currently in the GHA workflow file we're telling r cmd check to skip building the vignettes, but then we're not alerted to any errors when making changes to the vignettes. So it would be good to have r cmd check on GHA build the vignettes if possible. 

* Moves the OpenCL vignette to a subdirectory so that we can have it online but not on CRAN 

* Fixes a few other r cmd check NOTEs that weren't causing a failure but could be flagged by CRAN

#### Copyright and Licensing

**Columbia University**


jgabry:
This seems successful except for the profiling vignette. For some reason it can't find the stan file in the `profiling-files` subdirectory: 

https://github.com/stan-dev/cmdstanr/runs/4172425847?check_suite_focus=true#step:16:210

Not sure why.

jgabry:
I think the problem is that at the time the vignette code is being evaluated the `profiling-files` directory is no longer available. This is because subdirectories of `vignettes` don't get installed. The reason there's no error when using other subdirectories like `children/settings-knitr.Rmd` in vignettes is that this is used before installation to assemble the final version of the vignette to be installed, whereas we're trying to use `profiling-files/profiling_bernoulli_logit.stan` _after_ installation. 

One solution to this is to get rid of `profiling-files` and use `write_stan_file()` to write the Stan files in the vignette. Testing this now.

jgabry:
> One solution to this is to get rid of `profiling-files` and use `write_stan_file()` to write the Stan files in the vignette. Testing this now

Ok this works.

rok-cesnovar (approved):

rok-cesnovar:
Thanks for cleaning these up!

PR #661, opened 2022-06-13
Fix finding grep.exe path on Windows

rok-cesnovar:
#### Submission Checklist

- [x] Run unit tests

#### Summary

In the edge case where the users do not have `grep.exe` in their system path, we need to wrap the `Sys.which()` call in the with_path call just like we do the rest of the system calls.

#### Copyright and Licensing

Rok Češnovar


rok-cesnovar:
The 4.2 Windows tests are failing for other reasons (see https://github.com/stan-dev/cmdstanr/pull/660) so this should be good to go.

PR #677, opened 2022-07-17
Allow executing cmdstan through WSL

andrjohns:
#### Submission Checklist

- [x] Run unit tests

#### Summary
This PR adds the ability to install `cmdstan` and run all models through WSL, all from a regular Windows R session.

The end-user simply needs to add `wsl=TRUE` to the `install_cmdstan()` call, and all model compilation and execution for that installation will subsequently be run through WSL. No other user-facing changes are made.

In addition to the known speed differences between Stan models under Windows & WSL, this also provides an accessible means for running MPI models in Windows

#### Copyright and Licensing

Andrew Johnson


andrjohns:
Forgot to mention in the description - this PR also adds a Github actions workflow to run the unit tests under WSL as well

rok-cesnovar:
WOW! This is awesome stuff and will be a huge benefit to Windows users! Fantastic!

I am going to review this later today and tomorrow in detail, just a few questions so I know I am understanding this correctly:
- you can execute any command in WSL by running `"wsl something.exe"`, is that correct? Has this been the case always?
- The install location remains the same, but installs the WSL CmdStan's with a `wsl-` prefix?
- WSL is used in a fresh session if the selected CmdStan installation path starts with `wsl-` or `CMDSTANR_USE_WSL`  is set? 

Lastly, please add yourself to the list of contributors.

R/install.R
```diff
@@ -81,7 +83,13 @@ install_cmdstan <- function(dir = NULL,
                             version = NULL,
                             release_url = NULL,
                             cpp_options = list(),
-                            check_toolchain = TRUE) {
+                            check_toolchain = TRUE,
+                            wsl = FALSE) {
```
rok-cesnovar, line 87: Should we give a warning on non-Windows OSes if `wsl = TRUE`?

R/path.R
```diff
@@ -121,6 +121,11 @@ cmdstan_default_install_path <- function(old = FALSE) {
 #' Returns the path to the installation of CmdStan with the most recent release
 #' version.
 #'
+#' For Windows systems with WSL CmdStan installs, if there are side-by-side WSL
+#' and native installs with the same version then the WSL is preferred.
+#' Otherwise, the most recent release is chosen, regardless of whether it is
+#' native or WSL.
```
rok-cesnovar, line 127: 👍🏻

R/utils.R
```diff
@@ -41,6 +41,18 @@ os_is_windows <- function() {
   isTRUE(.Platform$OS.type == "windows")
 }
 
+os_is_wsl <- function() {
+  # Avoid errors if checking cmdstan_path before it has been set
+  wsl_in_path <- tryCatch({
+      grepl("wsl-cmdstan", cmdstan_path())
```
rok-cesnovar, line 47: Should we maybe instead have an internal variable that is set on calls to `set_cmdstan_path()` instead? Just grepping the path is not a big deal so this is optional to avoid the tryCatch.

rok-cesnovar (commented):
I went ahead and just reviewed. Ignore my questions above, I think I got all the answers from the code. The code change looks great and the tests are passing in all OSes. I just want to take a few minutes to test it live on my Windows machine before final approval.

I just have two comments, both are very minor and optional.

R/install.R
rok-cesnovar: Actually, I think this should stop in case of `wsl = TRUE` and the OS not being Windows. I just tried it on my Mac and it started installing in `wsl-cmdstan-2.30.0`.

rok-cesnovar (approved):
Thanks again! Will update NEWS.md and bump the version to 0.5.3.

PR #701, opened 2022-10-06
Add optional methods for log_prob, grad_log_prob, hessian, un/constrain pars

andrjohns:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR adds the (optional) ability to compile and expose several methods from the Stan model so that they can be called by `cmdstanr`. After fitting the model, the user just needs to call the `$init_model_methods()` method, and R will compile and expose the following methods to the fit object:

- `log_prob`
- `grad_log_prob`
- `constrain_pars`
- `unconstrain_pars`

Additionally, if the user calls `$init_model_methods()` with the `hessian = TRUE` flag, a Hessian member function will also be compiled and exposed. I've defaulted this to `FALSE` since it requires all functions in the Stan model to be `fvar<var>` compatible, and I've already run into some issues with `bernoulli_logit_glm` in the basic testing (Math issue/PR to come).

These methods and the dependencies (`Rcpp` and `RcppEigen`) are left as suggests/optional, since I don't think we want `cmdstanr` to require compilation/linking at installation.

#### Copyright and Licensing

Andrew Johnson


rok-cesnovar:
Awesome @andrjohns! Will review this later today.

andrjohns:
@rok-cesnovar The WSL backend seems to be a bit brittle still. I'll be working through the WSL fixes PR soon (after the Math library system libs PR), should we disable the WSL-backend for now until I've got the next PR stable and in?

rok-cesnovar:
I am also fine leaving it as is. Will just ignore the failure on that test for the time being.

andrjohns:
And before I forget, we can also cut down on the compilation time needed here in the future. The only function that needs to be compiled for each model specifically is the `model_ptr` function to initialise the stan model. The other functions can instead be compiled to take a generic `stan::model::model_base` type, which allows them to work with an arbitrary model pointer (the only exception being the `hessian` function, but that's about to be resolved with [this PR](https://github.com/stan-dev/stan/pull/3144)).

The idea would be to compile the `log_prob`/`grad_log_prob`/etc functions once and store the compiled `.dll`/`.so` in the `.cmdstanr` folder, then when the user calls `init_model_methods()` they only compile the `model_ptr` function for the model and the rest are just loaded from the disk

rok-cesnovar:
The changes look great to me, and we can merge as is. Thanks again, this is awesome!

I have a few ideas/thoughts to discuss: 
- I am guessing this works only from a certain version of CmdStan onward, right? I am okay saying this works for 2.30+, no need to explore the exact version. I would just like to add a guard so that it doesn't fail for those versions where it does not work.
- how do you feel about adding an argument to cmdstan_model() that would compile this automatically without having to call the `$init_model_methods()`?

Caching this would be great, but we can handle this once this is in.

andrjohns:
Great!

> I am guessing this works only from a certain version of CmdStan onward, right? I am okay saying this works for 2.30+, no need to explore the exact version. I would just like to add a guard so that it doesn't fail for those versions where it does not work.

Nope, it's not relying on anything introduced recently (as far as I know), just existing headers in the Stan library (with the `#ifdef` to handle the move in the `json` parser from `cmdstar`)

> how do you feel about adding an argument to cmdstan_model() that would compile this automatically without having to call the $init_model_methods()?

Good idea! I can add that tomorrow

rok-cesnovar (approved):
Looks great! Thanks so much!

PR #726, opened 2022-12-02
update pathfinder branch

jgabry:
[Don't merge this yet] 

Updating the pathfinder branch to incorporate latest changes to master. 

@SteveBronder there aren't any conflicts, but before I merge this I want to check with you to make sure this won't harm anything in your `pathfinder_testing` repo. It would be good to keep that up to date with master if possible (e.g., so Chris from BLS and anyone else testing pathfinder can use latest cmdstanr features and bug fixes and also to avoid merge conflicts in the future).

PR #727, opened 2022-12-03
update pathfinder branch 

rok-cesnovar:
@jgabry wrote in #726:

> [Don't merge this yet]
> 
> Updating the pathfinder branch to incorporate latest changes to master.
> 
> @SteveBronder there aren't any conflicts, but before I merge this I want to check with you to make sure this won't harm anything in your pathfinder_testing repo. It would be good to keep that up to date with master if possible (e.g., so Chris from BLS and anyone else testing pathfinder can use latest cmdstanr features and bug fixes and also to avoid merge conflicts in the future).


mitzimorris:
I'm working on https://github.com/stan-dev/cmdstan/issues/1149
which will add single and multi-path Pathfinder to CmdStan.

we should have a discussion about behavior of wrapper interfaces, arg names, etc.   tagging @SteveBronder and @WardBrian to join in.

SteveBronder:
Should this be closed and we can open up a new PR for pathfinder in cmdstanr?

rok-cesnovar:
I think we should merge this, which would merge master into the pathfinder branch and then review, test and get it merged?

mitzimorris:
how does the user call the Pathfinder method?   @WardBrian @jgabry - the usual concern - making the interfaces line up.

jgabry:
Agree with @rok-cesnovar that we should merge this (this PR doesn't merge pathfinder into master, just updates the pathfinder branch). Then we can have a discussion with CmdStanPy devs about what pathfinder should look like in CmdStanR/Py and either start a new pathfinder branch or update this one (if the design is similar enough to what's here already). That make sense to everyone?

jgabry:
Will go ahead and merge this (which just updates the pathfinder branch) and we'll eventually do a new PR (like @SteveBronder mentioned) to get pathfinder into the master branch when the design is agreed upon for all the interfaces.

SteveBronder:
Agree! I wrote it a while ago, and very quickly, so we might be able to save some things from this branch but may also just need to get rid of it.

PR #748, opened 2023-03-24
Allow hessian to fallback to finite-differences

### LINKED ISSUE #728: Add finite-difference based hessian method
For instances where the compilation of the `hessian()` method fails due to `fvar<T>` compatibility issues in the Math library ([as seen here](https://github.com/stan-dev/math/issues/2854)), it would be good to offer an alternative method based on finite-differencing. Possibly also with a warning/message that the finite-diff method will be slower and less precise than the autodiff.

For the interface, I was thinking of introducing this as a separate method, like `hessian_finite()` or similar rather than an option to the existing `hessian()` method. @jgabry any thoughts? I'm not particularly married to any kind of approach here

andrjohns:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR resolves #728 by allowing the `hessian()` method to be implemented using either autodiff or finite-differences, as some models will not compile with the existing autodiff `hessian()` method. This should allow the method to be compatible with all models (as far as I know).

The method will attempt to compile using autodiff and automatically fallback to finite-diff if the compilation fails, or users can request it using the `finite_diff_hessian` argument to `compile()`.

<s>Note that it's not currently possible to suppress/capture compile error output from `sourceCpp()` ([more background here](https://github.com/RcppCore/Rcpp/issues/1257)), so I've implemented a temporary workaround of performing the compilation in a separate R session via the `callr` package. We can remove this (and the package from `Suggests`) once the Rcpp behaviour is updated. </s>

#### Copyright and Licensing

Andrew Johnson


andrjohns:
Rcpp has resolved the issue [upstream](https://github.com/RcppCore/Rcpp/pull/1259), so I've dropped the `callr` workaround and dependency

andrjohns:
I'll also be submitting a PR to drop the additional `decor` dependency soon as well, just as an FYI

rok-cesnovar:
This is good to go, right?

andrjohns:
Yep, good to look at!

andrjohns:
Closing this as will be taking a different approach with hessians

PR #843, opened 2023-09-06
Compatibility fixes for cmdstan 2.33+

jgabry:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and agree to license (see below)

#### Summary

Fixes old array syntax in some examples. 
I've been teaching a Stan workshop and noticed when showing the cmdstanr doc that we still use the old array syntax in a few places. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting
(this will be you or your assignee, such as a university or company):
**Columbia University**

By submitting this pull request, the copyright holder is agreeing to
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #800, opened 2023-07-30
laplace method

jgabry:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and agree to license (see below)

#### Summary

Closes #760 
Builds off of PR #799 

Implements new method `CmdStanModel$laplace()` with new fitted model class `CmdStanLaplace`. 
Follows the design from @WardBrian in https://github.com/stan-dev/cmdstanr/issues/760#issuecomment-1516609693. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting
(this will be you or your assignee, such as a university or company):
**Columbia University**

By submitting this pull request, the copyright holder is agreeing to
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #848, opened 2023-09-11
Pathfinder

SteveBronder:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This implements pathfinder in cmdstanr. Opening as a draft as I'm getting an error on the tests I'm not sure how to fix. Where is the validation for each of the arguments done at? I put them all in the validation function and while the model shows up with errors such as the below they do not actually throw an error.

```
Chain 1 test_bad_val is not a valid value for "max_lbfgs_iters"
Chain 1   Valid values: 0 < max_lbfgs_iters
``` 

I also still need to fix the docs. Once the above is done it should be ready for review

#### Copyright and Licensing

Simons Foundation


jgabry:
Thanks for working on this! 

Regarding the argument validation, it looks like in PathfinderArgs you're using `max_lbfgs_iters` and but in the validation function you have `max_lbfgs_iter` (singular). Could that be the issue? (Although I thought `$` allowed partial matching which should have worked here, but worth checking out)

Also (unrelated), some of the errors in the GHA checks are due to https://github.com/stan-dev/cmdstanr/pull/843. So you'll probably just need to run tests locally until that PR is merged. I think @andrjohns is working on it.

WardBrian:
We should make sure that the names and behavior for the `draws` (`num_psis_draws` in cmdstan) argument match up between this and https://github.com/stan-dev/cmdstanpy/pull/686

In that PR:
arguments are named `draws` and `single_path_draws`. If you don't specifically request `single_path_draws`, it is set equal to `draws`.
If paths > 1, `draws` is cmdstan's `num_psis_draws` and `single_path_draws` is cmdstan's `num_draws`
if paths is exactly 1, both of these are equivalent to cmdstan's `num_draws`

andrjohns:
@SteveBronder I've merged in `master` with the fixes for the failing tests, so now anything that's left should be pathfinder-specific (i.e., git blame Steve)

andrjohns:
@SteveBronder I'm happy to bring this PR over the finish line and sort out any cmdstanr-trickiness, what's left to be done/fixed?

jgabry:
Just a heads up that we just merged #800, which might result in a few conflicts

SteveBronder:
@andrjohns I think this is ready for a first review! I pulled from the sampling method for tests but if you can think of other things I should add I'd be happy to add them

andrjohns:
> @andrjohns I think this is ready for a first review! I pulled from the sampling method for tests but if you can think of other things I should add I'd be happy to add them

Amazing, thanks for this! The current check failures are just from the docs needing to be re-generated and a stray test file being included. I'll fixup those and then review later today. Thanks again!

R/csv.R
```diff
@@ -254,7 +254,7 @@ read_cmdstan_csv <- function(files,
         "\""
       )
     } else {
-      fread_cmd <- paste0("grep -v '^#' --color=never '", output_file, "'")
+      fread_cmd <- paste0("grep -v '^#' --color=never '", path.expand(output_file), "'")
```
andrjohns, line 257: What was this change needed for?

R/model.R
```diff
@@ -1821,6 +1822,124 @@ variational <- function(data = NULL,
...
+                       tol_rel_grad = NULL,
+                       tol_param = NULL,
+                       history_size = NULL,
+                       num_draws = NULL,
+                       num_paths = NULL,
+                       max_lbfgs_iters = NULL,
+                       num_elbo_draws = NULL,
+                       save_single_paths = NULL) {
+  procs <- CmdStanProcs$new(
+    num_procs = 1,
```
andrjohns, line 1896: Is this because multiple chains/threads are handled directly by cmdstan instead?

tests/testthat/testthat-problems.rds
andrjohns, line 1: Can you drop this from the PR as well?

andrjohns (changes requested):
Just a couple of cleanups and some very minor q's, but otherwise this all looks good to me! Thanks for putting it together!

R/model.R
```diff
@@ -1821,6 +1822,124 @@ variational <- function(data = NULL,
...
+                        opencl_ids = NULL,
+                        num_threads = NULL,
+                       init_alpha = NULL,
+                       tol_obj = NULL,
+                       tol_rel_obj = NULL,
+                       tol_grad = NULL,
+                       tol_rel_grad = NULL,
+                       tol_param = NULL,
+                       history_size = NULL,
+                       num_draws = NULL,
```
jgabry, line 1890: We also need to make sure the arguments match with the cmdstanpy implementation. I think currently there are still discrepancies (e.g. `num_draws` was changed in cmdstanpy). See @WardBrian's comment here: https://github.com/stan-dev/cmdstanr/pull/848#issuecomment-1714501090

jgabry (changes requested):
Thanks @SteveBronder! And thanks @andrjohns for reviewing. In addition to Andrew's comments I pointed out in one comment that the arguments need to match the cmdstanpy implementation. We're close but there are a few discrepancies. See comment.

R/csv.R
SteveBronder: I'm not sure I'll revert this change

tests/testthat/testthat-problems.rds
SteveBronder: Sure. Odd, idk where this came from?

R/model.R
SteveBronder: I'll look those over again

R/model.R
SteveBronder: Yep!

SteveBronder:
@WardBrian for 

> arguments are named `draws` and `single_path_draws`. If you don't specifically request `single_path_draws`, it is set equal to `draws`.

Is `single_path_draws` a new argument for the python/R interface or does that overshadow another argument?

WardBrian:
It's a replacement for `num_draws`

SteveBronder:
> If you don't specifically request single_path_draws, it is set equal to draws.

That's pretty expensive, like if I set draws to 10K with 10 pathfinders then I'm going to be taking 100K draws for PSIS which feels like overkill. Why don't we just use the default settings?

R/args.R
```diff
@@ -944,6 +947,10 @@ validate_pathfinder_args <- function(self) {
   if (!is.null(self$num_draws)) {
     self$num_draws <- as.integer(self$num_draws)
   }
+  checkmate::assert_integerish(self$num_psis_draws, lower = 1, null.ok = TRUE, len = 1)
```
jgabry, line 950: I just remembered that there's a `.var.name` argument to the `assert_*` functions that lets you change the name that's displayed to the user in the error message. We could use that here so that the user sees a message about `draws` instead of `num_psis_draws`

WardBrian:
> > If you don't specifically request single_path_draws, it is set equal to draws.
> 
> That's pretty expensive, like if I set draws to 10K with 10 pathfinders then I'm going to be taking 100K draws for PSIS which feels like overkill. Why don't we just use the default settings?

The user can always request fewer, but the reasoning behind this logic is that allowing a user to set `draws` to some larger value without touching `single_path_draws` is a pretty big footgun/does not do what you think it does, because you'll end up resampling the same underlying number of draws. 

I ran the following experiments on slack when we were first discussing this:

> Right now, for just the bernoulli example, if I run pathfinder by default I end up with ~930 unique lines out of 1039 lines in output.csv
If I run pathfinder num_psis_draws=10000, the number of lines in output goes up to 10039, but the number of unique lines is 3684
So I went from having ~90% of my draws be unique to ~35%
if I run pathfinder num_psis_draws=10000 num_draws=10000, then the number of output lines stays 10039, but the number of unique lines is now 8783, back up to ~87%

SteveBronder:
> is a pretty big footgun/does not do what you think it does, because you'll end up resampling the same underlying number of draws.

So that's sort of the idea behind Pareto Smoothed Importance Sampling. Each sample is given a weight based on the ratio of the log likelihood and the approximate log likelihood from a normal approximation. That means some samples are more likely to be drawn more than once which is what we want

WardBrian:
The pathfinder paper, when specifying the multi-pathfinder algorithm, requires R << IM 

(R: num_psis_draws, I: num_paths, M: num_draws)

It doesn’t provide a precise bound besides that, but having the default behavior be that R=M, I > 1 seems like it would do a decent job

SteveBronder:
That's a good catch I'm going to add that to the docs. If we need `draws < (single_path_draws * num_paths)` for the algorithm why don't we just add that as a check during argument validation?

WardBrian:
It's not a bad idea to also have an error thrown if the user requests the something that doesn't meet that. 
The extra logic around `draws` is just to provide a single knob you can turn and get more draws out (it would be annoying if the user requested 10k draws, set no other arguments, and that somehow failed a check -- no other interface has something like that)

SteveBronder:
> (it would be annoying if the user requested 10k draws, set no other arguments, and that somehow failed a check -- no other interface has something like that)

I don't understand why that would be annoying? We could just have a very clear error message like 

```
The number of draws from PSIS for multi pathfinder (draws={draws}) must be greater than the number of total draws from the underlying pathfinder runs (single_path_draws * num_paths = {single_path_draws * num_paths})
```

SteveBronder:
My concern is that users are going to set this to like 10K or 20K and a bunch of pathfinders and get an out of memory error and not understand whatsup. I'd rather we just have an error message and make sure they understand the constraint of the algo

WardBrian:
There are just no other cases to my knowledge where setting one argument to a valid value and not changing any others can fail. Its not the end of the world but it’s also bad UX

If we think that the default ratio of draws to PSIS draws is file, we could set it to (num_psis_draws * 4 / num_paths), which would avoid the issue you’re worried about of someone requesting a lot of paths and a lot of PSIS draws but not setting the single path draws

SteveBronder:
> There are just no other cases to my knowledge where setting one argument to a valid value and not changing any others can fail. Its not the end of the world but it’s also bad UX

idt it's a question of UX but that none of the other algorithms have conditional parameters

> If we think that the default ratio of draws to PSIS draws is fine, we could set it to (num_psis_draws * 4 / num_paths), which would avoid the issue you’re worried about of someone requesting a lot of paths and a lot of PSIS draws but not setting the single path draws

I'm alright with something like this but imo we should change that at the cmdstan level so we don't have different behavior across cmdstan and it's higher level interfaces. tbh even if we have an error check we should probably also do that at the cmdstan level

@LuZhangstat or @bob-carpenter do either of you have thoughts on default behavior for the number of single pathfinder draws when the user sets the number of PSIS draws to be greater than `num_single_path_draws * num_paths`?  (just to recap the paper says we need `num_psis_draws < num_single_path_draws * num_paths`) and we are deciding between a default behavior that will Just Work™️ or throwing an error telling the user about the constraint

bob-carpenter:
> do either of you have thoughts on default behavior for the number of single pathfinder draws when the user sets the number of PSIS draws to be greater than num_single_path_draws * num_paths?

Mathematically, it's consistent to have it be any number, including greater.  If I want to resample 5 items from 1,2,3, then a valid sample is 2, 2, 1, 1, 2.  We tend to get repeats for PSIS unless you choose many fewer importance sample draws than the pool.

At the same time, I'm OK with just capping it and throwing an error.

> There are just no other cases to my knowledge where setting one argument to a valid value and not changing any others can fail.

At what level are you talking about?  If you look at something like CmdStanPy's `sample()` method, there are all kinds of dependencies among the arguments.

Here's CmdStanPy:

```
sample(data=None, chains=None, parallel_chains=None, threads_per_chain=None, seed=None, chain_ids=None, inits=None, iter_warmup=None, iter_sampling=None, save_warmup=False, thin=None, max_treedepth=None, metric=None, step_size=None, adapt_engaged=True, adapt_delta=None, adapt_init_phase=None, adapt_metric_window=None, adapt_step_size=None, fixed_param=False, output_dir=None, sig_figs=None, save_latent_dynamics=False, save_profile=False, show_progress=True, show_console=False, refresh=None, time_fmt='%Y%m%d%H%M%S', timeout=None, *, force_one_process_per_chain=None)
```

If `adapt_engaged=False`, then all the adaptation parameters don't matter.  `max_treedepth` makes no sense if `fixed_param=True`, and so on.

WardBrian:
> At what level are you talking about?

I mean that there is nothing in the sense of "if you set argument A, you must also set argument B to something other than it's default value or else an error is raised"

LuZhangstat:
do either of you have thoughts on default behavior for the number of single pathfinder draws when the user sets the number of PSIS draws to be greater than num_single_path_draws * num_paths?

Mathematically, it's consistent to have it be any number, including greater. If I want to resample 5 items from 1,2,3, then a valid sample is 2, 2, 1, 1, 2. We tend to get repeats for PSIS unless you choose many fewer importance sample draws than the pool.

At the same time, I'm OK with just capping it and throwing an error.

Just as Bob mentioned, "number of PSIS draws greater than num_single_path_draws * num_paths" is not an issue mathematically. Instead of throwing an error. I would recommend showing a warning message when number of PSIS draws is significantly greater than num_single_path_draws * num_paths. Or "num_single_path_draws * num_paths" is too small (say < 20).

Best,

Lu

________________________________

SteveBronder:
I like the idea of throwing a warning when the ratio of psis draws to sum of single Pathfinder draws is low

WardBrian:
I’m not sure if this is also true in cmdstanr, but in cmdstanpy we seem to be going out of our way to not duplicate the default values from cmdstan (we use None rather extensively). I guess the reasoning is it makes the code more robust to if cmdstan’s default values ever changed

In any case, a warning/check that also works in the case where only some of the arguments are specified requires us to have the default values of the current cmdstan present in the code

SteveBronder:
Then I think the warning should go directly in cmdstan so it flows up to the higher level interfaces

jgabry:
> I’m not sure if this is also true in cmdstanr, but in cmdstanpy we seem to be going out of our way to not duplicate the default values from cmdstan (we use None rather extensively)

Yeah we do the same for most of the arguments using `NULL`

SteveBronder:
@andrjohns I think this is ready for another review!

andrjohns:
> @andrjohns I think this is ready for another review!

All looks good to me, was there a consensus on what to do with the `num_single_draws` argument? I can see that it's the only argument in `cmdstanpy` not used here

Once that's settled, happy to approve

WardBrian:
I believe the naming is equivalent to cmdstanpy

andrjohns:
> I believe the naming is equivalent to cmdstanpy

Perfect!

jgabry:
Sorry for the delay! I’ll try to take a look tomorrow

jgabry (changes requested):
Thanks @SteveBronder for implementing this and @andrjohns for reviewing. This looks great, just a few minor things. First, I added a few comments/suggestions about minor doc things that I failed to notice in my last review (sorry). 

The bigger thing I noticed (although maybe I overlooked some discussion about this) is that the number of draws returned seems to be `draws - 1` instead of `draws`: 

```r
file <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")
mod <- cmdstan_model(file)
stan_data <- list(N = 10, y = c(0,1,0,0,0,0,0,0,0,1))
fit <- mod$pathfinder(data = stan_data, draws = 100)
nrow(fit$draws()) # 99 not 100
```

jgabry:
In addition to my previous comments, we should also add something about pathfinder to the vignette, but that doesn't need to be part of this PR. I'll open a separate issue for that.

LuZhangstat:
Thank you for keeping me updated. Please don't forget to add me to the other issue and let me know if I can provide any help.

Best,

lu
________________________________

jgabry:
> Thank you for keeping me updated. Please don't forget to add me to the other issue and let me know if I can provide any help. Best, lu
> […](#)

Hi Lu! Have you tried out the implementation in this PR? No obligation, but let us know if you have any feedback if you test it. Also which other issue were you referencing that you should be added to? Happy to include you in any discussion you're interested in. Or was that a comment directed at @SteveBronder or someone else other than me who already knows which issue?

LuZhangstat:
Hi Jonah,

    Sorry for the delayed response. Do you have an example code for me to test? I installed the latest versions of Stan and cmdstanr, but the Pathfinder function doesn't work on my macbook.

[cid:103c6e88-9885-430a-8fc5-ea0910e6b0c4]
Thanks a lot.

Best,

Lu

________________________________

jgabry:
@LuZhangstat because this hasn't been merged yet you would need to install from this branch: 

```r
# install from feature/pathfinder branch
remotes::install_github("stan-dev/cmdstanr", ref = "feature/pathfinder")

# after that finishes
# simplest example model to test but you can try any model
library("cmdstanr")
file <- file.path(cmdstan_path(), "examples/bernoulli/bernoulli.stan")
mod <- cmdstan_model(file)
stan_data <- list(N = 10, y = c(0,1,0,0,0,0,0,0,0,1))
fit <- mod$pathfinder(data = stan_data) # you can test arguments like num_paths and others
```

Does that work?

jgabry:
Sounds good thanks Lu!

jgabry:
@SteveBronder I have a sec now so I can go ahead and take care of the doc-related comments in my recent review. Can you just see what's up with pathfinder returning `draws-1` draws?

jgabry:
> @SteveBronder I have a sec now so I can go ahead and take care of the doc-related comments in my recent review. Can you just see what's up with pathfinder returning `draws-1` draws?

Ok I just fixed the remaining doc issues so I think the only things left to do before merging this are: 

- Figure out why there are only `draws-1` draws returned (@SteveBronder can you take a look at this?)
- Add Steve to the DESCRIPTION file
- If @LuZhangstat notices any problems address those 

Otherwise I think we're good to go!

SteveBronder:
Just fixed!

SteveBronder:
It was a silly error in csv.R

jgabry:
Thanks Steve!

SteveBronder:
@jgabry good to go?

jgabry (approved):

jgabry:
Good to go! Merging now.

jgabry:
Thanks again for working on this!

Issue #850, opened 2023-09-12
Add "Repository" entry to DESCRIPTION for better compatibility with renv?


andrewheiss:
Currently, [the r-packages repository](https://github.com/stan-dev/r-packages) has a list of instructions for installing non-CRAN Stan packages like {cmdstanr} ([see here for the live version](http://mc-stan.org/r-packages/)).

It includes instructions for adding `https://mc-stan.org/r-packages/` as a named repository for use with {renv}, either by using `options()`, or by manually adding an entry to `renv.lock`.

Since {renv} v1.0 (ish), having a second named repository in `renv.lock` triggers a "The project is out-of-sync" message even if the correct version of packages like {cmdstanr} are installed. [See this issue at {renv} for more details](https://github.com/rstudio/renv/issues/1683), or see this reproducible example (starting from a new empty project):

```r
# Start a new renv library
renv::init()

# Add Stan's repo as a valid source
repos <- c(Stan = "https://mc-stan.org/r-packages/", CRAN = "https://packagemanager.posit.co/cran/latest")
options(repos = repos)

# Install cmdstanr; it will come from the Stan repo; all other dependencies will come from CRAN/RSPM
install.packages("cmdstanr")

# Save to lockfile
renv::snapshot()

# Make sure everything is good
renv::status()
#> No issues found -- the project is in a consistent state.

# RESTART R SESSION

#> Restarting R session...
#> 
#> - Project '~/Desktop/renv-testing' loaded. [renv 1.0.2.9000; sha: dba0079]
#> - The project is out-of-sync -- use `renv::status()` for details.

# It says it's out of sync there ↑
# But if I run renv::status() again, all is well
renv::status()
#> No issues found -- the project is in a consistent state.
```

According to @kevinushey (https://github.com/rstudio/renv/issues/1683#issuecomment-1714530153), this happens because even though `https://mc-stan.org/r-packages/` is a named repository, {renv} checks for a "Repository" entry each package's DESCRIPTION file, and {cmdstanr} doesn't have that in its DESCRIPTION.

It's possible to fix this false positive warning by adding a "Repository" entry to DESCRIPTION:

```
Repository: Stan
```

[...]

PR #851, opened 2023-09-12
Add "Repository" entry to keep {renv} happy

andrewheiss:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add a repository entry to DESCRIPTION so that renv knows that this comes from a non-CRAN repository (addresses https://github.com/stan-dev/cmdstanr/issues/850)

#### Copyright and Licensing

**Andrew Heiss**

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,3 +50,4 @@ Suggests:
     Rcpp,
     RcppEigen
 VignetteBuilder: knitr
+Repository: Stan


jgabry:
Great, thanks for the PR! 

@rok-cesnovar @andrjohns I can't think of any reason why this would cause any issues, and it would resolve a spurious warning from renv, so I think we should do it. But before I merge this, just wanted to check if you know of any reason why we shouldn't do it.

rok-cesnovar:
All for it.

jgabry:
Also, @andrewheiss don't worry about the failing unit tests. Those are presumably due to #843 and should get sorted out when that's merged.

andrewheiss:
Ok good, I won't :) Everything tested fine on my local fork on macOS, but I haven't updated to 2.33 yet

kevinushey:
I don't think you want to add it _directly_ to the source package here; it should only be added when the package is built, if the built package is going to be uploaded to the public repository.

Otherwise, tools that install this package from GitHub (e.g. `remotes::install_github()`) could get confused when installing directly from this repository.

kevinushey:
Perhaps more succinctly: I think the binary packages that are built and published at https://github.com/stan-dev/r-packages should include `Repository: Stan` or something similar, and that should happen in some sort of pre-build step for the packages that are built and published there.

andrewheiss:
Ahh that makes sense. In that case, disregard this PR @jgabry. The `Repository: Stan` line will need to be added somewhere else (maybe here? https://github.com/stan-dev/r-packages/blob/gh-pages/src/contrib/PACKAGES and maybe added to the other packages that are hosted at https://mc-stan.org/r-packages/ like rstan, rstanarm, and StanHeaders - idk how Stan's package building infrastructure works 😕)

Issue #849, opened 2023-09-12
Improve the message in case of deprecated/old syntax + interactive canonicalization


avehtari:
Currently `cmdstanr_model()` outputs deprecation warning or syntax error and recommends running `stanc`, which is a completely foreign thing for R users. Improve the message to recommend using `$format(..., canonicalize = TRUE, overwrite_file = TRUE)`.

Also if the session is run interactively, ask the user whether they would like to do canonicalization right away [Y/n], and then run  `$format(..., canonicalize = TRUE, overwrite_file = TRUE)`

Also provide instructions how to make conversion for a larger number of Stan code files.

jgabry:
I changed the name of this issue now that we have a better message in #852

jgabry:
Closing since we have a better message for users now that doesn't mention `stanc`. I don't have the bandwidth to do interactive canonicalization and we'd have to assess whether we even want something like that. But we can reopen this if someone is interested in taking on that project and wants to discuss.

PR #836, opened 2023-08-24
Improvemements to compile tests

martinmodrak:
#### Submission Checklist

- [ ] Run unit tests
- [x] Declare copyright holder and agree to license (see below)

#### Summary

Improves tests, resolves #835

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting
(this will be you or your assignee, such as a university or company):
Institute of Microbiology of the Czech Academy of Sciences

By submitting this pull request, the copyright holder is agreeing to
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #852, opened 2023-09-13
suggest format method after error due to old syntax 

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Advances #849 (doesn't address the interactive part)

Suggests `format` method if compilation error due to old syntax. This doesn't yet do the formatting if in interactive mode, just suggests looking at the example in the doc for `format`. I added an example to that doc showing how to change old syntax. 

#### Copyright and Licensing

**Columbia University**


jgabry:
### Example

```r
f <- write_stan_file(
"
parameters {
  real x[3];
}
model {
 x ~ normal(0, 1);
}
"
)
m <- cmdstan_model(f)
```
```
Compiling Stan program...
Error in '/var/folders/s0/zfzm55px2nd2v__zlw5xfj2h0000gn/T/RtmpjGrUuu/model-14ede482935d4.stan', line 3, column 2: Declaration
    of arrays by placing brackets after a variable name was removed in Stan
    2.33.0. Instead use the array keyword before the type. This can be
    changed automatically using the auto-format flag to stanc

make: *** [/var/folders/s0/zfzm55px2nd2v__zlw5xfj2h0000gn/T/RtmpjGrUuu/model-14ede482935d4.hpp] Error 65

Error: An error occured during compilation! See the message above for more information. 
To fix deprecated or removed syntax please see ?cmdstanr::format for an example.
```

The last line of the error message is now. But instead of suggesting _how_ to call the format method I have it point to the example in the format doc that I updated. The user needs to set `compile=FALSE` when calling `cmdstan_model` and then run the `format` method, which is a bit much to describe in the error message. Easier to just see the example in the doc. 

We can also do this interactively in the future, but as a start this message should help.

jgabry:
This might conflict with changes in https://github.com/stan-dev/cmdstanr/pull/836, so we should merge that first.

jgabry:
> This might conflict with changes in #836, so we should merge that first.

Just merged that branch into this one, so I can avoid conflicts

jgabry:
@andrjohns @rok-cesnovar @avehtari This is ready for a look if any of you have a chance. We should wait to merge it until #836 is merged (I'll merge that one if the checks that are running pass).

R/model.R
```diff
@@ -929,6 +934,21 @@ CmdStanModel$set("public", name = "check_syntax", value = check_syntax)
...
+#' file <- write_stan_file("
+#' parameters {
+#'   real x[2];
+#' }
+#' model {
+#'   x ~ std_normal();
+#' }
+#' ")
+#' mod <- cmdstan_model(file, compile = FALSE)
+#' mod$format(canonicalize = TRUE)
```
avehtari, line 949: I would strongly prefer the example for fixing old syntax to be
```
mod$format(canonicalize = list("deprecations"))
```
to keep the changes and possible git diffs minimal, when people want to just update the deprecated syntax.

I also suggest showing
```
mod$format(canonicalize = list("deprecations"), overwrite_file=TRUE)
mod$compile()
```

avehtari (commented):
The proposed error message is fine, I'd make changes to the format doc

R/model.R
jgabry: Thanks, sounds good. I'll make the updates.

jgabry:
I updated the doc based on @avehtari's suggestion. Will merge if everything passes.

PR #854, opened 2023-09-21
Update r-markdown.Rmd

jgcolman:
Redraft aimed at clarifying when `override = FALSE` must be used.

#### Submission Checklist

- [ ] Run unit tests
- [ ] Declare copyright holder and agree to license (see below)

#### Summary

Please describe the purpose of the pull request.

#### Copyright and Licensing

**INSERT COPYRIGHT HOLDER HERE**

### DIFF
--- a/vignettes/r-markdown.Rmd
+++ b/vignettes/r-markdown.Rmd
@@ -42,14 +42,16 @@ This is the default option. In that case we can write, for example:
 rstan::sampling(model)
 ```
 ````
-## Option 2: Using  CmdStanR for all chunks
+
+## Option 2: Using CmdStanR for all chunks
+
 If CmdStanR is being used a replacement engine needs to be registered along the following lines:
 
 ```{r register-engine, message=FALSE}
 library(cmdstanr)
 check_cmdstan_toolchain(fix = TRUE, quiet = TRUE)
 
-register_knitr_engine(override = FALSE)
+register_knitr_engine(override = TRUE)
 ```
 
 This overrides knitr's built-in `stan` engine so that all `stan`
@@ -92,6 +94,8 @@ fit <- ex1$sample(
 
 print(fit)
 ```
+
+
 ## Option 3: Using both RStan and CmdStanR in the same R Markdown document
 
 While the default behavior is to override the built-in `stan` engine because the
--- a/vignettes/r-markdown.Rmd
+++ b/vignettes/r-markdown.Rmd
@@ -21,13 +21,17 @@ knitr::opts_chunk$set(
 ```
 
 R Markdown supports a variety of languages through the use of knitr language
-engines. One such engine is the `stan` engine, which allows users to write Stan
-programs directly in their R Markdown documents by setting the language of the
-chunk to `stan`.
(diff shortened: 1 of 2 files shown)


jgabry:
Thanks! I will probably take a look at this next week when I have more time.

vignettes/r-markdown.Rmd
```diff
@@ -38,20 +42,20 @@ the `output.var` chunk option. For example:
...
-CmdStanR provides a replacement engine, which can be registered as follows:
+## Option 2: Using  CmdStanR for all chunks
+If CmdStanR is being used a replacement engine needs to be registered along the following lines:
 
 ```{r register-engine, message=FALSE}
 library(cmdstanr)
 check_cmdstan_toolchain(fix = TRUE, quiet = TRUE)
 
-register_knitr_engine()
+register_knitr_engine(override = FALSE)
```
jgabry, line 52: This should be `override=TRUE` (i.e., override the default of 'stan' chunks using RStan). I will just go ahead and fix this now.

vignettes/r-markdown.Rmd
```diff
@@ -21,13 +21,17 @@ knitr::opts_chunk$set(
...
+engines. Where users wish to write Stan programs as chunks directly in R Markdown documents there are three options: 
+1. the user wishes all the Stan chunks in the R Markdown document to be processed using RStan;  
+2. all Stan chunks are to be processed using CmdStanR; and
+3. some chunks are to be processed by RStan and some by CmdStanR. 
 
-Behind the scenes, the engine relies on RStan to compile the model code into an
-in-memory `stanmodel`, which is assigned to a variable with the name given by
-the `output.var` chunk option. For example:
+Behind the scenes in each option, the engine compiles the model code in each chunk into an
+in-memory model: a `stanmodel` if Rstan is being used, or a `CmdStanModel` in the CmdStanR case. The in-memory model is assigned to a variable with the name given by the `output.var` chunk option.
```
jgabry, line 30: One subtlety is that CmdStanR creates a compiled executable that technically isn't in-memory (from R's perspective). The CmdStanModel object itself is in-memory (so what you say here isn't wrong), but to run Stan it runs the compiled executable outside of R and reads the results back in (unlike RStan which does everything in-memory in R via Rcpp). To avoid any confusion about this maybe we should just remove "in-memory" and instead say something like the following, which I hope gets the idea across: 

> Behind the scenes in each option, the engine compiles the model code in each chunk and creates an object that provides methods to run the model: a `stanmodel` if Rstan is being used, or a `CmdStanModel` in the CmdStanR case. This model object is assigned to a variable with the name given by the `output.var` chunk option.

If you're ok with that change I can make it.

jgabry (commented):
I actually had a few moments so I took a look now. I think this is an improvement, I only noticed two issues. One I will just go ahead and fix and the other relates to the use of the term "in-memory", for which I proposed an alternative that I'd like your feedback on. 

I will also take another look at this next week (I'll be traveling the next few days) to see if I notice anything else.

vignettes/r-markdown.Rmd
jgcolman: I am entirely content with that change.

PR #866, opened 2023-10-17
Add option for installing from release archive

### LINKED ISSUE #865: install_cmdstan() fails because it can't be downloaded
I'm actually pretty frustrated because for some hours now I've tried to install the latest version of install_cmdstan() but it just doesn't work as the download fails. The reason is that where I live (China) github.com is not blocked but the connectivity without VPN is abysmal and most of the time not functioning. 

**The solution I'd like**
A simple solution would be to allow to use already downloaded .tar.gz archives. I thought I'd be smart and just download the archive from [here](https://github.com/stan-dev/cmdstan/releases/download/v2.33.1/cmdstan-2.33.1.tar.gz) myself and then use the path with the argument release_url but unfortunately this doesn't work. 

However, this would also be super helpful in cases where one can't use any VPNs like high-performance clusters of universities but also for me now. This would make cmdstan(r) much more accessible in my opinion. 

PS So far I haven't found out how to install cmdstan without this function. Any guidance would be highly appreciated.

andrjohns:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Closes #865 by adding the option to install cmdstan from a downloaded release archive:
```r
install_cmdstan(release_file = "./cmdstan-2.33.1.tar.gz")
```

#### Copyright and Licensing

Andrew Johnson

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -51,6 +51,10 @@
 #'   The URL should point to the tarball (`.tar.gz.` file) itself, e.g.,
 #'   `release_url="https://github.com/stan-dev/cmdstan/releases/download/v2.25.0/cmdstan-2.25.0.tar.gz"`.
 #'   If both `version` and `release_url` are specified then `version` will be used.
+#' @param release_file (string) A file path to a CmdStan release tar.gz file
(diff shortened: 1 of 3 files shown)


jgabry (approved):
Thanks!

PR #881, opened 2023-11-21
fix: #770, get_cmdstan_flags('STANCFLAGS') in recursive make extraneo…

### LINKED ISSUE #770: get_stan_cflags behaves poorly under recursive `make`
(This seems to belong here, but arguably it could belong on the `brms` issues list. I'd be happy to post it there instead if advised.)

As noted [here](https://myprogrammingnotes.com/entering-directory-leaving-directory-messages-make-output.html), when `make` is called recursively it automatically prints

```
make[N]: Entering directory ...
make[N]: Leaving directory ...
```
messages (where `N` is the stack level).

According to [the documentation](https://www.gnu.org/software/automake/manual/html_node/Tricks-For-Silencing-Make.html),  these messages can be silenced by calling `make -s`.   [This commit](https://github.com/bbolker/cmdstanr/commit/5e77e1988f07f0911e4a3d8840a2b76fb5c8b1e1) makes that change; I'd be happy to submit a PR (I can't see how it would break anything else - the `-s` flag is supposed to be POSIX-portable - but you never know ...)

More details below as to what these messages break and why ...

----

If I run 

```
Rscript -e "sink('test2.Rout'); cat(cmdstanr:::get_cmdstan_flags('STANCFLAGS'), '\n')"
```

in the shell I get an empty `STANCFLAGS` (i.e., `test2.Rout` has only a newline in it).

If I write a Makefile with the rule

```
test.Rout:
	Rscript -e "sink('test2.Rout'); cat(cmdstanr:::get_cmdstan_flags('STANCFLAGS'), '\n')"
```

then `test2.Rout` contains this:

```
make[1]: Entering directory '/home/bolker/.cmdstan/cmdstan-2.32.2'STANCFLAGS =make[1]: Leaving directory '/home/bolker/.cmdstan/cmdstan-2.32.2
```

Downstream, this leads to errors of the following form when I try to run `brms` (with the `cmdstanr` backend, of course).

```
Error in `if (stancflags_local != "") ...`:
! the condition has length > 1
Backtrace:
  1. global test_prior(b_prior)
  4. brms::brm(...)
  5. brms::do_call(compile_model, compile_args)
  6. brms:::eval2(call, envir = args, enclos = envir)
  7. base::eval(expr, envir, ...)
     ...
 17. base::eval(expr, envir, ...)
 18. cmdstanr (local) .fun(stan_file = .x1)
 19. CmdStanModel$new(...)
 20. cmdstanr (local) initialize(...)
 21. self$compile(...)
                                                                                                         
Execution halted
```

pearsonca:
Fix #770 by invoking `make` with `-s` argument.

#### Submission Checklist

- [X] Run unit tests

#### Summary

This PR addresses issue #770. The problem highlighted in that issue is that recursive calls to `make` can generate messages on stdout. Since `cmdstanr` parses `make`-generated stdout to ascertain certain configuration info (e.g. in `get_cmdstan_flags()`), this poses a problem. Since `make` is also a useful tool for research work generally, recursive calls are a practical consideration.

The change here has `get_cmdstan_flags()` invoke `make` in silent mode (`-s`), which means only the explicitly requested print-information is returned. I initially proposed `--no-print-directory` as the solution, but that is apparently a gnumake-specific flag.

N.b. this is the same fix @bbolker proposed (which I saw when looking for an existing issue after I'd fixed this locally, so ... yeah, should have read first then coded).

~~This should probably come with a unit test, but I'm uncertain on the best formulation of such test.~~

I have added a unit test which:
 - creates a temporary makefile with a target that sends `get_cmdstan_flags()` output to stdout
 - invokes that makefile (using cmdstanr infrastructure) ,
 - captures the output and compares it to the output of invoking `get_cmdstan_flags()` directly.
 
 I confirmed (on Ubuntu 22 only) this fails against 357a07, but passes with this modification.

#### Copyright and Licensing

**London School of Hygiene & Tropical Medicine**

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -250,14 +250,11 @@ test_that("as_mcmc.list() works", {
 })
 
 test_that("get_cmdstan_flags() can be used recursively in `make`", {
-  tmp <- withr::local_tempfile()
-  writeLines(
-    "test:\n\t@Rscript --vanilla -e 'cat(cmdstanr:::get_cmdstan_flags(\"STANCFLAGS\"))'",
-    tmp
-  )
+  mkfile <- normalizePath(test_path("testdata", "Makefile"))
   nonrecursive_flags <- get_cmdstan_flags("STANCFLAGS")
-  recursive_flags <- readLines(textConnection(wsl_compatible_run(
-    command = "make", args = c("-f", tmp), wd = cmdstan_path()
-  )$stdout))
+  stdo <- wsl_compatible_run(
+    command = "make", args = sprintf("--file=%s", mkfile), wd = cmdstan_path()
+  )$stdout
+  recursive_flags <- readLines(textConnection(stdo))
   expect_equal(nonrecursive_flags, recursive_flags)
 })
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -257,7 +257,7 @@ test_that("get_cmdstan_flags() can be used recursively in `make`", {
   )
   nonrecursive_flags <- get_cmdstan_flags("STANCFLAGS")
   recursive_flags <- readLines(textConnection(wsl_compatible_run(
-    command = "make", args = c("-f", tmp)
+    command = "make", args = c("-f", tmp), wd = cmdstan_path()
   )$stdout))
   expect_equal(nonrecursive_flags, recursive_flags)
 })
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -248,3 +248,16 @@ test_that("as_mcmc.list() works", {
     "Currently only CmdStanMCMC objects can be converted to mcmc.list"
   )
 })
+
+test_that("get_cmdstan_flags() can be used recursively in `make`", {
+  tmp <- withr::local_tempfile()
+  writeLines(
+    "test:\n\t@Rscript --vanilla -e 'cat(cmdstanr:::get_cmdstan_flags(\"STANCFLAGS\"))'",
+    tmp
+  )
+  nonrecursive_flags <- get_cmdstan_flags("STANCFLAGS")
+  recursive_flags <- readLines(textConnection(wsl_compatible_run(
+    command = "make", args = c("-f", tmp)
+  )$stdout))
+  expect_equal(nonrecursive_flags, recursive_flags)
+})
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -252,8 +252,8 @@ test_that("as_mcmc.list() works", {
 test_that("get_cmdstan_flags() can be used recursively in `make`", {
   mkfile <- normalizePath(test_path("testdata", "Makefile"))
   nonrecursive_flags <- get_cmdstan_flags("STANCFLAGS")
-  stdo <- wsl_compatible_run(
-    command = "make", args = sprintf("--file=%s", mkfile), wd = cmdstan_path()
+  stdo <- processx::run(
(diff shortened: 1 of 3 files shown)


pearsonca:
Re unit tests, @tjmahr could you translate your reprex on the issue discussion into a unit test to add?

pearsonca:
@andrjohns, re the other branch code mentioned in the issue:
 - I'm a bit wary of the `any(...)` check in there, but not because of anything specific.
 - It might make sense to and/or check for the `make`-related output (for a `make` flavor that does something funny with `-s`, either not surpressing or overriding the `-print-` bit), but: I'd generally lean towards using the tool flags (explicitly declared interface) over parsing the output (probably-reliable-but-implicit interface).

pearsonca:
Hmm - not sure why the new test is failing in the CI while working locally. The error messages are a bit obscure - seems like a permissions thing, maybe?

I'm going try explicitly adding the cmdstan_path() as the working directory. Though it seems this shouldn't be strictly necessary (the makefile is made w/ `-f` option, so should run anywhere), perhaps it's a matter of `processx` being sandboxed?

tests/testthat/test-utils.R
bbolker, line 1: I love it (NOT!) when constructing a reproducible test takes 100 times longer than fixing the problem in the first place ...

pearsonca: right?

pearsonca: more broadly, i'd say the take away here is to store configuration data in a common data format (e.g. json) which then gets parsed by the associated standard for the pertinent ends.

When running the software engineering workshop, inevitably some PhD students project is like ... 75% screwing around with the custom parser. And then their mind is blown by switching format + some library for ingesting it.

pearsonca:
Using a real file added to `tests/testthat/testdata` seems to have worked. The remaining failing CI items seem ... totally unrelated?

Standing by for any other notes.

bbolker:
Any thoughts from developers about this?  Would love to see it integrated so I can stop making sure that I have the patched version ...

pearsonca:
> Any thoughts from developers about this? Would love to see it integrated so I can stop making sure that I have the patched version ...

I can add a NEWS bullet if needed? assuming the no-ci-triggering note is `[no ci]`?

pearsonca:
> Any thoughts from developers about this? Would love to see it integrated so I can stop making sure that I have the patched version ...

I also don't have convenient access to WSL (i.e. fixing that exceeds my practical availability) and the CI messages aren't enough to go off. I suspect it's not a the-action-is-misconfigured problem (feels like there should be other WSL issues as well?), but can't really tell how the addition of a flag to make invocation has broken things.

andrjohns:
@pearsonca @bbolker sorry for the delay (the joys of publish or perish dictating the week) and thanks for opening this PR! It all looks good to me, I'll investigate and fix the WSL failure in the next couple of days and merge

pearsonca:
huh, bit wild that the fix is to *not* use the `wsl_compatible_run()` function?

andrjohns:
> huh, bit wild that the fix is to _not_ use the `wsl_compatible_run()` function?

`wsl_compatible_run()` runs the intended command though WSL, which means that the test is trying to call `R` and `cmdstanr` installed in WSL - rather than calling the windows installation

pearsonca:
> > huh, bit wild that the fix is to _not_ use the `wsl_compatible_run()` function?
> 
> `wsl_compatible_run()` runs the intended command though WSL, which means that the test is trying to call `R` and `cmdstanr` installed in WSL - rather than calling the windows installation

Ah; clearly, I have not understood the bowels of `cmdstanr`.

PR #937, opened 2024-03-21
Create inits from fit and draws objects

SteveBronder:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Adds a `process_inits` function to turn fit objects into a list of list of inits. All types of fits from cmdstanr as well as draws objects can be used as inits now.

Tests can be run by calling 

```
testthat::test_file("./tests/testthat/test-fit-init.R")
```

#### Copyright and Licensing

Simon's Foundation


SteveBronder:
@avehtari Sorry for closing the other PR but I started doing some heavy feature creep there and this one is much simpler

avehtari:
An example Pathfinder object which caused a failure with the previous PR is at https://users.aalto.fi/~ave/casestudies/Birthdays/pth5.rds

SteveBronder:
@avehtari when I try loading that into R I'm getting 

```R
 load("~/Downloads/pth5.rds")
Error in load("~/Downloads/pth5.rds") : 
  bad restore file magic number (file may be corrupted) -- no data loaded
In addition: Warning messages:
1: In readChar(con, 5L, useBytes = TRUE) :
  truncating string with embedded nuls
2: file ‘pth5.rds’ has magic number 'X'
  Use of save versions prior to 2 is deprecated 
```

Does that rds file come from one of the scripts in that folder?

avehtari:
The R code is at https://github.com/avehtari/casestudies/blob/master/Birthdays/birthdays.R
and Stan files in https://github.com/avehtari/casestudies/tree/master/Birthdays
and data in https://github.com/avehtari/casestudies/tree/master/Birthdays/data

Which R version you are using?

SteveBronder:
> Which R version you are using?

4.3.0. Maybe I need to upgrade? But I'll run the scripts you sent over and see if that causes my code to break

SteveBronder:
@avehtari running pth5 from the script I got the error about not enough distinct draws which I think is what we want.

```
Not enough distinct draws (4) in pathfinder fit to create inits. Try running Pathfinder with psis_resample=FALSE
```

Turning psis resampling stopped the error so I think that is what we want

avehtari:
Yes, this is what we want. It just didn't work like that with the previous PR. I'll test with this PR

avehtari:
@SteveBronder was using an older `posterior` version. The default value for `pareto_smooth()` argument `return_k` has changed from TRUE to FALSE. I pushed a commit that explicitly sets the argument to FALSE, so that the call should work now both with old and new `posterior` version. Now the Birthday case study runs further. I'll comment after I've been able to run it to the end (need to make some minor changes in my notebook, but can't do those at this minute)

jgabry:
> @SteveBronder was using an older `posterior` version. The default value for `pareto_smooth()` argument `return_k` has changed from TRUE to FALSE. I pushed a commit that explicitly sets the argument to FALSE, so that the call should work now both with old and new `posterior` version. Now the Birthday case study runs further. I'll comment after I've been able to run it to the end (need to make some minor changes in my notebook, but can't do those at this minute)

Ok thanks @avehtari. I think the windows unit tests are failing due to an unrelated reason. Looks like some issue setting up R, which often is resolved at some point without us having to do anything. 

@SteveBronder I've started taking a look and so far it looks great, but I will do a deeper dive after @avehtari finishes his testing in case he finds any larger issues.

venpopov:
just a comment on the windows tests, not related to this PR per se - we are seeing the same failures for other packages, so definitely unrelated to the PR. They seem intermittent, sometimes passing, sometimes failing...

venpopov:
just a comment on the windows tests, not related to this PR per se - we are seeing the same failures for other packages, so definitely unrelated to the PR. They seem intermittent, sometimes passing, sometimes failing...

avehtari:
My Birthdays case study runs now completely with new inits and other recent CmdStanR features. I will next check the documentation added in this PR

R/args.R
```diff
@@ -1116,7 +1170,160 @@ process_init_function <- function(init, num_procs, model_variables = NULL) {
...
-  process_init_list(init_list, num_procs, model_variables)
+  process_init(init_list, num_procs, model_variables)
+}
+
+#' Write initial values to files if provided as a `CmdStanMCMC` class
+#' @noRd
+#' @param init A `CmdStanMCMC` class
+#' @param num_procs Number of inits requested
+#' @param model_variables  A list of all parameters with their types and
+#'   number of dimensions. Typically the output of model$variables().
```
avehtari, line 1181: The output of model$variables() includes also other variables than parameters (data, transformed parameters, and generated quantities). More accurate would be "output of model$variables()$parameters". The same comment repeats in many function docs. As the dimensionality of generated quantities is often much bigger than the parameters, it is not useful to include them in init.

R/args.R
```diff
@@ -1116,7 +1170,160 @@ process_init_function <- function(init, num_procs, model_variables = NULL) {
...
+  if (unique_draws < (0.95 * nrow(draws_df))) {
+    temp_df = stats::aggregate(.draw ~ lw, data = draws_df, FUN = min)
+    draws_df = posterior::as_draws_df(merge(temp_df, draws_df, by = 'lw'))
+    draws_df$pareto_weight = exp(draws_df$lw - max(draws_df$lw))
+  } else {
+    draws_df$pareto_weight = posterior::pareto_smooth(
+      exp(draws_df$lw - max(draws_df$lw)), tail = "right", return_k=FALSE)
+  }
+  init_draws_df = posterior::resample_draws(draws_df, ndraws = num_procs,
+                                            weights = draws_df$pareto_weight, method = "simple_no_replace")
```
avehtari, line 1247: If I read correctly, this resampling is done always, even if Stan did already do resampling. If model$pathfinder(..., psis_resample=TRUE) was used and there are enough unique draws, then the resampling here should be done with uniform weights.

R/args.R
```diff
@@ -1116,7 +1170,160 @@ process_init_function <- function(init, num_procs, model_variables = NULL) {
...
+    stop("We are unable to create initial values from a model with no samples. Please check the results of the model used for inits before continuing.")
+  } else if (!any(names(model_variables$parameters) %in% init$metadata()$stan_variables)) {
+    stop("None of the names of the parameters for the model used for initial values match the names of parameters from the model currently running.")
+  }
+  draws_df = init$draws(format = "df")
+  if (is.null(model_variables)) {
+    model_variables = list(parameters = colnames(draws_df)[2:(length(colnames(draws_df)) - 3)])
+  }
+  init_draws_df = posterior::resample_draws(draws_df, ndraws = num_procs,
+                                            method = "simple")
```
avehtari, line 1323: CmdStanMLE should only have one draw. Resampling just copies that one draw several times. There is no performance hit, but it just looks strange.

avehtari (commented):
Added some comments. In addition, the PR seems to be missing a doc update for the init arguments in model.R

avehtari:
Additional observations using the Birthday case study
- Init with `opt1` (CmdStanMLE object) works, but does not work with `as_draws(opt1)` unless if more than one init is required (multiple chains or pathfinders)
- Init with CmdStanLaplace, CmdStanPathfinder, CmdStanMCMC and corresponding `as_draws()` works
- If draws object has fewer draws than inits, there is an error, but the behavior could be the same as when using CmdStanMLE object, that is resample to get the needed number. This would make sense especially when using a single init value. There could be a warning in cases when a single init is copied.
- warn_partial argument indicates possibility of providing init only for some parameters, but at least with draws object as init, this causes an error
- it would be good to check that the inits don't include -Inf or Inf

SteveBronder:
> Init with opt1 (CmdStanMLE object) works, but does not work with as_draws(opt1) unless if more than one init is required (multiple chains or pathfinders)
> If draws object has fewer draws than inits, there is an error, but the behavior could be the same as when using CmdStanMLE object, that is resample to get the needed number. This would make sense especially when using a single init value. There could be a warning in cases when a single init is copied.

Yes I changed this now so that if the number of draws is less than the number of processes it repeats the inits so that we have the amount of inits as processes. Since all the other process_init functions always return the same number of draws as the number of processes this logic will only happen when the user passes a draws object. 

> warn_partial argument indicates possibility of providing init only for some parameters, but at least with draws object as init, this causes an error

Oh that should just give a warning. I think it makes sense to only have some parameters if for instance you are doing iterative model building

> it would be good to check that the inits don't include -Inf or Inf

Agree! Just added that with a nice error message

@jgabry one more thing. It would be pretty easy to add something here so that if a user passes a sampling object, has not included their own inverse metric, and is doing sampling with the same model / parameters we can use the sampling init object's inverse metric. Do you want me to include that here or should that be another PR?

R/args.R
SteveBronder: Yeah I'll fix this to just rep the draw a bunch of times

R/args.R
SteveBronder: Makes sense!

R/args.R
SteveBronder: 👍

SteveBronder:
@jgabry fyi I upgraded the rtools and r github action for windows and that seemed to fix the errors on the CI

jgabry:
> @jgabry one more thing. It would be pretty easy to add something here so that if a user passes a sampling object, has not included their own inverse metric, and is doing sampling with the same model / parameters we can use the sampling init object's inverse metric. Do you want me to include that here or should that be another PR?

sounds like a good idea. Fine by me to do it here or separately, whatever you prefer. Thanks @SteveBronder!

jgabry:
> > @jgabry one more thing. It would be pretty easy to add something here so that if a user passes a sampling object, has not included their own inverse metric, and is doing sampling with the same model / parameters we can use the sampling init object's inverse metric. Do you want me to include that here or should that be another PR?
> 
> sounds like a good idea. Fine by me to do it here or separately, whatever you prefer. Thanks @SteveBronder!

I guess one question is whether this should happen automatically or if there should be a way for them to specify whether to reuse the metric. Thoughts? In the particular case you mentioned it seems like this should just happen by default but maybe there’s a reason not to?

avehtari:
I would make the use of metric explicit and allow `inv_metric` argument to be a `CmdStanMCMC` object. Getting it from `init` by default can be confusing, and adding an argument to take it from there would make the arguments more complex, than just using the existing `inv_metric` argument.

avehtari:
The natural next step is to allow also getting the step size from a `CmdStanMCMC object`, and then one could turn of the warmup and continue sampling.

SteveBronder:
Let's push that to a separate PR. It sounds like this might be a bigger thing with some Qs about impl

jgabry (changes requested):
This looks great. Thanks @SteveBronder. I just pushed one commit (instead of asking for you to do it in a review comment) that registers the new S3 methods (it doesn't expose them to users, but CRAN now complains if you don't have even internal S3 methods registered in the NAMESPACE file). 

The only thing that I think is missing is to add this new feature to the doc of the `init` argument. It would be in the file `man-roxygen/model-common-args.R`. There's a list of valid `init` types so we just need to add this new one and it would also be great to document briefly how inits are created depending on the type of fitted model object. Do you mind adding that? 

Other than that I think this is ready to go! Thanks again for working on this.

SteveBronder:
Thanks! Yes sorry I'll update the docs today

SteveBronder:
@jgabry just updated the docs, can you take a look to make sure they are thorough enough?

jgabry:
Thanks! I’ll take a look at the doc. Do you mind taking a look at the two unit test runs that failed?

avehtari (changes requested):
Made some edit suggestions

man-roxygen/model-common-args.R
```diff
@@ -44,6 +44,28 @@
 #'  has argument `chain_id` it will be supplied with the chain id (from 1 to
 #'  number of chains) when called to generate the initial values. See
 #'  **Examples**.
+#'  * A [`CmdStanMCMC`], [`CmdStanMLE`], [`CmdStanVB`], or [`CmdStanPathfinder`]
```
jgabry, line 47: Also `CmdStanLaplace`?

man-roxygen/model-common-args.R
```diff
@@ -44,6 +44,28 @@
...
+#'  fit object. If the fit object's parameters are only a subset of the model
+#'  parameters then the other parameters will be drawn by Stan's default
+#'  initialization. The fit object must have at least some parameters that are the
+#'  same name and dimensions as the current Stan model. For the `sampling` and
```
jgabry, line 51:
```suggestion
#'  same name and dimensions as the current Stan model. For the `sample` and
```

jgabry (changes requested):
Doc looks good, just a few minor comments/suggestions (and I also agree with @avehtari's suggestions).

R/args.R
```diff
@@ -1116,7 +1204,172 @@ process_init_function <- function(init, num_procs, model_variables = NULL) {
...
+process_init_approx <- function(init, num_procs, model_variables = NULL,
+                                warn_partial = getOption("cmdstanr_warn_inits", TRUE),
+                                ...) {
+  validate_fit_init(init, model_variables)
+  # Convert from data.table to data.frame
+  draws_df = init$draws(format = "df")
+  if (is.null(model_variables)) {
+    model_variables = list(parameters = colnames(draws_df)[3:(length(colnames(draws_df)) - 3)])
+  }
+  draws_df$lw = draws_df$lp__ - draws_df$lp_approx__
```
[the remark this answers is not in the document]
SteveBronder: I had to rework this. Now if `calculate_lp` is false then the weights are just all set to 1. Is that alright?

man-roxygen/model-common-args.R
```diff
@@ -44,6 +44,28 @@
...
+#'  `pathfinder` method, if the fit object has less samples than the requested
+#'  number of chains/paths then the inits will be drawn using sampling with
+#'  replacement. Otherwise sampling without replacement will be used.
+#'  When a [`CmdStanPathfinder`] fit object is used as the init, if
+#'  `psis_resample` was set to `FALSE` and `calculate_lp` was
+#'  set to `TRUE` (default), then PSIS resampling will be used as weights.
+#'  if `calculate_lp` is `FALSE` then sampling without replacement will be used
+#'  to select the draws.
+#'  PSIS resampling is used to select the draws for  [`CmdStanVB`] fit objects.
```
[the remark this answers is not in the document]
avehtari: Sorry, I meant `CmdStanLaplace`

R/args.R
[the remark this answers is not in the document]
avehtari: Yes

man-roxygen/model-common-args.R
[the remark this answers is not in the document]
SteveBronder: Oh! Yes I'll add that

SteveBronder:
@jgabry I can't figure out this error locally

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-fit-init.R:87:3'): Pathfinder method with calculate_lp as false works as init ──
Expected `test_inits(mod_logistic, fit_path_init, data_list_logistic)` to run without any errors.
ℹ Actually got a <simpleError> with text:
  Assertion on 'files' failed: File does not exist: '/var/folders/c9/jqpw9nhs7jj7vm5185nyw05h0000gn/T/Rtmpsw80z8/working_dir/Rtmpz5369I/logistic-202404181728-1-3f5304.csv'.
```

I there a reason the file would not exist? Maybe the model fails sometimes? But I think that would get caught earlier

jgabry:
Yeah, I just ran the tests locally and I can't reproduce this either. It also seems to be nondeterministic since it's now failing on one of the Mac GHA runs too when previously it was just windows and linux. 

It's a bit tricky to tell from the error message where it's actually failing since the `test_inits` function runs multiple methods and the line number reported in the error is the call to `test_inits`, not the individual line within that function where it errored. Do you think it's worth not using `test_inits` inside that particular test (I think it's just the test starting at line 79 of the file) and instead just put the code to run all the fitting methods right there inside of the `test_that` so we can see which one is actually failing? Then we can switch back after we figure it out?

SteveBronder:
Sure yeah that makes sense

SteveBronder:
I started doing that locally and put in some debugs and am finding some odd things.

When calling `sample` like 

```
    set.seed(1234)
    mod_logistic <- testing_model("logistic")
    fit_path_init <- mod_logistic$pathfinder(seed=1234, data = data_list_logistic,
      refresh = 0, num_paths = 4, psis_resample = TRUE, calculate_lp = FALSE)
    data_list = data_list_logistic
    mod = mod_logistic
    fit_init = fit_path_init
    fit_sample <- mod$sample(data = data_list, chains = 1,
      init = fit_init, iter_sampling = 100, iter_warmup = 100, refresh = 0,
      seed = 1234)
```

It's running sample and then when getting the values back from the csv after cmdstan ran it is reporting an error in `csv.R` that **the files for the sample do not exist**!! That's really weird!! I wonder if this is some sort of weird system thing. Like cmdstan is requesting to write these files and for some reason the system is saying no? That would be incredibly weird. I'll make a few more checks to see whatsup

jgabry:
@SteveBronder Anything I can do to help test/debug this?

jgabry:
@SteveBronder Anything I can do to help test/debug this? 

Also I just fixed a merge conflict with master, moved you up in the DESCRIPTION file with the other full authors, and added @venpopov to DESCRIPTION as a contributor. I think we're super close to merging this.

SteveBronder:
Not yet, I want to test if I get this same issue on master from just running a bunch of the algorithms multiple times. It looks like this code is not touching anything related to reading / writing so I'm pretty confused by the failures

SteveBronder:
If you have time to test that today it would be nice but otherwise I'll try that tmrw

jgabry:
I can make a branch off of master than runs basically the same test file you added but overrides the `fit_init` argument to `test_inits` and just passes in old fashioned initial values (since it would otherwise fail if trying to init from a model object).

SteveBronder:
Much appreciated!

SteveBronder:
This and master are getting a weird error in windows latest when trying to build stringi?

SteveBronder:
Besides that though I _think_ I figured out the original bug with the tests

andrjohns:
> This and master are getting a weird error in windows latest when trying to build stringi?

Yeah `stringi` is broken on windows `r-devel` atm, so safe to ignore.

> Besides that though I think I figured out the original bug with the tests

Brilliant! Let me know when you're done with changes and I'll do a final look-over before merging

SteveBronder:
I think it is ready to merge!

andrjohns (approved):
Thanks, this is going to be very well-appreciated!

PR #954, opened 2024-04-22
debug inits tests

jgabry:
Don't merge this. Just additional tests related to https://github.com/stan-dev/cmdstanr/pull/937#issuecomment-2070893827 as discussed with @SteveBronder

### DIFF
--- a/tests/testthat/test-fit-init-debug.R
+++ b/tests/testthat/test-fit-init-debug.R
@@ -3,106 +3,21 @@ set_cmdstan_path()
 
 
 
-data_list_logistic <- testing_data("logistic")
 
-test_inits <- function(mod, fit_init, data_list = NULL) {
-
-  # override fit_init so that all the code below runs
+test_inits <- function() {
   fit_init <- function() {
     list(alpha = rnorm(1), beta = rnorm(3))
   }
-
-  Sys.sleep(1)
-  fit_sample <- mod$sample(data = data_list, chains = 1, init = fit_init,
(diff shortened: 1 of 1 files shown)


SteveBronder:
It looks like it failed because of something else?

jgabry:
It's just failing on R 3.6 because we're now requiring R 4.0. Just merged a PR into that one that should fix that. Aside from the check using R 3.6 I don't see any failures yet.

jgabry:
Ugh, same error. Really strange. Not sure what's going on. 

https://github.com/stan-dev/cmdstanr/actions/runs/8790487616/job/24122605861#step:13:3310

```
── Failure ('test-fit-init-debug.R:61:3'): Pathfinder method with psis_resample as false works as init ──
Expected `test_inits(mod_logistic, fit_path_init, data_list_logistic)` to run without any errors.
ℹ Actually got a <simpleError> with text:
  Assertion on 'files' failed: File does not exist: '/tmp/RtmpNMWjuB/working_dir/RtmpIOmJZ0/logistic-202404222153-1-84d8fc.csv'.
```

jgabry:
@SteveBronder now it's passing everything here, but not sure if that's just because we got lucky (since the error seems nondeterministic). I will trigger the tests to run again.

SteveBronder:
@jgabry I think I figured it out but it's very confusing. Removing the `set.seed()` in the file I changed above will allow let the while loop run forever instead of failing to generate the file. I don't know about about R internals to understand why that would be the issue though? but I've had it that while loop for 10 minutes now without issue

andrjohns:
Closing as issue resolved

PR #984, opened 2024-05-24
fix process_init_approx

avehtari:
When using approximate inference algorithm fit objects as init there could be some errors that this PR fixes

- Issue: `lp__` or `lp_approx__` may contain NaN. NA, or Inf. Fix: replace non-finite `lw` with `-Inf`
- Issue: `pareto_smooth()` was estimating `r_eff`, but for independent draws from approximate normals `r_eff=1` should be used: Fix: add argument `r_eff=1` to `pareto_smooth()` call

### DIFF
--- a/R/args.R
+++ b/R/args.R
@@ -1290,6 +1290,8 @@ process_init_approx <- function(init, num_procs, model_variables = NULL,
     model_variables = list(parameters = colnames(draws_df)[3:(length(colnames(draws_df)) - 3)])
   }
   draws_df$lw = draws_df$lp__ - draws_df$lp_approx__
+  # Replace NaN and Inf with -Inf
+  draws_df$lw[!is.finite(draws_df$lw)] <- -Inf
   # Calculate unique draws based on 'lw' using base R functions
   unique_draws = length(unique(draws_df$lw))
   if (num_procs > unique_draws) {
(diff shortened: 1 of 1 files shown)


R/args.R
```diff
@@ -1290,6 +1290,8 @@ process_init_approx <- function(init, num_procs, model_variables = NULL,
     model_variables = list(parameters = colnames(draws_df)[3:(length(colnames(draws_df)) - 3)])
   }
   draws_df$lw = draws_df$lp__ - draws_df$lp_approx__
+  # Replace NaN and Inf with -Inf
+  draws_df$lw[!is.finite(draws_df$lw)] <- -Inf
```
andrjohns, line 1294: Is this something that needs to be fixed at the Cmd/Stan level? Should this ever be `Inf`/`NaN` with a "good" run?

andrjohns (commented):
Thanks! Looks good, just one question

R/args.R
avehtari: It should be fixed also in Stan, so that all non-finite lp__ values are replaced with -Inf, but while waiting for it, it's good to have here

avehtari: https://github.com/stan-dev/stan/issues/3290

andrjohns: Perfect! I'll merge when tests pass. Thanks!

PR #987, opened 2024-05-27
sampling statement -> distribution statement

avehtari:
The docs for 2.35 will have the change sampling statement -> distribution statement. This PR fixes the only mention in `cmdstanr` + uses new URL for the chapter


jgabry (approved):
Thanks! I think the one GitHub actions failure must be unrelated to this but I will rerun the checks and then we can merge.

PR #993, opened 2024-06-07
Fixes 975 by only removing leftmost array dimension if equal to 1

### LINKED ISSUE #975: Dimension mismatch when using Pathfinder object as initial values
Hi,

The following Stan program declares `mu` as a matrix with 1 column.

```
data {
  int N;
  vector[N] y;
}

parameters {
  matrix[N, 1] mu;
  real<lower=0> sigma;
}

model {
  target += normal_lupdf(y | mu[:, 1] , sigma);
}

```

After running Pathfinder and using this object as initial values for HMC, the following error is reported:

```
mod <- cmdstan_model("test.stan")
data <- list(N = 100, y = rnorm(100))
pf <- mod$pathfinder(data = data)
fit <- mod$sample(data = data, init = pf)


Chain 1 Unrecoverable error evaluating the log probability at the initial value.
Chain 1 Exception: mismatch in number dimensions declared and found in context; processing stage=parameter initialization; variable name=mu; dims declared=(100,1); dims found=(100)
```

So it seems like Stan doesn't recognise that `mu` should be a matrix.

Thanks,

Matt

SteveBronder:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #975 by removing only the leftmost dimension of the init if it is equal to 1. Previously the code was calling `drop()` but that drops all dimensions equal to 1. @andrjohns I'd appreciate if you can check my logic here, but I think with a subset of draws we want to drop only the first dimension if it is equal to 1.  This passes all the tests I have setup for weird dimension sized parameter types so I think this logic is correct but I'd appreciate another set of eyes.

#### Copyright and Licensing

Simons Foundation

@mhollanders would you mind installing the branch of cmdstanr and checking whether this works for the larger model you are working with? I have this passing your minimal test case but would rather be certain I'm not just catching a small edge case


andrjohns:
Thanks! I'll sort out the windows tests and review tomorrow

andrjohns:
@SteveBronder for some reason the test model only had a single unique `lw` value at [this step](https://github.com/stan-dev/cmdstanr/blob/356fa046ac63c8800e4162bc10690fb5784c9269/R/args.R#L1296) in `process_init_approx` when `psis_resample=TRUE` - but only on Windows, and only when called in `testthat::expect_no_error()` (which is all kinds of bizarre)

I've set the test to use `psis_resample=FALSE` for now, but it might suggest there's some kind of weird interaction/issue with the process on Windows

andrjohns (approved):
LGTM thanks!

mhollanders:
> @mhollanders would you mind installing the branch of cmdstanr and checking whether this works for the larger model you are working with? I have this passing your minimal test case but would rather be certain I'm not just catching a small edge case

I checked and it works with my model. Thanks!

PR #1057, opened 2025-01-14
fix translation of log liks before exponentiation

avehtari:
Should fix https://github.com/stan-dev/loo/issues/272. 

`$loo()` had the following line
```
r_eff <- loo::relative_eff(exp(LLarray + max(-LLarray)), cores = r_eff_cores)
```
It was possible to get `Inf` from exp, and the autocovariance function failed with input which was all `Inf`. There are two issued with this line, 1) it's using the maximum of whole -LLarray, but it would be better to translate each variable in LLarray with its own maximum, and 2) it's making the smallest value 0, while it should be the largest value which is 0. This PR fixes these two issues.


andrjohns (approved):
Great, thanks Aki!

PR #1118, opened 2025-12-01
New pkgdown theme and workflow

VisruthSK:


jgabry:
> The site has been pushed to gh-pages but the deployment should be checked once by swapping the repo. The GHA didn't fail so I hope it works but should check.

If I change it to gh-pages in the repo settings will that automatically trigger the deployment?

VisruthSK:
Yeah it should work

jgabry:
Ok I'll try changing the settings now

PR #1119, opened 2025-12-03
Trying to remove red logo

VisruthSK:
Trying to move to the new, blue logo in the website.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -153,30 +153,14 @@
 #' fit_optim_w_init_list$init()
 #' }
 #'
-cmdstan_model <- function(
-  stan_file = NULL,
-  exe_file = NULL,
-  compile = TRUE,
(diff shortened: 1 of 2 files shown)


VisruthSK:
Simple fix, looks like the correct, blue favicon is live on dev site.

R/model.R
```diff
@@ -1,6 +1,6 @@
 #' Create a new CmdStanModel object
 #'
-#' @description \if{html}{\figure{logo.png}{options: width=25}}
+#' @description \if{html}{\figure{logo.svg}{options: width=25}}
```
jgabry: I think this is the only necessary change in the model.R file but there's a huge diff for the file due to a bunch of linting. Can you remove that from this PR and just keep the logo change in model.R?

PR #1125, opened 2025-12-26
Add qs2 option for saving model objects

### LINKED ISSUE #1040: Add `qs2` fomat to `save_object`
`qs2` is a very fast format for saving and reading data. For buge models, `fit$save_object()` is slow in `rds` format.

It would be great to ass the `qs2` (or `qs`) option to  `save_object`.

Thanks!

VisruthSK:
Fixes #1040 by offering the option to use the qs2 format which performed well in tests across model sizes.


VisruthSK:
Should be good once the data.table change is reflected in tests, since I think all tests would pass then.

PR #1138, opened 2026-02-13
Add a `progressr`-based progress bar for sampling.

### LINKED ISSUE #18: Progress bar
CmdStanPy has a nice progress bar that we can implement in R too. 

We are currently using `processx::run(..., echo=TRUE)` so that the R user sees the output directly from CmdStan, but processx provides a lot more control than that and it looks like it has what we’d need behind the scenes for a progress bar even for chains in parallel.

josswright:
#### Submission Checklist

- [x] Run unit tests

#### Summary

As discussed in https://github.com/stan-dev/cmdstanr/issues/18 , this pull request adds a progress bar to CmdStanR's `sample()` function via the `progressr` framework.

This can be enabled via two new options to `model$sample(...)`:
- `show_progress_bar` - boolean - default `FALSE`
   Enables the progress bar.  
- `suppress_iteration_messages` - boolean - default `FALSE` if `show_progress_bar` is `FALSE`, and `TRUE` if `show_progress_bar` is `TRUE`.
   Causes any output lines reporting an iteration (such as `Chain 1 Iteration   1 / 2000 (Warmup)`) to be suppressed.

For the progress bar to be displayed, a progress bar handler must be created and registered by the end user. (See: https://progressr.futureverse.org/) For ease of use, a `register_default_progress_handler()` function has been added. Calling `register_default_progress_handler()` before a `model$sample(...)` call will spawn a default progress bar via the [cli](https://cli.r-lib.org/index.html) library.

#### Copyright and Licensing

Joss Wright, University of Oxford


VisruthSK:
Looks like some tests are failing because the lower bound for iter_sampling should be 0 instead of 1.

josswright:
Thanks for the help with the unit tests.

In the spirit of peer review, some notes and thoughts:
- I'm not entirely happy with `CmdStan`'s default `iter_warmup` and `iter_sampling` values being hard-coded into `MCMCProcs` default call rather than derived somehow. I imagine that the default values are fairly fixed, but I would have liked not to rely on that. Every route I could see to getting them was clunky and brought in unnecessary code.
- As mentioned above, I toyed with the idea of setting `refresh` to 1 by default when showing the progress bar with iteration messages suppressed. In the end that seemed to introduce yet another case of a parameter depending on another parameter, as well as a potential source of fragility. I'll note, though, that `refresh` can safely be set to 1 for a nice, smooth progress bar without any noticeable overhead or slowdown that I saw in testing.
- The code to derive the number of steps for the progress bar is more complicated than I originally imagined because of how `CmdStan` reports iterations. 
 
  At one stage, when the number of steps was incorrectly calculated as too low, the progress bar was crashing, which also killed the sampling run. I've now ensured that the progress bar automatically closes when it reaches its finished number of steps, which should avoid crashes and just print a warning in the case of some output oddity causing more update steps to be signalled. 

  I've tested as extensively as I can, but I live in fear of this killing someone's multi-day sampling run.

R/run.R
```diff
@@ -1050,6 +1111,53 @@ CmdStanMCMCProcs <- R6::R6Class(
...
+              # iterations. (For example, if refresh_rate is 12 and iter_warmup
+              # is 100, the first reported iteration for sampling will be 112,
+              # not 108.)
+
+              # Get the current iteration count.
+              # Subtract iter_warmup if greater than that.
+              iter_current <- as.numeric(gsub(".*Iteration:\\s*([0-9]+) \\/.*", "\\1", line, perl = TRUE))
+              if (iter_current > private$iter_warmup_) {
+                iter_current <- iter_current - private$iter_warmup_
+              }
```
jgabry: Question about the logic here (and below). Is my understanding here correct? 

  1. You parse Iteration: X / ... from CmdStan output.
  2. If X > iter_warmup you subtract warmup, so iter_current becomes sampling-only (e.g., 1..iter_sampling)
  3. But then you check the final-step (below) with iter_current == iter_warmup + iter_sampling, which now back on the total iterations scale.

So during sampling, can that final-step condition can ever actually be true? I'm not entirely confident in my analysis here, but here's a simple example (not very realistic, but just to demonstrate). Suppose: 

- iter_warmup = 100
- iter_sampling = 70
- Total iterations per chain = 170
- refresh = 50

Then at the final line (Iteration: 170 / 170):

  - We would subtract warmup: iter_current = 170 - 100 = 70
  - Then check 70 == 170 (false)
  - 70 %% 50 != 0, so no progress update happens for the final chunk?

R/run.R
```diff
@@ -511,6 +511,10 @@ check_target_exe <- function(exe) {
     }
     procs$check_finished()
   }
+  # Ensure at this point that any created progress bar is closed.
+  if (!is.null(private$progress_bar_)) {
+    private$progress_bar_(type = "finish")
```
jgabry: Should this be `procs$progress_bar()(type = "finish")`? Does `private$progress_bar` exists in the CmdStanRun object or just in CmdStanProcs?

R/utils.R
```diff
@@ -1046,3 +1046,48 @@ expose_stan_functions <- function(function_env, global = FALSE, verbose = FALSE)
...
+    cli.progress_clear = FALSE
+  )
+
+  # Default informative progress output for sampling
+  progressr::handlers(
+    progressr::handler_cli(
+      format = "{cli::pb_spin} Progress: |{cli::pb_bar}| {cli::pb_current}/{cli::pb_total} | {cli::pb_percent} | ETA: {cli::pb_eta}",
+      clear = FALSE
+    )
+  )
```
jgabry: These settings will affect things people do outside of CmdStanR too, right? I would think it would be very rare that this would cause a problem for someone, but if we change global settings we should document it. 

Typically I would approach something like this by restoring the options to their previous state before exiting the function, but that doesn't make sense here. The user would then need to call this function before every sampling run instead of just once. So documenting this seems sufficient. Maybe something like this (if what I wrote is actually correct): 

```
 `register_default_progress_handler()` modifies global progress settings for the current R session. It installs a global progressr handler and sets cli progress options, which may change progress display behavior in other packages. 
```

Another option (in addition to the doc) would be to add a function the user can use to restore the settings. 

@VisruthSK any thoughts on this?

jgabry (commented):
Thanks again for working on this! I just added a first round of review comments. And here are also some responses to some of your comments above: 

> In the spirit of peer review, some notes and thoughts:
> 
> * I'm not entirely happy with `CmdStan`'s default `iter_warmup` and `iter_sampling` values being hard-coded into `MCMCProcs` default call rather than derived somehow. I imagine that the default values are fairly fixed, but I would have liked not to rely on that. Every route I could see to getting them was clunky and brought in unnecessary code.

Yeah not ideal, but this is tricky. I'll think about it but we may be stuck with this. 

> * As mentioned above, I toyed with the idea of setting `refresh` to 1 by default when showing the progress bar with iteration messages suppressed. In the end that seemed to introduce yet another case of a parameter depending on another parameter, as well as a potential source of fragility. I'll note, though, that `refresh` can safely be set to 1 for a nice, smooth progress bar without any noticeable overhead or slowdown that I saw in testing.

When I test `refresh = 1` it's nice and smooth but it seems to add a _lot_ of overhead. Here's what I tried along with `system.time` output: 

```
system.time({
  cmdstanr_example(show_progress_bar = TRUE)
})
user  system elapsed 
3.235   0.982   4.106 

system.time({
  fit <- cmdstanr_example(show_progress_bar = TRUE, refresh = 1)
})
   user  system elapsed 
182.875  13.757 199.954 
```

Drastic difference! Are you not seeing that on your end? Maybe I'm doing something wrong. This is also all with the default progress bar. Maybe other ones have less overhead? 


> * The code to derive the number of steps for the progress bar is more complicated than I originally imagined because of how `CmdStan` reports iterations.

I put a few questions/comments about this in the review comments. 


>   I've tested as extensively as I can, but I live in fear of this killing someone's multi-day sampling run.

I haven't yet gotten around to thinking about what tests we should add to the unit tests. But we should definitely add some that will run automatically whenever our unit tests are run.

avehtari:
1. Can you add this also for other methods? At least `laplace` and `pathfinder` take often long time and provide iteration output.
2. Could it be possible to have global option, e.g. `cmdstanr_progressbar` with default value `FALSE`? Then there would not be need to change it for every call separately. This would be especially handy when working with notebooks, so that you like to see the progressbar when running code interactively, but not when rendering the notebook. For example, `projpred` has global option `projpred.use_progressr`
3. Seems to work also with `brms` when using `options(brms.backend = "cmdstanr")`
4. If `register_default_progress_handler()` is called when quarto rendering, there is an error. The function could use `if (interactive()) {... }` to register the handler only in interactive session.
```
[31mError in `globalCallingHandlers()`:
! should not be called with handlers on the stack
Backtrace:
    ▆
 1. └─cmdstanr::register_default_progress_handler()
 2.   └─progressr::handlers(global = TRUE) at cmdstanr/R/utils.R:1073:3
 3.     └─progressr:::register_global_progression_handler(action = action)
 4.       └─base::globalCallingHandlers(condition = global_progression_handler)
```

5. The output in the end looks sometimes silly with progressbar in the middle
```
Start sampling
Running MCMC with 4 chains, at most 2 in parallel...

Chain 1 finished in 0.1 seconds.
Chain 2 finished in 0.1 seconds.
Chain 3 finished in 0.1 seconds.
🌔  Progress: |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ | 8000/8000 | 100% | ETA:  0s
Chain 4 finished in 0.1 seconds.
```

josswright:
> > 1. Can you add this also for other methods? At least `laplace` and `pathfinder` take often long time and provide iteration output.
> 
> I think this would be great, but we don't need to wait for that for this PR. It could be done separately and even potentially by someone else (using the same approach), depending on how much time @josswright has available to dedicate to this.

I'd be happy to have a look, although I wouldn't want to give any guarantees about how much time I'm likely to have in the near future. 

In that case, though, it would almost certainly be worth abstracting some of the iteration extraction and calculation code, though.

jgabry:
> Given this, I wonder if it might be worth throwing a warning if `refresh` is set too low?

Since it could make sense to set `refresh=1` for slower models (like your GP model), maybe just adding some documentation about this is enough instead of throwing a warning? We can just say that the overhead of setting `refresh` to a low number will be noticeable for fast models but `refresh=1`  can give you "a nice, smooth progress bar" for slower models, like you mentioned in some comments above.

PR #1170, opened 2026-04-04
Test snapshots

VisruthSK:
Swap some tests to use snapshots. Should be discussed.


jgabry (commented):
Thank you! At first glance this looks good. I'll take another look soon to take a closer look at exactly which tests we're converting to snapshots. 

For now a question: if we set `cran=FALSE` I assume the snapshot tests should still run on CI, right? If we do eventually submit to CRAN we can't run most of these tests because CRAN won't have CmdStan installed and probably won't let us install it on CRAN systems just to run our tests. So we'd probably end up doing what RStudio does with their tensorflow package: 

https://github.com/rstudio/tensorflow/blob/main/tests/testthat.R

VisruthSK:
Ah yes, forgot about that. Will fix that in a bit.

tests/testthat/test-model-compile-user_header.R
```diff
@@ -1,8 +1,7 @@
 skip_if(os_is_macos())
 
-file_that_exists <- "placeholder_exists"
-file_that_doesnt_exist <- "placeholder_doesnt_exist"
-withr::local_file(file_that_exists)
```
jgabry, line 5: This line actually creates a file, but the new version doesn't. The file does happen to get created in one of the later test blocks, but I think we should still create it at the top of the file.

jgabry: Maybe just `file.create(file_that_exists)`, since cleanup should still be handled by `local_tempfile`?

jgabry (changes requested):
I think this is pretty much good to go. I made one small change request and suggested adding one code comment, but otherwise I've come around to the idea that switching to snapshots for these print tests is probably for the best, despite losing the explicitness of the assertions in the old tests.

tests/testthat/helper-custom-expectations.R
```diff
@@ -127,3 +127,20 @@ expect_equal_ignore_order <- function(object, expected, ...) {
 }
 
 expect_not_true <- function(...) expect_false(isTRUE(...))
+
+transform_print_snapshot <- function(x) {
```
jgabry, line 133:
```suggestion
# strips numeric values (which may change slightly with different hardware or compilers) 
# allowing us to still verify names, ordering, column headers, row counts, etc.
transform_print_snapshot <- function(x) {
```

VisruthSK:
Yes the main thing now is that we need to be very protective of the snaps and ensure they aren't changed spuriously. If the API does change its a lot easier to update tests.

jgabry (approved):

jgabry:
> Yes the main thing now is that we need to be very protective of the snaps and ensure they aren't changed spuriously. 

Yeah definitely. We need to make sure no snapshot updates are snuck into large PRs without realizing it.

Issue #1153, opened 2026-03-09
Use air for formatting


VisruthSK:
Use Posit's new opinionated formatter, [Air](https://posit-dev.github.io/air/formatter.html) to [review pull requests](https://posit-dev.github.io/air/integration-github-actions.html#format-with-github-suggestions) to ensure they conform to the style as well as format the entire project once.

VisruthSK:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

PR #1171, opened 2026-04-06
fix: fix init with draws or fit object and without .stan file

### LINKED ISSUE #1088: initiailizing with a draws or fit object fails if the model is created without a .stan file
**Describe the bug**
If a model is created without a `.stan` file, that is only with an `exe_file`, using another fit object as an `init` fails with an uninformative error message.

**To Reproduce**
```r
library(cmdstanr)

model_code <- "data {
 int N;
 array[N] real x;
 }
parameters {
  real mu;
}
model {

  // likelihood including constants
  x ~ normal(mu, 1);
}

"

model <- cmdstan_model(write_stan_file(model_code), exe_file = "model_bin")

model_nostanfile <- cmdstan_model(exe_file = "model_bin")

d <- list(x = c(1,2,3,4,5), N = 5)

pth <- model$pathfinder(
  data = d
)

# works fine
mcmcfit <- model$sample(
  data = d,
  init = pth
)

# fails  Error: 'init' contains empty lists.
mcmcfit_nostanfile<- model_nostanfile$sample(
  data = d,
  init = pth
)

```

**Expected behavior**
The sampling succeeds, or a more informative error message is given.

**Operating system**
Linux (Ubuntu 24.04.2)

**CmdStanR version number**
0.9.0

**Additional context**
I believe this is due to `model_nostanfile$variables()` failing to determine which parameters are in the model, but I think passing all the variable inits without determining which variables are actually in the model should still work

avehtari:
Closes #1088 

Fix init from fit objects when model has no Stan file

When a `CmdStanModel` is created with only `exe_file` (no `.stan` file), using a fit object (e.g. from `$pathfinder()`) as `init` fails with the uninformative error `"'init' contains empty lists."`.

#### Root cause

Without a Stan file, `model$variables()` is unavailable, so `model_variables` is `NULL` throughout the init processing pipeline. Four functions (`process_init.CmdStanMCMC`, `process_init_approx`, `process_init.CmdStanPathfinder`, `process_init.CmdStanMLE`) had fallback code that constructed `model_variables$parameters` as a **character vector** of column names, but `process_init.draws()` expects it to be a **named list** with `$dimensions` metadata. This caused `names(model_variables$parameters)` to return `NULL`, producing empty init lists.

#### Fix

- Added `model_variables_from_draws()` helper that infers parameter names and number of dimensions from `posterior::variables(as_draws_df(...))`: names without `[` are scalars, names with `[` are containers (base name extracted via `sub("\\[.*", "", ...)`), and the number of `,` determine the number of dimensions.
- `process_init.draws()` calls this helper when `model_variables` is `NULL`, so it works without a Stan file.
- Removed the broken column-index-based fallbacks from the four intermediate `process_init.*` methods — they now pass `model_variables` through as-is and let `process_init.draws()` handle the `NULL` case.
- Cleaned up extra columns (`lw`, `weight`) that `process_init_approx` and the Pathfinder no-lp branch add to draws before passing downstream.
- "Pathfinder inits do not drop dimensions" test was renamed to "Inits from fit/draws work for exe-only models with various parameter types" and modified to test more different containers, and to include test case for exe-file-only model to .

#### Notes

- Using  `model$variables()` when Stan file is available still has benefit of creating the init list only with the actual model variables.
- Using `model_variables_from_draws()`  has downside of including transformed paramaters and generated quantities, although that is unlikely to have big performance effect.
- When the init is based on a fit  object, it would be possible to to get the names of the parameters and dimensions (and drop transformed parameters and generated quantities) with `fit$variable_skeleton(transformed_parameters = FALSE, generated_quantities = FALSE)` but that causes compilation of model methods which adds overhead, and thus was not added here.
- The design of the fix is by me. Coding and testing was assisted by Claude

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/R/args.R
+++ b/R/args.R
@@ -1035,6 +1035,38 @@ validate_exe_file <- function(exe_file) {
 }
 
 
+#' Build a model_variables structure from a draws object
+#'
+#' When a model has been created without a Stan file,
+#' `model$variables()` is unavailable. This helper infers parameter
+#' names and whether each is a scalar or container from
+#' `posterior::variables(as_draws_df(...))`. In that representation
+#' containers are expanded (e.g. `beta[1]`, `gamma[1,2]`) while
+#' scalars appear as bare names (e.g. `sigma`). A name without `[` is
+#' a scalar; a name with `[` is a container whose base name is
+#' extracted.
+#'
+#' @noRd
+#' @param draws A draws object (any format supported by posterior).
+#' @return A list with a `parameters` element in the same format as
+#'   `model$variables()`. The dimensions is stored as 0 for scalars
+#'   and 1 for containers, which is sufficient for `process_init.draws()`.
+model_variables_from_draws <- function(draws) {
+  df_vars <- posterior::variables(posterior::as_draws_df(draws))
+  df_vars <- df_vars[!grepl("__$", df_vars)]
+  has_bracket <- grepl("\\[", df_vars)
+  scalars <- df_vars[!has_bracket]
+  containers <- unique(sub("\\[.*", "", df_vars[has_bracket]))
+  parameters <- list()
+  for (var_name in scalars) {
+    parameters[[var_name]] <- list(type = "real", dimensions = 0L)
+  }
+  for (var_name in containers) {
+    parameters[[var_name]] <- list(type = "real", dimensions = 1L)
+  }
+  list(parameters = parameters)
+}
+
 #' Generic for processing inits
 #' @noRd
 process_init <- function(init, ...) {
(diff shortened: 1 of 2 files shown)


jgabry (changes requested):
There's an edge case that causes an error. I think it's inferring the wrong number of dimensions for higher dimensional objects that only contain 1 element: 

```r
stan_code <- write_stan_file(
  "
  data { real y; }
  parameters { matrix[1,1] mu; }
  model { y ~ normal(mu[1,1], 1); }
  "
)
mod <- cmdstan_model(stan_code, force_recompile = TRUE)
fit <- mod$sample(data = list(y = 0))
mod_exe <- cmdstan_model(exe_file = mod$exe_file())

# both of these error 
mod_exe$sample(data = list(y = 0), chains = 1, init = fit$draws())
mod_exe$sample(data = list(y = 0), chains = 1, init = fit)
```

I think this would also error for `array[1,1] real`, `array[1,1,1] real`, etc., although I haven't tested those. These probably aren't very common in Stan programs, but could exist.

jgabry (approved):
Looks good now, thanks. We can merge if everything passes. (I also pushed one commit cleaning up a couple of trivial unrelated things in args.R that I noticed while I was looking at it. I probably should do that separately from this PR, but it was just changing some `=` to `<-` for assignment)

jgabry:
Actually @VisruthSK maybe we should do this before #1153 so that the diffs are separate?

PR #1174, opened 2026-04-08
fix: handle Stan tuple and complex types

### LINKED ISSUE #925: NA dimensions for complex and tuple types
**Describe the bug**

When calling `$draws()` on a fit object whose model has complex or tuple types in it I receive warning messages about NA's being made for the variable sizes.

**To Reproduce**

I made the stan program below to test this

```stan
parameters {
  real real_p;
  vector[2] vector_p;
  matrix[2, 2] matrix_p;
  array[2] matrix[2, 2] array_matrix_p;
  corr_matrix[2] corr_p;
  complex complex_p;
  complex_matrix[2, 2] complex_matrix_p;
  complex_vector[4] complex_vector_p;
  tuple(real, vector[3], array[2] matrix[2, 2], complex) tuple_int_vector_arraymatrix_complex_p;
  array[2] tuple(real, tuple(vector[2], array[2] tuple(real, complex, matrix[2, 2]))) arraytuple_big_p;
}
```

You can get the error with the following

```
# I put this in the tests/testthat/resources/stan folder
mod_params <- testing_model("parameter_types")
fit <- mod_params$pathfinder(data=data_list, seed=1234, refresh = 0)
foo = fit$draws(format = "df")
# Warning messages:
# 1: In variable_dims(metadata$variables) : NAs introduced by coercion
# 2: In variable_dims(metadata$variables) : NAs introduced by coercion
# 3: In variable_dims(metadata$variables) : NAs introduced by coercion
# 4: In variable_dims(metadata$variables) : NAs introduced by coercion
# 5: In variable_dims(metadata$variables) : NAs introduced by coercion
meta = fit$metadata()
meta$stan_variable_sizes
## Output
$lp_approx__
[1] 1

$lp__
[1] 1

$real_p
[1] 1

$vector_p
[1] 2

$matrix_p
[1] 2 2

$array_matrix_p
[1] 2 2 2

$corr_p
[1] 2 2

$complex_p
[1] NA

$complex_matrix_p
[1]  2  2 NA

$complex_vector_p
[1]  4 NA

$`tuple_int_vector_arraymatrix_complex_p:1`
[1] 1

$`tuple_int_vector_arraymatrix_complex_p:2`
[1] 3

$`tuple_int_vector_arraymatrix_complex_p:3`
[1] 2 2 2

$`tuple_int_vector_arraymatrix_complex_p:4`
[1] NA

$arraytuple_big_p
[1] NA NA  2  2
```

**Expected behavior**

Tuples and complex types would have the correct size

**Operating system**
Linux redhat

**CmdStanR version number**
master [28329d7](https://github.com/stan-dev/cmdstanr/commit/28329d7c22bc995d1c15840de8a1bbc212d46e92)

**Additional context**
I found this while working on the functions for adding pathfinder types

avehtari:
Closes #925 

Stan examples by me, @spinkney and @WardBrian (from https://discourse.mc-stan.org/t/proof-of-concept-binary-output-format-for-cmdstan/40846/67). Careful iterative prompting by me. Code , tests and PR description assisted by Claude.

#### Summary

CmdStan 2.38+ introduced tuple and complex types, which use new naming
conventions in CSV output that CmdStanR did not understand. This PR adds
full support for these types across metadata parsing, model methods, and
init value handling.

- Fix `repair_variable_names()` and `variable_dims()` to handle complex
  suffixes (`.real`/`.imag`) and tuple separators (`:`) in CSV column names
- Fix `variable_skeleton()` and `unconstrain_draws()` for models with
  tuple and complex parameters
- Add tuple init support: `mod$sample(init = fit)` now correctly passes
  tuple parameter values as nested JSON objects
- Add `write_stan_json()` support for tuple values as named lists

#### Problem

CmdStan CSV headers now use three different separators:

| Separator | Meaning | Example |
|---|---|---|
| `.` between digits | array/matrix index | `beta.1.2` → `beta[1,2]` |
| `:` | tuple element | `b_tuple:1:1.2` → `b_tuple:1:1[2]` |
| `.real` / `.imag` | complex part | `z.real` → `z[real]` |

These can combine: `arr_pair.1:1` (array + tuple), `z3D.1.1.1.real`
(array + complex), `nested:2:2.real` (tuple + complex).

Before this PR, the following issues occurred:

1. **Warning**: `NAs introduced by coercion` from `variable_dims()` on
   every model with complex or tuple types — `as.numeric("imag")` returns
   NA.

2. **Wrong metadata**: `fit$metadata()$stan_variable_sizes` contained `NA`
   for complex variables and incorrect dimensions.

3. **Broken `variable_skeleton()`**: Returned `NA` names for tuple
   parameters because `create_skeleton()` couldn't match Stan-level names
   (`"b_tuple"`) against C++ leaf names (`"b_tuple.1.1"`).

4. **Crashing `unconstrain_draws()`**: Tuple parameters were silently
   dropped from the draw subset (Stan-level name `"b_tuple"` not found in
   leaf names `"b_tuple:1:1"`), causing CmdStan to receive the wrong number
   of scalars.

5. **Missing init values**: `mod$sample(init = fit)` dropped tuple
   parameters entirely, and `write_stan_json()` could not serialize tuple
   values (heterogeneous lists crashed `list_to_array()`).

##### `R/csv.R`

**`repair_variable_names()`** — Detects and strips `.real`/`.imag` suffixes
before the dot-to-bracket conversion, then re-attaches them in the correct
position. The `:` tuple separator passes through unchanged.

**`variable_dims()`** — For non-numeric indices (`"real"`, `"imag"`, tuple
indices like `"1:2"`), counts unique values across all entries for that
dimension position instead of calling `as.numeric()`.

##### `R/utils.R`

New helper functions for bridging between Stan-level names and leaf names:

- **`stan_param_has_leaf()`** — Checks if Stan-level names have matching
  leaf names using `":"` prefix matching.
- **`expand_stan_params_to_leaves()`** — Expands Stan-level names to
  their leaf equivalents for `posterior::subset_draws()`.
- **`is_tuple_type()`** — Detects tuple parameters from `model_variables`
  type info (`$type` is a list for tuples).
- **`build_tuple_init_value()`** — Recursively reconstructs a nested
  named-list init value from flat leaf draws, also validating for NA/Inf.
- **`.extract_draw_value()`** — Shared helper for the draw extraction
  pipeline.

**`create_skeleton()`** — Expands tuple Stan-level names to leaf components
from `param_metadata_` using `"."` prefix matching.

##### `R/fit.R`

**`unconstrain_draws()`** and **`unconstrain_variables()`** — Use
`stan_param_has_leaf()` for the zero-length parameter check and
`expand_stan_params_to_leaves()` for draw subsetting.

##### `R/args.R`

**`validate_fit_init()`** — Uses `stan_param_has_leaf()` instead of `%in%`.

**`process_init.draws()`** — Separates parameters into tuple and non-tuple.
For tuples, uses `build_tuple_init_value()` to reconstruct nested init
values. NA/Inf validation is done inline during extraction.

##### `R/data.R`

**`write_stan_json()`** — Detects tuple-style named lists (keys `"1"`,
`"2"`, ...) and processes them recursively instead of calling
`list_to_array()`.

New helpers: `is_tuple_list()`, `prepare_tuple_for_json()`.

##### Tests

**`tests/testthat/resources/stan/tuple_complex.stan`** — Test model with
tuple and complex types as both parameters and generated quantities:
nested tuples, tuples with complex, arrays of tuples, complex vectors,
complex matrices, arrays of complex matrices.

**`tests/testthat/test-tuple-complex.R`** — 137 tests covering:
- Unit tests for `repair_variable_names`, `unrepair_variable_names`,
  `variable_dims` with all tuple/complex name patterns
- Helper function tests (`stan_param_has_leaf`, `expand_stan_params_to_leaves`)
- `write_stan_json` with tuple values
- `build_tuple_init_value` reconstruction
- End-to-end: sampling with no warnings, correct metadata, `variable_skeleton()`,
  `unconstrain_draws()`, `init = fit` round-trip, manual init lists

#### Test plan

- [x] All 397 existing `test-csv.R` tests pass
- [x] All 72 existing `test-data.R` tests pass
- [x] All 15 existing `test-fit-init.R` tests pass
- [x] 137 new `test-tuple-complex.R` tests pass
- [x] Manual verification with three Stan models covering: scalar/array/matrix
  params, complex scalar/vector/matrix/array params, simple/nested/arrayed
  tuples, tuples as parameters and generated quantities

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/R/args.R
+++ b/R/args.R
@@ -1096,37 +1096,77 @@ process_init.draws <- function(init, num_procs, model_variables = NULL,
                                        method ="simple_no_replace")
   }
   draws_rvar = posterior::as_draws_rvars(draws)
-  variable_names <- variable_names[variable_names %in% names(draws_rvar)]
-  draws_rvar <- posterior::subset_draws(draws_rvar, variable = variable_names)
+
+  # Separate tuple and non-tuple parameters. Tuple parameters use leaf names
+  # in draws (e.g., "b_tuple:1:1") rather than the Stan-level name ("b_tuple"),
+  # so they need special handling via build_tuple_init_value().
+  is_tuple <- if (!is.null(model_variables)) {
+    vapply(variable_names, function(nm) {
+      is_tuple_type(model_variables$parameters[[nm]])
+    }, logical(1))
+  } else {
+    rep(FALSE, length(variable_names))
+  }
+  tuple_names <- variable_names[is_tuple]
+  scalar_names <- variable_names[!is_tuple]
+
+  # Filter non-tuple names to those present in draws
+  scalar_names <- scalar_names[scalar_names %in% names(draws_rvar)]
+
+  # For tuple names, check that their leaf draws exist
+  rvar_names <- names(draws_rvar)
+  tuple_names <- tuple_names[vapply(tuple_names, function(nm) {
+    any(startsWith(rvar_names, paste0(nm, ":")))
+  }, logical(1))]
+
+  all_names <- c(scalar_names, tuple_names)
+
+  if (length(scalar_names) > 0) {
+    draws_rvar <- posterior::subset_draws(
+      draws_rvar,
+      variable = expand_stan_params_to_leaves(all_names, rvar_names)
+    )
+  }
+
   inits = lapply(1:num_procs, function(draw_iter) {
-    init_i = lapply(variable_names, function(var_name) {
-      x = .remove_leftmost_dim(posterior::draws_of(
-        posterior::subset_draws(draws_rvar[[var_name]], draw=draw_iter)))
+    bad_names <- character(0)
+
+    # Extract non-tuple parameters
+    init_i = lapply(scalar_names, function(var_name) {
+      x = .extract_draw_value(var_name, draws_rvar, draw_iter)
+      if (any(is.infinite(x)) || any(is.na(x))) {
+        bad_names[[length(bad_names) + 1L]] <<- var_name
+      }
       if (model_variables$parameters[[var_name]]$dimensions == 0) {
         return(as.double(x))
       } else {
         return(x)
       }
     })
-    bad_names = unlist(lapply(variable_names, function(var_name) {
-      x = drop(posterior::draws_of(drop(
-        posterior::subset_draws(draws_rvar[[var_name]], draw=draw_iter))))
-      if (any(is.infinite(x)) || any(is.na(x))) {
-        return(var_name)
+    names(init_i) <- scalar_names
+
+    # Extract tuple parameters (build_tuple_init_value also validates)
+    for (var_name in tuple_names) {
+      tuple_result <- build_tuple_init_value(
+        var_name, model_variables$parameters[[var_name]],
+        draws_rvar, draw_iter
+      )
+      init_i[[var_name]] <- tuple_result$value
+      if (length(tuple_result$bad_leaves) > 0) {
+        bad_names <- c(bad_names, var_name)
       }
-      return("")
-    }))
-    any_na_or_inf = bad_names != ""
-    if (any(any_na_or_inf)) {
-      err_msg = paste0(paste(bad_names[any_na_or_inf], collapse = ", "), " contains NA or Inf values!")
-      if (length(any_na_or_inf) > 1) {
+    }
+
+    if (length(bad_names) > 0) {
+      err_msg = paste0(paste(bad_names, collapse = ", "), " contains NA or Inf values!")
+      if (length(bad_names) > 1) {
         err_msg = paste0("Variables: ", err_msg)
       } else {
         err_msg = paste0("Variable: ", err_msg)
       }
       stop(err_msg)
     }
-    names(init_i) = variable_names
+
     return(init_i)
   })
   return(process_init(inits, num_procs, model_variables, warn_partial))
@@ -1245,7 +1285,7 @@ validate_fit_init = function(init, model_variables) {
   # Convert from data.table to data.frame
   if (all(init$return_codes() == 1)) {
     stop("We are unable to create initial values from a model with no samples. Please check the results of the model used for inits before continuing.")
-  } else if (!is.null(model_variables) &&!any(names(model_variables$parameters) %in% init$metadata()$stan_variables)) {
+  } else if (!is.null(model_variables) && !any(stan_param_has_leaf(names(model_variables$parameters), init$metadata()$stan_variables))) {
     stop("None of the names of the parameters for the model used for initial values match the names of parameters from the model currently running.")
   }
 }
--- a/R/csv.R
+++ b/R/csv.R
@@ -923,11 +923,36 @@ check_csv_metadata_matches <- function(csv_metadata) {
 }
 
 # convert names like beta.1.1 to beta[1,1]
+# also handles complex suffixes (.real/.imag) and tuple separators (:)
 repair_variable_names <- function(names) {
+  # 1. Detect and strip .real/.imag suffix before dot conversion
+  complex_suffix <- ifelse(
+    grepl("\\.real$", names), ",real",
+    ifelse(grepl("\\.imag$", names), ",imag", "")
+  )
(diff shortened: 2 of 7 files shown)


jgabry:
Thanks Aki! This one might take me some time to review.

jgabry (approved):
I haven't reviewed the code yet but the failure on WSL is because 

```
Error: Additional model methods are not currently available with WSL CmdStan and will not be compiled
```

Adding `skip_if(os_is_wsl())` to those specific tests should fix it. 

EDIT: oops I meant to "request changes" not "approve" yet

jgabry (changes requested):
I haven't reviewed the code yet but the failure on WSL is because a few of them error with 

```
Error: Additional model methods are not currently available with WSL CmdStan and will not be compiled
```

Adding `skip_if(os_is_wsl())` to those specific tests should fix it.

WardBrian:
Just to add a note, it would be nice if some of these IO utilities were moved somewhere independent, so we could use them from e.g. bridgestan or similar (I know that having more packages is more CRAN pain...)
https://github.com/stan-dev/stanio/issues/7

jgabry:
I still haven't had a chance to go through this myself yet (I definitely will at some point), but I asked codex to take a look to see if it could find anything that's clearly an issue: 

> 
>   1. Comment on tuple detection in write_stan_json():
> 
>   I don't think `write_stan_json()` can safely infer "tuple" from names alone here. This function has
>   historically treated lists as array-like containers, and it does not have model-type information to
>   tell whether a sequentially named list is actually meant to be a tuple.
> 
>   For example, `write_stan_json(list(x = list("1" = 1:3, "2" = 4:6)), ... )` used to serialize `x` as an
>   array, but with this change it becomes a JSON object with `"1"` and `"2"` keys. That looks like a
>   breaking change for existing code that happens to produce named lists via `split()`, `setNames()`, or
>   subsetting.
> 
>   Other than backwards compatibility, I don't think those names add useful semantics for Stan arrays
>   anyway. If we want tuple serialization here, I think it needs to be driven by model metadata or an
>   explicit opt-in, not by a name heuristic.
> 
>   2. Comment asking for a regression test around the ambiguity:
> 
>   Can we add a regression test for the ambiguous non-tuple case too? Right now the new tests cover tuple
>   serialization, but they don't cover the case where a sequentially named list should still behave like
>   an array.
> 
>   For example, I'd want `write_stan_json(list(x = list("1" = 1:3, "2" = 4:6)), ...)` to keep matching
>   the existing array behavior unless we have explicit type information saying `x` is a tuple. Without
>   that test, it's easy to lock in the new heuristic and accidentally break old callers.
> 
>   3. Comment on tuple root variable filtering:
> 
>   I think tuple support is still incomplete in the variable-filter path. The helpers used by
>   `fit$draws()`, `fit$print()`, and `read_cmdstan_csv(..., variables = ...)` still only expand `name[`
>   prefixes, so tuple roots like `b_tuple`, `pair`, or `nested` won't resolve to their `name:...` leaves.
> 
>   For example, if the CSV contains variables like `b_tuple:1:1[1]`, `b_tuple:1:1[2]`, and
>   `b_tuple:2[1,1]`, I'd expect `fit$draws(variables = "b_tuple")` to return those columns, just like
>   `fit$draws(variables = "theta")` expands to `theta[1]`, `theta[2]`, etc. But with the current matching
>   logic, `"b_tuple"` is treated as not found.
> 
>   I'd update the same prefix-matching logic here to treat `name:` as another expansion case.
> 

I haven't verified yet that these are all actually issues. But if 1 is true then I guess we _could_ break backwards compatibility here because we're going to do v1.0. We could say that any list with names will be interpreted as a tuple. It's not ideal, but it's doable if there's not another solution.

jgabry:
I just opened #1287, which is a different approach to this. @avehtari can you try it out? 

@WardBrian I'm still open to moving IO utilities to a separate package, but it's easier for development to have them in cmdstanr until I'm sure they're working properly. Then we could potentially move them into a separate package that cmdstanr depends on, without breaking anything user facing since these are internal functions.

PR #1176, opened 2026-04-08
feat: provide guidance if tmp csv file is not found in Quarto rendering

### LINKED ISSUE #1012: Errors re-rendering quarto html with cached chunks and cmdstanr.
**Describe the bug**

Quarto fails to re-render to html following trivial changes to code chunks that use cmdstanr objects.

```
processing file: 01_data-exclusions.qmd
  |............................................... |  98% [model_1_postsum]    
Quitting from lines  at lines 616-637 [model_1_postsum] (01_data-exclusions.qmd)
Error in `read_cmdstan_csv()`:
! Assertion on 'files' failed: File does not exist: '/var/folders/ms/m90tgryx74362fqn7d65tl7w0000gn/T/Rtmp5yoGqf/model_8cda69ac3b06c9b214d4a12ede85f4f8-202407191622-1-7cf94c.csv'.
Backtrace:
 1. fit01$draws(variables = c("a_bar", "sigma", "p"), format = "df")
 2. private$read_csv_(...)
 3. cmdstanr::read_cmdstan_csv(...)
 4. cmdstanr:::assert_file_exists(files, access = "r", extension = "csv")
 5. checkmate::makeAssertion(files, res, .var.name, add)
 6. checkmate:::mstop(...)
                                                                                                                
Execution halted
```

**To Reproduce**

Turn on Quarto cache in the header: 

```
...
execute:
   cache: true
...
```

Read in a stan model from file using the chunk options: 

```
```{cmdstan file = "path/to/stan/model.stan", output.var = 'm01'}
``
```

Sample:

```
``` {r}
fit01 <- m01$sample( [...] )
``
```
Write some code to summarise the model output or whatever.

Render to html.

Make a small change to anything that calls the model fit object `fit01`. In my case it will happen if I just add an empty line after an unedited chunk.

Try to render to html again.

Error occurs.

**Expected behavior**
The document is rendered after making small changes without the need to clear the cache and refit all models etc,

**Operating system**
MacOs 14.0

**CmdStanR version number**
0.7.1

**Additional context**
None

avehtari:
Quarto caches the fit object which has paths to the output csv files. By default the output csv files are in tmp directory which gets removed when rendering stops. Cache works if the cmdstanr_output_dir is not a temporary directory. Set cmdstanr_output_dir before creating the fit object
```
options(cmdstanr_output_dir = desired_path_to_a_non_temporary_directory)
```

This PR adds an informative message if the csv file is not found, running knitr, and file path has "/tmp"

This PR is pure organic without any AI assistance

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/R/csv.R
+++ b/R/csv.R
@@ -138,7 +138,15 @@ read_cmdstan_csv <- function(files,
     files <- file.path(temp_storage, basename(files))
   }
   format <- assert_valid_draws_format(format)
-  assert_file_exists(files, access = "r", extension = "csv")
+  csv_file_exists <- check_file_exists(files, access = "r", extension = "csv")
+  if (!isTRUE(csv_file_exists)) {
+    if (isTRUE(getOption("knitr.in.progress")) &&
+          any(grepl("/tmp", files))) {
+      stop(paste0(csv_file_exists, "\n  To avoid this error when using Quarto cache,\n  see `cmdstanr_output_dir` in `?cmdstanr_global_options`"))
+    } else {
+        stop(csv_file_exists)
+    }
+  }
   metadata <- NULL
   warmup_draws <- list()
   draws <- list()
--- a/R/csv.R
+++ b/R/csv.R
@@ -141,8 +141,9 @@ read_cmdstan_csv <- function(files,
   csv_file_exists <- check_file_exists(files, access = "r", extension = "csv")
   if (!isTRUE(csv_file_exists)) {
     if (isTRUE(getOption("knitr.in.progress")) &&
-          any(grepl("/tmp", files))) {
-      stop(paste0(csv_file_exists, "\n  To avoid this error when using Quarto cache,\n  see `cmdstanr_output_dir` in `?cmdstanr_global_options`"))
+          is.null(getOption("cmdstanr_output_dir")) &&
+          any(grepl("csv$", files))) {
+      stop(paste0(csv_file_exists, "\n  If this error happened when using Quarto cache,\n  see `cmdstanr_output_dir` in `?cmdstanr_global_options`"))
     } else {
         stop(csv_file_exists)
     }
--- a/R/options.R
+++ b/R/options.R
@@ -27,7 +27,9 @@
 #' * `cmdstanr_output_dir`: The directory where CmdStan should write its output
 #' CSV files when fitting models. The default is a temporary directory. Files in
 #' a temporary directory are removed as part of \R garbage collection, while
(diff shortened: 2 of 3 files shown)


R/csv.R
```diff
@@ -138,7 +138,15 @@ read_cmdstan_csv <- function(files,
     files <- file.path(temp_storage, basename(files))
   }
   format <- assert_valid_draws_format(format)
-  assert_file_exists(files, access = "r", extension = "csv")
+  csv_file_exists <- check_file_exists(files, access = "r", extension = "csv")
+  if (!isTRUE(csv_file_exists)) {
+    if (isTRUE(getOption("knitr.in.progress")) &&
+          any(grepl("/tmp", files))) {
```
jgabry: In the bug report in #1012 the file path was 

```
/var/folders/ms/m90tgryx74362fqn7d65tl7w0000gn/T/Rtmp5yoGqf/model_8cda69ac3b06c9b214d4a12ede85f4f8-202407191622-1-7cf94c.csv
```

which doesn't include `/tmp` anywhere. I think `/tmp` may only be reliable to use on linux.

jgabry (changes requested):

avehtari:
I modified the condition, added explanation to options.R, and added a test for the knitr behavior

R/csv.R
```diff
@@ -138,7 +138,16 @@ read_cmdstan_csv <- function(files,
     files <- file.path(temp_storage, basename(files))
   }
   format <- assert_valid_draws_format(format)
-  assert_file_exists(files, access = "r", extension = "csv")
+  csv_file_exists <- check_file_exists(files, access = "r", extension = "csv")
+  if (!isTRUE(csv_file_exists)) {
+    if (isTRUE(getOption("knitr.in.progress")) &&
+          is.null(getOption("cmdstanr_output_dir")) &&
```
jgabry, line 144: It occurs to me that this is checking the option value at read time but what matters is what the option value was at run time. 

If between the first render of the document and the later rerender if the user has since set `cmdstanr_output_dir` then this message will be suppressed. 

I have an idea for how we can handle this. I will make some commits when I have time.

jgabry: I just opened PR #1178, which takes a different approach to this that avoids this particular problem. Let me know what you think. One issue is that my version solves this problem but is more complicated. If we can figure out how to make your version work then I would prefer that to my version.

avehtari:
Superseded by #1178

PR #1178, opened 2026-04-09
Informative error message if temp csv files not found by Quarto

### LINKED ISSUE #1012: Errors re-rendering quarto html with cached chunks and cmdstanr.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #1012

This an alternative to #1176 that takes a different approach to giving the user an informative message when quarto fails to find the csv files. 

#### Copyright and Licensing

Jonah Gabry

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -1432,11 +1432,29 @@ CmdStanMCMC <- R6::R6Class(
       if (!length(self$output_files(include_failed = FALSE))) {
         stop("No chains finished successfully. Unable to retrieve the draws.", call. = FALSE)
       }
-      csv_contents <- read_cmdstan_csv(
-        files = self$output_files(include_failed = FALSE),
-        variables = variables,
-        sampler_diagnostics = sampler_diagnostics,
-        format = format
+      csv_contents <- tryCatch(
+        read_cmdstan_csv(
+          files = self$output_files(include_failed = FALSE),
+          variables = variables,
+          sampler_diagnostics = sampler_diagnostics,
+          format = format
+        ),
+        error = function(e) {
+          err_msg <- conditionMessage(e)
+          if (isTRUE(getOption("knitr.in.progress")) &&
+              isTRUE(self$runset$args$using_tempdir) &&
+              grepl("File does not exist:", err_msg, fixed = TRUE)) {
+            stop(
+              paste0(
+                err_msg,
+                "\n  If this error happened when using Quarto or Rmarkdown caching,\n",
+                "  see `cmdstanr_output_dir` in `?cmdstanr_global_options`"
+              ),
+              call. = FALSE
+            )
+          }
+          stop(e)
+        }
       )
       private$inv_metric_ <- csv_contents$inv_metric
       private$metadata_ <- csv_contents$metadata
--- a/R/fit.R
+++ b/R/fit.R
@@ -95,7 +95,28 @@ CmdStanFit <- R6::R6Class(
     init_ = NULL,
     profiles_ = NULL,
     model_methods_env_ = NULL,
-    return_codes_ = NULL
+    return_codes_ = NULL,
+    read_cmdstan_csv_with_cache_hint_ = function(...) {
+      tryCatch(
+        read_cmdstan_csv(...),
+        error = function(e) {
+          err_msg <- conditionMessage(e)
+          if (isTRUE(getOption("knitr.in.progress")) &&
+              isTRUE(self$runset$args$using_tempdir) &&
+              grepl("File does not exist:", err_msg, fixed = TRUE)) {
+            stop(
+              paste0(
+                err_msg,
+                "\n  If this error happened during a cached Quarto or R Markdown render,\n",
+                "  see `cmdstanr_output_dir` in `?cmdstanr_global_options`"
+              ),
+              call. = FALSE
+            )
+          }
+          stop(e)
+        }
+      )
+    }
   )
 )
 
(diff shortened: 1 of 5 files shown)


R/fit.R
```diff
@@ -95,7 +95,28 @@ CmdStanFit <- R6::R6Class(
...
     model_methods_env_ = NULL,
-    return_codes_ = NULL
+    return_codes_ = NULL,
+    read_cmdstan_csv_with_cache_hint_ = function(...) {
+      tryCatch(
+        read_cmdstan_csv(...),
+        error = function(e) {
+          err_msg <- conditionMessage(e)
+          if (isTRUE(getOption("knitr.in.progress")) &&
+              isTRUE(self$runset$args$using_tempdir) &&
```
jgabry, line 109: If we do it here instead of inside `read_cmdstan_csv()` we can access the internal `using_tempdir` flag, which is more reliable than checking `getOption(cmdstan_output_dir)` because the option may have changed between the initial quarto caching and the later re-rendering (so it's not a 100% reliable indicator of whether the temp directory was used).

jgabry:
@avehtari This PR is more complicated than yours because with this strategy we need to do it for every different type of fitted model object (mcmc, mle, pathfinder, etc.). In your version we could just do it in `read_cmdstan_csv()`, which is much simpler but I couldn't figure out how to solve [this problem](https://github.com/stan-dev/cmdstanr/pull/1176#discussion_r3060219907) without using an approach like this. I guess we could use your initial strategy of checking for evidence of tempdir in the path, but we'd need to make sure it works on Mac, windows, linux, WSL, and I wasn't sure of a reliable way to do that (but maybe there is). I would prefer to merge a simpler PR like yours, so maybe we can figure something out.

avehtari:
This PR doesn't do anything for me
```
[31m

processing file: quarto_test.R
[39m1/7                  
2/7 [unnamed-chunk-1]
3/7 [mod1]           
4/7 [fit1]           
5/7 [unnamed-chunk-2]
[31mError in `read_cmdstan_csv()`:
! Assertion on 'files' failed: File does not exist: '/tmp/RtmpNEy1UM/bioassay1-202604082047-1-769fdd.csv'.
Backtrace:
     ▆
  1. ├─base::withVisible(knit_print(x, ...))
  2. ├─knitr::knit_print(x, ...)
  3. └─knitr:::knit_print.default(x, ...)
  4.   └─knitr::normal_print(x)
  5.     ├─base::print(x)
  6.     └─R6:::print.R6(x)
  7.       └─.subset2(x, "print")(...)
  8.         └─self$summary(variables_to_print, ...)
  9.           └─self$draws(variables)
 10.             └─private$read_csv_(...)
 11.               └─cmdstanr::read_cmdstan_csv(...)
 12.                 └─cmdstanr:::assert_file_exists(files, access = "r", extension = "csv") at cmdstanr/R/csv.R:141:3
 13.                   └─checkmate::makeAssertion(files, res, .var.name, add)
 14.                     └─checkmate:::mstop(...)
[39m[31m
Quitting from quarto_test.R:41-45 [unnamed-chunk-2]
[39m[31mExecution halted
[39m[33mWARN: Error encountered when rendering files[39m
```

jgabry:
Weird, it worked when I tried it. But I did make a few edits after testing it so maybe I broke it. I’ll investigate.

jgabry:
@avehtari Was your fitted model object created using the cmdstanr from this branch or did you already have it from before? It needs to be created using this version of cmdstanr on this branch because of the way R6 objects work

avehtari:
I did switch to this branch, rendered without cache which should then crearte fit object with this branch, and then rendered with cache

jgabry:
Ok interesting, I’ll try to figure it out later when I have some time. Maybe I did something wrong when testing it myself yesterday.

avehtari:
My mistake, I remembered wrong when cache gets cleared, and when I did run with cache=false and then with cache=true, the cache was still having the old version. After clearing the cache, this PR works. So this PR works only after the fit object in the cache has been created with this PR, while my PR works also if the cache has been created with earlier version.

jgabry:
> So this PR works only after the fit object in the cache has been created with this PR, while my PR works also if the cache has been created with earlier version.

yes, however your PR checks cmdstanr_output_dir at re-render time but what we actually want to know is what its value was when the model was originally fit (_probably_ the same but not necessarily). So there are downsides to both PRs. Actually, the first version of your PR, which checked for evidence of the temporary directory in the path, solves the problem with your current PR and the problem with my PR, but we would need a way to reliably detect temp paths on all versions of all operating systems we support.

avehtari (approved):
Let's go with your version

PR #1197, opened 2026-06-15
Support model methods & standalone functions for models with external C++ (user_header)

ssp3nc3r:
#### Submission Checklist

- [x] Run unit tests

#### Summary

`compile_model_methods = TRUE` (the `$log_prob()` / `$grad_log_prob()` / `$hessian()` methods) and `compile_standalone = TRUE` (`expose_functions()`) currently fail to **compile** for any model that uses an external C++ `user_header` — even though the very same model samples without issue. This blocks a common workflow: validating a hand-written C++ log density / analytic gradient against the pure-Stan autodiff version via `grad_log_prob()`.

##### Reproducible example

`bernoulli_external.stan` (declares `make_odds`, defined in the header):

```stan
functions { real make_odds(real theta); }
data { int<lower=0> N; array[N] int<lower=0,upper=1> y; }
parameters { real<lower=0,upper=1> theta; }
model { theta ~ beta(1, 1); y ~ bernoulli(theta); }
generated quantities { real odds = make_odds(theta); }
```

`make_odds.hpp` (in the model's namespace, per the CmdStan convention):

```cpp
#include <stan/math.hpp>
#include <ostream>
namespace bernoulli_external_model_namespace {
  template <typename T0__, stan::require_all_t<stan::is_stan_scalar<T0__>>* = nullptr>
  stan::promote_args_t<T0__> make_odds(const T0__& theta, std::ostream* pstream__) {
    return theta / (1 - theta);
  }
}
```

```r
mod <- cmdstan_model("bernoulli_external.stan",
                     user_header = "make_odds.hpp",
                     compile_model_methods = TRUE)
```

`mod$sample()` works, but compiling the model methods aborts with:

```
error: use of undeclared identifier 'make_odds'
```

##### Root cause

Two independent problems in the standalone / model-methods build path:

1. **Mangled namespace.** The standalone & model-methods C++ is generated by calling `stanc` directly through an argument vector (no shell) in `get_standalone_hpp()`, reusing the same `--name='foo_model'` flag that is built for the *make-based* executable build. The make path is processed by a shell that strips the quotes; the direct call is not, so the literal quotes become part of the model name, producing the namespace `'foo_model'_namespace` (`x39model...x39_namespace` after C++ name-mangling). External C++ functions are defined by the user in `foo_model_namespace`, so they are not visible to the generated code.

2. **Header not included.** `rcpp_source_stan()` does not `-include` the user header when compiling the methods via `Rcpp::sourceCpp()`, unlike CmdStan's makefile, which adds `-include $(USER_HEADER)`.

##### Fix

- Build an *unquoted* variant of the stanc options for the two direct `get_standalone_hpp()` calls; the make path keeps the quoted form (so names/paths with spaces are still safe there).
- Stash the user-header path on the model-methods env (and on `self$functions`) and force-include it in `rcpp_source_stan()`, mirroring CmdStan's `-include $(USER_HEADER)`.

This also enables `compile_standalone = TRUE` / `expose_functions()` for models with external C++, which share `rcpp_source_stan()`.

##### Verification

For a normal model and an identical model whose single likelihood term is routed through an external C++ function, with the fix `log_prob()` **and** `grad_log_prob()` match the pure-Stan autodiff version exactly:

```
PLAIN     log_prob=-51.45881541  grad=(22.628038, 70.420847)
EXTERNAL  log_prob=-51.45881541  grad=(22.628038, 70.420847)
max|lp diff| = 0   max|grad diff| = 0
```

Adds a regression test in `tests/testthat/test-model-methods.R` using the existing `bernoulli_external` fixture (full `test-model-methods.R` passes locally on macOS / CmdStan 2.39.0).

#### Copyright and Licensing

Scott Spencer

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # cmdstanr (development version)
 
+* `compile_model_methods = TRUE` and `compile_standalone = TRUE` now work for
+models that use an external C++ `user_header`. The standalone / model-methods
+code is generated in the same namespace as the rest of the model (rather than a
+mangled `'foo_model'_namespace`), and the user header is force-included when
+those translation units are compiled, mirroring CmdStan's
+`-include $(USER_HEADER)`. (#1197)
 * `pathfinder()` now respects `save_single_paths = TRUE` instead of always
 passing `0` to CmdStan.
 * `pathfinder()` now uses `threads` argument (`num_threads` is deprecated),
--- a/R/utils.R
+++ b/R/utils.R
@@ -800,6 +800,13 @@ rcpp_source_stan <- function(code, env, verbose = FALSE, ...) {
   cppflags <- get_cmdstan_flags("CPPFLAGS")
   cmdstanr_includes <- system.file("include", package = "cmdstanr", mustWork = TRUE)
   cmdstanr_includes <- paste0(" -I\"", cmdstanr_includes,"\"")
+  # When the model uses an external C++ header, force-include it (as CmdStan's
+  # makefile does with -include $(USER_HEADER)) so that functions declared in the
+  # Stan program and defined in the header are available here too. (#1197)
+  user_header_include <- ""
+  if (!is.null(env$user_header_)) {
+    user_header_include <- paste0(" -include \"", env$user_header_, "\"")
+  }
   libs <- c("LDLIBS", "LIBSUNDIALS", "TBB_TARGETS", "LDFLAGS_TBB", "SUNDIALS_TARGETS")
   libs <- paste(sapply(libs, get_cmdstan_flags), collapse = " ")
   if (.Platform$OS.type == "windows") {
@@ -810,7 +817,7 @@ rcpp_source_stan <- function(code, env, verbose = FALSE, ...) {
       c(
         USE_CXX14 = 1,
         PKG_CPPFLAGS = cppflags,
-        PKG_CXXFLAGS = paste0(cxxflags, cmdstanr_includes, collapse = " "),
+        PKG_CXXFLAGS = paste0(cxxflags, cmdstanr_includes, user_header_include, collapse = " "),
         PKG_LIBS = libs
       ),
       Rcpp::sourceCpp(code = code, env = env, verbose = verbose, ...)
--- a/R/model.R
+++ b/R/model.R
@@ -616,27 +616,47 @@ compile <- function(quiet = TRUE,
   if (is.null(stanc_options[["name"]])) {
     stanc_options[["name"]] <- paste0(self$model_name(), "_model")
   }
+  # Two forms of the stanc options are needed. The quoted form is embedded in
+  # the STANCFLAGS string passed to make(), where it is processed by a shell and
+  # the quotes guard against spaces. The unquoted form is for get_standalone_hpp(),
+  # which calls stanc directly (no shell) via an argument vector, so quotes would
+  # otherwise become literal characters in option values - e.g. --name='foo_model'
+  # would produce the namespace 'foo_model'_namespace and hide external C++
+  # functions from the standalone / model-methods code. (#1197)
   stanc_built_options <- c()
+  stanc_built_options_unquoted <- c()
   for (i in seq_len(length(stanc_options))) {
     option_name <- names(stanc_options)[i]
     if (isTRUE(as.logical(stanc_options[[i]]))) {
-      stanc_built_options <- c(stanc_built_options, paste0("--", option_name))
+      opt <- paste0("--", option_name)
+      stanc_built_options <- c(stanc_built_options, opt)
+      stanc_built_options_unquoted <- c(stanc_built_options_unquoted, opt)
     } else if (is.null(option_name) || !nzchar(option_name)) {
-      stanc_built_options <- c(stanc_built_options, paste0("--", stanc_options[[i]]))
+      opt <- paste0("--", stanc_options[[i]])
+      stanc_built_options <- c(stanc_built_options, opt)
+      stanc_built_options_unquoted <- c(stanc_built_options_unquoted, opt)
     } else {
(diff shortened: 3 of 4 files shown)


R/utils.R
jgabry, line 852: `user_header_` will be dropped here, is that desirable? the PR  stores the header, but saved/reloaded models keep only `hpp_code_`. 

To see the potential issue: in a fresh R process save a CmdStanModel with `compile_model_methods = TRUE` and `user_header`. Then reload it, fit, then
call `fit$log_prob()`. There's an error about undeclared identifier 'make_odds'.


Should we preserve `user_header_` in `drop_stale_model_methods()`? Is that the desired behavior? If so, we should probably also add a regression test for saved/reloaded external-header models.

jgabry, line 1028: Here I think we might lose `user_header_` when `global=TRUE`. `rcpp_source_stan()` reads `env$user_header_`, but when `global=TRUE` we pass `globalenv()` instead of the function env that has the header.

I haven't reproduced this myself, but this should do it: 

In a fresh R session with an external standalone function, try `mod$expose_functions()` (should work)  and then try `mod$expose_functions(global = TRUE)` (probably fails with no function definition in the namespace).

To fix this I think we'd either need to (a) separate the target environment from the metadata/header source, or (b) compile with the model function env and copy exports to global afterward. Option (b) seems like a cleaner approach, so I'd lean towards that unless you have any other ideas.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # cmdstanr (development version)
 
+* `compile_model_methods = TRUE` and `compile_standalone = TRUE` now work for
+models that use an external C++ `user_header`. The standalone / model-methods
```
jgabry, line 4: I think we need to test `expose_functions` (with `global = FALSE` and `global = TRUE`) to be able to fully claim `compile_standalone = TRUE` / `expose_functions()` support

jgabry (commented):
Thanks for the PR! Here are some initial review comments after a first quick look. 

Also, was any of this written by AI? Either way is fine, I'm just asking because we recently adopted a new [AI policy](https://github.com/stan-dev/stan/wiki/AI-Contribution-Policy). The policy allows using AI tools but "if AI tools were used, the contributor additionally affirms that they have reviewed and understood the code, and can explain and defend their changes during the review process." Eventually we'll have some more info about this that comes up when opening a PR.

ssp3nc3r:
Thanks for the review help. 

I wrote it but I'm sure I used AI to figure out stuff. "if AI tools were used, the contributor additionally affirms that they have reviewed and understood the code, and can explain and defend their changes during the review process." I affirm.

On the other points, I hadn't thought about them because I just wrote it for my own pain point, which was just comparing the forward and reverse for two models (pure Stan's AD vs hand-coded C++, hence the user header) to check the gradient work before using it, if that makes sense. 

I've only been using it in that isolated case.

I'll need to think through the above.

jgabry:
@andrjohns Do you by any chance have time to take a look at this at some point? I made a few review comments but you're obviously much more familiar with the model methods and expose functions code.

andrjohns:
> @andrjohns Do you by any chance have time to take a look at this at some point? I made a few review comments but you're obviously much more familiar with the model methods and expose functions code.

Yep, happy to take this one!

jgabry:
> > @andrjohns Do you by any chance have time to take a look at this at some point? I made a few review comments but you're obviously much more familiar with the model methods and expose functions code.
> 
> Yep, happy to take this one!

Awesome, thanks!

andrjohns:
@ssp3nc3r thanks for proposing these changes, it's definitely functionality that's been missing!

We can implement this functionality with a simpler changeset. Because the model `.hpp` is stored at compile time, we can just store the external `.hpp` at the same time, and then simply prepend it to the code which is passed to `Rcpp` for the model methods and standalone functions.

In order to handle the namespacing issue when exporting the standalone function, we just need to remove the namespace from the function call.

So when the generated `--standalone-functions` code would have:
```cpp
double make_odds(const double& theta) {   return bernoulli_model_namespace::make_odds(theta, &Rcpp::Rcout); }
```

Given that we're prepending the external header, we just need to remove the `bernoulli_model_namespace` spec:
```cpp
double make_odds(const double& theta) {   return ::make_odds(theta, &Rcpp::Rcout); }
```

I've got a branch with these changes here: https://github.com/stan-dev/cmdstanr/compare/master...user-header-support

All works fine for me with the [manual's external c++ example](https://mc-stan.org/docs/cmdstan-guide/external_code.html), but can you try it out with your usecases and see if all works as expected for you?

jgabry:
> I've got a branch with these changes here: [master...user-header-support](https://github.com/stan-dev/cmdstanr/compare/master...user-header-support)

@ssp3nc3r were you able to try out this branch that @andrjohns made to see if it works for you?

ssp3nc3r:
Apologies for losing track of the conversation; I'll try this out tomorrow!

jgabry:
No worries, thanks!

ssp3nc3r:
Thanks for working on this it will be very useful! Sorry it took a bit to work this in as I didn't want to disturb mine in active use.

I tried it out and a couple of things. 

First, relative #includes in the user header break. You can see this in your working example: move make_odds's body into a make_odds_impl.hpp and have make_odds.hpp just #include "make_odds_impl.hpp" — then compile_model_methods/compile_standalone fails with fatal error: 'make_odds_impl.hpp' file not found. 

Prepending the header's text to the Rcpp compile drops its directory, so the relative include resolves against Rcpp's temp dir instead of the header's folder. Force-including the file (-include <path>, like #1197 does) or adding dirname(user_header) to the Rcpp include path may fix it. This one affected me because my real headers are split across component files.

Second, the namespace-strip keys off whether the header text contains namespace { so a header that namespaces its internal helpers but exposes the entry-point function at global scope skips the strip and won't resolve. Keying off whether each exposed name is actually reachable at global scope I think would be preferred.

PR #1208, opened 2026-07-21
Fix pathfinder initialization duplicate draws edge cases

### LINKED ISSUE #1207: Edge case problems with using pathfinder for inits
Even after #1206 that are still some more problems with using Pathfinder to initialize. These are mostly edge cases so shouldn't affect most users. 

`process_init_approx()` currently uses the number of unique log weights (`length(unique(draws_df$lw))`) as a proxy for the number of distinct draws. However I think that means it's possible for different parameter vectors to have the same log weight and be incorrectly treated as one. Also duplicate PSIS rows remain in the data, so sampling can still pick identical init vectors (even though we claim to have a “distinct draws” check). 

The aggregate()/merge() step in the code also doesn't actually remove duplicate rows like it seems to try to! It seems like that was intended to deduplicate the draws but the way it's currently implemented keeps the duplicate rows. That said, we shouldn't just remove duplicates because for PSIS-resampled Pathfinder output this should affect resampling frequency. We need some way to  collapse duplicate parameter vectors while retaining the information we need for selection weights.

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Fixes #1207

AI disclaimer: Claude was used to write the three line summary below and to generate the tests included in this PR. I also had it check the code I wrote for mistakes and suggest efficiency improvements. 

- Determine distinct initialization candidates using exact values of model parameters instead of unique log weights.
- Collapse duplicate initialization vectors and sum their selection weights, preserving duplicate frequency in PSIS-resampled Pathfinder output.
- Add `vctrs` as a direct dependency and add regression coverage for equal log weights, unused variables, and duplicate candidates. (This doesn't really add a dependency because it's already a dependency of `posterior`)

#### Copyright and Licensing

Jonah Gabry


avehtari:
Looks good. I had assumed the probability of this edge case to be very small, but apparently it did hit someone

jgabry:
Ok thanks. I'll merge if all the checks pass.

PR #1211, opened 2026-07-22
Simplify the lookup process for windows toolchains

andrjohns:
#### Submission Checklist

- [x] Run unit tests

#### Summary

The current process for the toolchain on Windows is a little brittle, and requires hard-coding a new conditional on each RTools release. Starting from R4.2 the `R_TOOLS_SOFT` config variable was introduced to provide the path to the toolchain directory for the current installation - also allowing alternate packagers to specify a different default, which we can query directly. 

The current PR updates to toolchain resolution to first check for a valid toolchain in the `R_TOOLS_SOFT` directory. If one is not found, it falls back to searching the PATH (via `Sys.which()`) as R automatically prepends the toolchain directories to the PATH on startup.

For R4.0 (before `R_TOOLS_SOFT` existed or the toolchain was automatically added to the PATH), it uses the original lookup of the `RTOOLS40_HOME` environment variable

#### Copyright and Licensing

Andrew Johnson


R/install.R
```diff
@@ -751,104 +708,42 @@ toolchain_PATH_env_var <- function() {
...
-    return("")
-  }
-  # Return the first usable candidate (ordered by preference above).
-  usable <- vapply(candidates, is_rtools4x_toolchain_usable, logical(1))
-  if (any(usable)) {
-    return(candidates[which(usable)[1]])
+  # R 4.0 did not include the toolchain in the PATH by default
+  # or set the R_TOOLS_SOFT or LOCAL_SOFT config variables, so
+  # we use the RTOOLS40_HOME environment variable instead
+  if (R.version$major == 4 && R.version$minor < 2) {
```
jgabry:
```suggestion
  if (current_r_version() < "4.2.0") {
```

Not a problem anytime soon, but because R.version$major actually returns a character we can get weird things like "10.0" < 2 evaluates to TRUE, so R 4.10 would incorrectly use Rtools40.  `current_r_version()` returns `getRversion()` and has class `numeric_version`, which avoids this.

jgabry (changes requested):
Cool thanks for doing this! I just saw the PR and happened to have a sec so I took a quick glance (I'll wait to look more thoroughly until you're ready) and I noticed one thing that has bitten me before so I made a suggested change

R/install.R
andrjohns: Ah that is classic R. Thanks for catching!

jgabry: I wonder if we could just require R >= 4.2? Or do you think there are still enough older R versions in the wild that we should continue supporting it?

andrjohns: Yeah I think there might still be a bunch of older systems out there - where I work in gov we only left 4.2 a few months ago 😮‍💨 

Pretty happy to only do the bare minimum for it though!

jgabry: Yeah good point. I would guess some companies  still use old R versions too, so we should keep supporting it

andrjohns:
@jgabry this one's good to look at! CI is all green for standard RTools and I've also verified locally that we now support the [MSYS-packaged R](https://packages.msys2.org/base/mingw-w64-r) too

jgabry:
Cool, thanks @andrjohns! I'll take a look

R/install.R
```diff
@@ -751,104 +708,38 @@ toolchain_PATH_env_var <- function() {
...
-  if (!nzchar(path) || !dir.exists(path)) {
-    return(FALSE)
+  # R 4.0 did not set the R_TOOLS_SOFT config variable, so
+  # we use the RTOOLS40_HOME environment variable instead
+  if (current_r_version() < "4.2.0") {
+    rtools40_home <- repair_path(Sys.getenv("RTOOLS40_HOME", "C:\\rtools40"))
+    r_arch <- ifelse(Sys.getenv("R_ARCH") == "/i386", "mingw32", "mingw64")
+    rtools_soft <- file.path(rtools40_home, r_arch)
+  } else {
+    rtools_soft <- repair_path(tools::Rcmd(c("config", "R_TOOLS_SOFT"), stdout = TRUE))
```
jgabry: This will start an R CMD config subprocess every time `toolchain_PATH_env_var()` is called, right? Since that gets called in a ton of places in the package (every model fit, every CSV read, cpp_opts, etc.), should we compute it once and cache it somewhere (maybe an environment internal to the package)? I'm assuming it can't change within a single session. 

That said, maybe it's not worth bothering to do this it if it's pretty fast. Could it add up enough to be noticeable? I defer to your judgement on this, since I'm not a windows users.

jgabry: Also, what happens if `R_TOOLS_SOFT` isn't in Makeconf? Maybe that's not possible with CRAN R >= 4.2 but maybe it's possible for something like conda R? 

I tried testing it on my Mac, which kind of simulates a windows distribution that doesn't have it, and R CMD config has a nonzero exit status and system2 throws a "running command ... had status 1" warning on every call to toolchain_PATH_env_var(). 

What if we do something like this: 

```r
out <- suppressWarnings(
  tools::Rcmd(c("config", "R_TOOLS_SOFT"), stdout = TRUE)
)
if (is.null(attr(out, "status")) && length(out) == 1) {
  rtools_soft <- repair_path(out) 
} else {
  rtools_soft <- ""
}
```

R/install.R
```diff
@@ -614,54 +614,11 @@ check_wsl_toolchain <- function() {
...
-      "\nPlease install or reinstall the appropriate Rtools version for this R installation,",
-      "\nrestart R, and then run cmdstanr::check_cmdstan_toolchain().",
-      call. = FALSE
-    )
-  }
-  # If RTools is installed in a path with spaces or brackets
-  # we error as this path is not valid
-  if (grepl("\\(|)| ", rtools_path)) {
-    stop(
-      "\n", rtools_version, " is installed in a path with spaces or brackets, which is not supported.",
```
jgabry, line 636: Should we keep something like this to make sure we error early when rtools is in a directory with spaces? Seems somewhat common

tests/testthat/test-install.R
```diff
@@ -304,215 +304,6 @@ test_that("deprecated CMDSTANR_USE_MSYS_TOOLCHAIN is ignored with warning", {
...
-      rtools4x_toolchain_candidates = function() character()
-    )
-    expect_error(
-      check_rtools4x_windows_toolchain(),
-      "restart R, and then run cmdstanr::check_cmdstan_toolchain()",
-      fixed = TRUE
-    )
-  })
-})
-
```
jgabry, line 515: We're deleting hundreds of lines of tests, which is great, but is there anything we should be adding to the tests for the new logic? Makes me slightly nervous to not have anything. Or do you think our other existing tests cover everything we need?

jgabry (commented):
This is great, thanks for doing this. I did a little research to try to understand `R_TOOLS_SOFT` and I tested a couple things. I have a couple of questions/concerns, but not sure how big a deal they are (see review comments).

jgabry:
Hey @andrjohns, just checking in on the status of this. Have you had a chance to take a look at my comments? Not insanely urgent (I'm still working through my v1.0 to-do list), but would be good to get this finished and merged when you have time. Thanks!

andrjohns:
I've updated the handling to cache the path lookup in the central `.cmdstanr` variable, added handling for missing `R_TOOLS_SOFT`, and added a bunch of tests & test coverage back - let me know if anything extra would be good!

andrjohns:
Almost forgot! I also fixed the handling for rtools installations with spaces in the path and added that to the CI workflow

jgabry (commented):
I have one question about c++ vs g++. What's the reason for checking if c++.exe exists instead of g++.exe? Same with Sys.which("c++") instead of g++. Previously we were checking g++ and CmdStan defaults to g++ on Windows, right? I guess standard installations would have both, but I'm just curious if your change was deliberate for a particular reason. 


Otherwise, I hope you don't mind, I pushed a couple of smallish changes  mostly to handle a few edge cases, not changing your approach here. 

One thing I added is a check in R-CMD-check.yaml to make sure it's actually using R_TOOLS_SOFT and not just using the fallback (claude suggested this and it seemed reasonable). I don't think we otherwise really had a way of knowing this. But we can remove it if I'm wrong about that. 

The other code changes were made by me, except for the tests, which I had Claude write (they seemed like reasonable additions to me, except for a few unnecessary ones that I removed before pushing). 

Aside from that question about c++ vs g++ (and making sure CI passes with my latest changes), I think this is ready to go!

andrjohns:
> I have one question about c++ vs g++. What's the reason for checking if c++.exe exists instead of g++.exe? Same with Sys.which("c++") instead of g++.

That's to help with non-standard R installations and Windows ARM. On Windows ARM64 the compiler is `clang` with `g++.exe` just being a shortcut/link, so if they remove/change the link we'll fail to find the toolchain. Non-CRAN installations or custom toolchains don't always use `gcc` on windows (e.g., MSYS) so this adds support for free

andrjohns:
The failure you were seeing in the workflow for Windows ARM toolchain `R_TOOLS_SOFT` check is because of a quirk/bug in the environment variables being set for the runner, I've logged an issue upstream: https://github.com/r-lib/actions/issues/1102

jgabry:
> That's to help with non-standard R installations and Windows ARM. On Windows ARM64 the compiler is `clang` with `g++.exe` just being a shortcut/link, so if they remove/change the link we'll fail to find the toolchain. Non-CRAN installations or custom toolchains don't always use `gcc` on windows (e.g., MSYS) so this adds support for free

Ok cool, sounds good to me. 

> The failure you were seeing in the workflow for Windows ARM toolchain `R_TOOLS_SOFT` check is because of a quirk/bug in the environment variables being set for the runner, I've logged an issue upstream: [r-lib/actions#1102](https://github.com/r-lib/actions/issues/1102)

Thanks for opening the issue. I guess we can reenable the `R_TOOLS_SOFT` check for windows ARM if they fix that. For now we can leave it bypassed. 

I think we can go ahead and merge this!

jgabry (approved):

PR #1217, opened 2026-07-23
Read compressed CSVs

### LINKED ISSUE #1027: Support compression in read_cmdstan_csv
Cmdstan output files can get pretty large. `data.table::fread` used in `read_cmdstan_csv` and `read_csv_metadata` supports gzip and bzip2 compressed files if the R.utils package is installed.

On the `cmdstanr` side this would require at least modifying the suffix assertions (`assert_file_exists`) in both functions. And, switching out `grep` in favor of  `zgrep`  to handle the commented headers with the caveat (at least on linux) that the argument order be switched so that `--color=never` comes before the regex pattern.

So in `read_cmdstan_csv`:

```R
fread_cmd <- paste0("zgrep --color=never -v '^#' '", path.expand(output_file), "'")
```

`read_metadata_csv`:

```R
fread_cmd <- paste0("zgrep --color=never '^[#a-zA-Z]' '", path.expand(csv_file), "'")
```
and similarly for the WSL specific code.

VisruthSK:
This PR was generated by Codex.

#### Submission Checklist

- [x] Run unit tests

#### Summary

Support compressed CSVs. Closes #1027.

#### Copyright and Licensing

Visruth Srimath Kandali

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -13,3 +13,4 @@
 ^vignettes/articles-online-only$
 ^release-prep\.R$
 ^\.vscode$
+^dev-notes$
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -33,18 +33,18 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: macOS-latest, r: 'devel', opencl: true}
+          - {os: macOS-latest, r: 'devel', opencl: false}
           - {os: macOS-latest, r: 'release', opencl: true}
           - {os: macos-26, r: 'release', opencl: false}
           - {os: macos-15-intel, r: 'release', opencl: false}
-          - {os: windows-latest, r: 'devel' }
-          - {os: windows-latest, r: 'release'}
-          - {os: windows-2022, r: '4.1'}
-          - {os: windows-latest, r: 'oldrel'}
-          - {os: ubuntu-latest, r: 'devel', opencl: true}
+          - {os: windows-latest, r: 'devel', opencl: false}
+          - {os: windows-latest, r: 'release', opencl: false}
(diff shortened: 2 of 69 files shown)


jgabry:
Thanks for keeping this up to date. I haven't had a chance to seriously look at this yet, although I'm still interested in it. Trying to make a push on the compilation/c++ options stuff first

jgabry (changes requested):
This looks quite good. I committed a couple of small improvements and the missing doc. But there's one bigger issue we should probably deal with, even though it's probably uncommon. I asked codex about it: 

> With the current `gzip/bzip2 -dc | grep` a truncated or damaged archive can produce part of the CSV before the decompressor fails. grep can still exit successfully, so fread() treats the command as successful and read_cmdstan_csv() returns only the rows it received. I reproduced this with both gzip and bzip2.

Maybe we should run gzip -t or bzip2 -t once per compressed file before reading it? And if that fails then we throw an informative error message (and leave the file unmodified)? 

That said, this would add a fourth decompression (metadata, sampler diagnostics, variables, and now this check), which is annoying. But we could probably keep it to three by reading all the requested diagnostic and variable columns once. Something like this (although I haven't checked if there are subtleties I'm overlooking):

```r
selected <- c(sampler_diagnostics, variables)
csv_data <- data.table::fread(cmd = fread_cmd, select = selected, ...)

diagnostics <- csv_data[sampler_diagnostics]
draws <- csv_data[variables]
```

What do you think?

Also, after we merge this we could extend to #1276, which would make this much more convenient. But that should be done separately I think

VisruthSK:
Agree, that approach seems good too to collapse to three reads. I think it makes sense to stack 1276 with this so they are merged together? Or to do 1276 in this PR too.

jgabry:
I think I'd prefer stacking rather than combining, but not super strongly, so whatever you prefer is fine

PR #1254, opened 2026-08-27
Design note: v1.0 compilation state and C++ options

jgabry:
Adds `dev-notes/compilation-state.md`. No code changes. 

This is the current plan for v1.0, developed based on conversation with @SteveBronder and @WardBrian.

AI disclosure: the compilation-state.md file is being written iteratively via a back and forth with Claude and Codex. It is based on the list below, which is my own summary of the planned changes. *This is somewhat of an experiment, and if it goes poorly I may end up writing the document myself from scratch.*

---

### New API

- `cmdstan_model()` checks if the existing executable matches the requested Stan file (and includes and user headers) and options (see section on new build record below). If everything matches we reuse the executable, otherwise we recompile. 
- `cmdstan_model(exe_file = )` stays. A pre-built executable still works, but `$code()`, `$variables()`, `$check_syntax()` and `$format()` need a Stan file, and passing build options (`cpp_options`, `stanc_options`, `include_paths`, `user_header`, `force_recompile`, `pedantic`) with only an executable is an error, since there's nothing to rebuild. 
- `cmdstan_model(stan_file = , exe_file = )` together is now an error. That usage of `exe_file` was just used to indicate where to put it, and `dir` does that anyway (albeit without filename customization).
- deferred compilation goes away!! (remove the `compile = FALSE` argument to `cmdstan_model` and the `$compile()` method). This means we lose `compile` arguments like `compile_model_methods` and `compile_standalone`. But `$expose_functions()` and `$init_model_methods()` already do the same job.  
  - New standalone functions replace methods that could be called pre-compilation. This avoids having a CmdStanModel object where only a small subset of methods are usable (generally considered poor design). 
      - `format_stan_file()`
      - `check_syntax_stan_file()`
      - `stan_variables()`
      - `compile_stan_file()`
      - `stan_build_info()` for inspecting how an executable was built
- `$code()` and `$variables()` refer to the Stan file used to build the executable, even if the Stan file has changed since (needs to be recompiled)
- any method that uses the executable checks that it's still up to date and throws an error telling the user to recompile (the methods themselves don't force recompilation). The error tells them to call `cmdstan_model()` again, not `force_recompile = TRUE` since `cmdstan_model()` will now know what to do 
- unnamed `cpp_options` are rejected, e.g. `list("STAN_THREADS=TRUE")` has to be `list(stan_threads = TRUE)`
- `user_header` is only settable through the `user_header` argument (not `cpp_options`) and a new method `$user_header()` is added to read it back.

### When do we recompile

If the user sets `force_recompile = TRUE` or when any of these change:

- the Stan program
- an include (or which file the include actually resolves to)
- `include_paths` (when the change means different content, not e.g. a directory rename) 
- the user header (or you point `cmdstan_model()` at a different one) or its path
- make/local
- the `cpp_options` or `stanc_options` the user supplied (options cmdstanr fills in itself are recorded but don't trigger a rebuild, except the model name, which we get from the file name and we do compare, so renaming your .stan file recompiles)
- the CmdStan installation
- the build record was written in a format version this cmdstanr doesn't read (we only read the format we write)

Or when we can't trust what we recorded:

- the executable isn't the one the build record describes (someone replaced it, or it's corrupt)
- the build record is missing or unreadable
- the executable is old enough that it doesn't have one

If more than one of these applies we report all of them (if possible).

One exception:

- executable-only models (`cmdstan_model(exe_file = )`) can't recompile automatically since there's no Stan file to build from

### The new build record

The current plan is a file `.<exe>.cmdstanr.json` that is written next to the executable. It contains: 

  - the `cpp_options` and `stanc_options` the user supplied 
  - the stanc options cmdstanr adds itself (separately from the user's, because only the user's are compared, apart from the model name)
  - the model name cmdstanr derives from the file name 
  - `include paths` and user header path
  - what the executable actually reports (e.g. threading and OpenCL), and whether we could tell at all (not reported doesn't mean off)
  - content hashes of the stan file, includes and user header, plus the path each had at build time (paths are not actually compared to decide rebuilding except the user header's path, see above)
  - the ordered include list returned by `stanc --info`
  - make/local hash
  - enough info to identify cmdstan installation that created it
  - the TBB directory the build actually used (on Windows we have to put it on the PATH ourselves, only the build knows where it is)
  - a hash of the executable so we can connect the build record to the exact binary
  - anything we know we're not tracking (like a make/local that includes another makefile)
  - a version for how this info is interpreted 

----

Tracking issue:  #1258


jgabry:
@andrjohns when you have a chance take a peek at my list above (you don't need to read the full document in the PR that Claude, it's just more details on all the items that I wrote in that list above). This redesign of the compilation/build process for 1.0 came out of discussion with @WardBrian and @SteveBronder. I think it's a much cleaner design than what we currently do (and actually simpler in many ways, despite the additional build record) and it replaces the previous half-done C++ options work that never got finished. 

You can ignore all the issues that have been opened lately, they're just based off of this list and I'll close them as I go through the implementation. I'm hoping to start working on this ASAP.

jgabry:
@paul-buerkner this is going to affect how brms does cmdstanr compilation a little bit, but I will submit a brms PR to make it compatible once I'm at the right place in this work. I think ultimately this will be an improvement for brms too.

jgabry:
@andrjohns One consequence of removing the `compile` method as part of removing deferred compilation is also removing the `compile_standalone` and `compile_model_methods` arguments, unless we add them directly to `cmdstan_model()`. But I think that's ok and honestly cleaner to have just `$expose_stan_functions()` and `$init_model_methods()` rather than multiple different ways of getting those.

PR #1266, opened 2026-09-12
no duplicate make/local flags

avehtari:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Current `cmdstan_make_local()` copies `make/local` flags, but doesn't handle duplicates and using `cmdstan_make_local()` after every new release (and new release candidate) adds, for example, line `CXXFLAGS += -Wno-deprecated-declarations`. This PR copies the flags without creating duplicate lines.

This PR was made with Claude assistance

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -314,9 +314,8 @@ rebuild_cmdstan <- function(dir = cmdstan_path(),
 #' @export
 #' @param append (logical) For `cmdstan_make_local()`, should the listed
 #'   makefile flags be appended to the end of the existing `make/local` file?
-#'   The default is `TRUE`. If `FALSE` the file is overwritten. Appending is
-#'   idempotent: a flag that is already present in `make/local` is not written
-#'   a second time, and repeated flags in `cpp_options` are written once.
+#'   The default is `TRUE`. If `FALSE` the file is overwritten. A flag that is
+#'   already present in `make/local` is not written a second time.
 #'
 cmdstan_make_local <- function(dir = cmdstan_path(),
                                cpp_options = NULL,
@@ -338,18 +337,13 @@ cmdstan_make_local <- function(dir = cmdstan_path(),
         }
       }
     }
-    if (append) {
-      # Don't write a flag that is already in the file.
-      existing <- tryCatch(
-        suppressWarnings(readLines(make_local_path, warn = FALSE)),
-        error = function(e) character(0)
-      )
-      built_flags <- built_flags[!trimws(built_flags) %in% trimws(existing)]
+    if (append && file.exists(make_local_path)) {
+      # Don't write a flag that make would already apply anyway.
+      existing <- suppressWarnings(readLines(make_local_path, warn = FALSE))
+      built_flags <- built_flags[!make_flag_already_applies(built_flags, existing)]
     }
-    built_flags <- unique(built_flags)
-    # Skip the write if there is nothing left to add and the file is
-    # already there.
-    if (length(built_flags) > 0 || !append || !file.exists(make_local_path)) {
+    # Skip the write if there is nothing left to add.
+    if (length(built_flags) > 0 || !append) {
       write(built_flags, file = make_local_path, append = append)
     }
   }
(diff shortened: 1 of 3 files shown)


jgabry (changes requested):
Thanks Aki. I think we should do this (especially if doing #1267) but I think we can simplify it a little bit and avoid an edge case or two. 

Currently this checks against every line in the file, but make only uses the last assignment, so a value that's overridden lower down still counts as "already present":

```
    cmdstan_make_local(cpp_options = list(STAN_THREADS = TRUE))   # STAN_THREADS=true
    cmdstan_make_local(cpp_options = list(STAN_THREADS = FALSE))  # true, false
    cmdstan_make_local(cpp_options = list(STAN_THREADS = TRUE))   # still true, false
```

That third one should turn threading back on, but with this PR it won't, right? It's a bit of an edge case, but if the user actually has TRUE then FALSE then TRUE we should respect that probably? How about something like this rule, which is basically what make does: a line using `=` counts as already present only if it's still the last assignment to that variable, but `+=` lines count as present wherever they are in the file. 

Some other simplifications:
- I think we can drop the `unique()` on cpp_options and drop the test for it. I think it's probably overly defensive, if the user explicitly passes duplicates (very rare I assume) then no big deal 
- We can also drop the readLines tryCatch because just checking `file.exists()` is enough (and there's already that tryCatch below for reading the file back).
- Keep checking `length(build_flags)` and `!append` before `write()` but drop the `!file.exists()` check because it doesn't do anything (if the file is missing nothing was filtered out) 

I asked Claude to draft the simpler version I had in mind and it said roughly something like this (didn't check it thoroughly, so this is a rough draft): 

```r
    if (append && file.exists(make_local_path)) {
      existing <- trimws(readLines(make_local_path, warn = FALSE))
      # Make uses the last assignment to a variable, so a plain `=` line is
      # only already present if it is still the last one for its variable.
      # `+=` lines accumulate, so those count wherever they are.
      variable <- function(x) trimws(sub("[+:?]?=.*", "", x))
      last_line <- tapply(existing, variable(existing), function(x) x[length(x)])
      flags <- trimws(built_flags)
      present <- flags %in% existing &
        (grepl("+=", flags, fixed = TRUE) | flags == last_line[variable(flags)])
      built_flags <- built_flags[!present]
    }
    if (length(built_flags) > 0 || !append) {
      write(built_flags, file = make_local_path, append = append)
    }
```

For the tests, we can remove the "repeated flags in cpp_options" one and maybe add one like the edge case I mentioned above of not removing a flag set to TRUE if it comes in a sequence of TRUE then FALSE then TRUE. 

Also I don't think we should use the work `idempotent` in the doc. That whole sentence could just be "A flag that is already present in `make/local` is not written a second time." (I also dropped the part about repeat `cpp_options`, see comment above)

avehtari:
- The rule now follows make

```r
if (append && file.exists(make_local_path)) {
  existing <- suppressWarnings(readLines(make_local_path, warn = FALSE))
  built_flags <- built_flags[!make_flag_already_applies(built_flags, existing)]
}
if (length(built_flags) > 0 || !append) {
  write(built_flags, file = make_local_path, append = append)
}
```

`make_flag_already_applies()` (internal, next to `check_install_dir()`): `+=` counts as present if an identical line appears anywhere; a plain assignment counts only if it is still the last line touching that variable. "Touching" includes a later `+=` to the same variable, so `CXXFLAGS=-g` / `CXXFLAGS += -O2` / `CXXFLAGS=-g` writes the third line. Comments and blanks aren't assignments, so `# STAN_THREADS=true` doesn't mask a real one. A helper `make_assignment_part()` pulls the variable and operator out with one pattern covering `=`, `:=`, `?=`, `+=`.

- The example now behaves as:

```r
cmdstan_make_local(cpp_options = list(STAN_THREADS = TRUE))   # true
cmdstan_make_local(cpp_options = list(STAN_THREADS = FALSE))  # true, false
cmdstan_make_local(cpp_options = list(STAN_THREADS = TRUE))   # true, false, true
cmdstan_make_local(cpp_options = list(STAN_THREADS = TRUE))   # unchanged — now last again
```

- Code, message and doc simplified

jgabry (approved):
Thanks, looks good! I just pushed one small change so that when checking `+=` we only only look at lines after the last `=` or `:=` to the _same variable_, since that already overrides the earlier `+=` and we probably should leave the user's edits alone in that case. I think this is ready to go if CI passes

jgabry:
> Thanks, looks good! I just pushed one small change so that when checking `+=` we only only look at lines after the last `=` or `:=` to the _same variable_, since that already overrides the earlier `+=` and we probably should leave the user's edits alone in that case. I think this is ready to go if CI passes

Actually I just noticed a few edge cases we should probably handle. Will make another push shortly

jgabry:
It occurred to me that we need to be careful about lines continued with backslashes, so I asked codex to look into that and fix any edge cases. I think it's ok to merge now when CI passes.

PR #1267, opened 2026-09-12
ask to copy make/local before build

avehtari:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Current `install_cmdstan()` downloads CmdStan, builds, and then asks whether to copy `make/local` from an older existing version, and then CmdStan need be rebuild. This PR asks in an interactive session whether to copy `make/local` flags before building, and thus the second build is not needed. In case of reinstalling a new version, this will also allow copying current version `make/local` flags, which would be otherwise deleted.

This PR was made with Claude assistance

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -78,7 +78,9 @@
 #'   installation currently in use be copied to the new installation? The copy
 #'   happens before CmdStan is built, so the flags are already in effect for
 #'   that build. The default is `NULL`, which shows the previous `make/local`
-#'   and asks in an interactive session, and copies nothing otherwise.
+#'   and asks in an interactive session, and copies nothing otherwise. The
+#'   question comes before the download, so that the rest of the installation
+#'   runs unattended.
 #'   Use `TRUE` or `FALSE` to decide without being asked. Flags given in
 #'   `cpp_options` are written after the copied ones and therefore take precedence.
 #'
@@ -214,6 +216,12 @@ install_cmdstan <- function(dir = NULL,
   if (!check_install_dir(dir_cmdstan, overwrite)) {
     return(invisible(NULL))
   }
+  # Ask before downloading
+  copy_make_local <- resolve_copy_make_local(
+    previous_make_local,
+    old_cmdstan_path,
+    copy_make_local
+  )
   if (is.null(release_file)) {
     tar_downloaded <- download_with_retries(download_url, dest_file, quiet = quiet)
     if (inherits(tar_downloaded, "try-error")) {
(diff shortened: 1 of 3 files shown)


R/install.R
```diff
@@ -239,6 +261,14 @@ install_cmdstan <- function(dir = NULL,
     assert_supported_requested_cmdstan_version(extracted_version, source = "archive")
   }
 
+  # Carry the previous installation's makefile flags over before the build, so
+  # that the build already uses them. Written first, so that cpp_options and
+  # the platform flags below take precedence over an inherited assignment.
+  if (maybe_copy_make_local(dir_cmdstan, previous_make_local, old_cmdstan_path,
```
jgabry: I think we should prompt the user with this question earlier. Everything needed (`previous_make_local`, `old_cmdstan_path`) is already known up at line ~153 before any downloading happens. If we don't ask earlier then something like this could happen: someone starts the installation, they see "Downloading...", they switch to some other task to let cmdstanr download and install cmdstan, they come back a while later are find it stuck on waiting for a response instead of building cmdstan. If we ask them earlier (at least before `download_with_retries()`) then all the waiting users have to do happens in one place.

R/install.R
```diff
@@ -284,13 +314,15 @@ install_cmdstan <- function(dir = NULL,
 
   message("* Finished installing CmdStan to ", dir_cmdstan, "\n")
   set_cmdstan_path(dir_cmdstan)
-  if (!is.null(make_local_msg) && old_cmdstan_path != cmdstan_path()) {
+  if (!is.null(make_local_msg) && !identical(old_cmdstan_path, cmdstan_path())) {
```
jgabry: This only avoids the message when the copy actually happens, so if a user does `copy_make_local = FALSE` or answers `N` when prompted about copying make/local, they would still get the message about copying make/local afterwards, including the new line suggesting `install_cmdstan(copy_make_local = TRUE)`. But at that point following the advice would mean a full reinstall. I'm thinking that we should probably only use the message when no decision was made (the non-interactive `NULL` case), but I'm not 100% sure. What do you think?

R/install.R
```diff
@@ -386,6 +418,58 @@ check_cmdstan_toolchain <- function(fix = FALSE, quiet = FALSE) {
...
+#' @param previous_path (string) Where those contents came from.
+#' @param copy_make_local (logical or `NULL`) `TRUE`/`FALSE` decide directly.
+#'   `NULL` asks in an interactive session and declines otherwise, so that
+#'   scripts, R CMD check and CI never block on a prompt.
+#' @return `TRUE` if the flags were written to the new installation.
+maybe_copy_make_local <- function(dir_cmdstan,
+                                  previous_contents,
+                                  previous_path,
+                                  copy_make_local = NULL) {
+  if (length(previous_contents) == 0 || identical(previous_contents, "")) {
```
jgabry, line 477: If someone explicitly does `copy_make_local = TRUE` and there's nothing to copy (which I guess could by an empty or missing `make/local` or `cmdstan_path()` pointed at a different install than they thought), this returns `FALSE` silently. They won't find out that their make/local flags didn't copy over correctly until a compile later behaves differently. How about a message like "No make/local found in <old path>, nothing to copy"?

jgabry (changes requested):
Thanks Aki. The logic looks good to me overall. I made a few small suggestions in review comments. In addition to those comments, a couple other things: 

- Can you also add a NEWS entry? 
- I had wrote a whole thing about how we would get duplicate lines in make/local before seeing your other PR. Taking a look at that one too, since it should be done before this one.

R/install.R
avehtari: Now asking before download

avehtari:
- Question is not asked if explicit argument was given
- The post-build message no longer repeats an answered question

| `make/local` in previous install | question settled?         | install path            | message    |
|--------------------------------|---------------------------|-------------------------|------------|
| has flags                      | no (non-interactive `NULL`) | new                     | shown      |
| has flags                      | yes (argument or prompt)  | new                     | suppressed |
| no flags                       | —                         | new                     | suppressed |
| has flags                      | no                        | same (`overwrite = TRUE`) | suppressed |

- An explicit `TRUE` with nothing to copy is reported

jgabry (approved):
Looks good, thanks! I just pushed one commit moving the user prompt up even a tiny bit earlier. I think this is ready, but I'm going to wait to merge it until #1266 is merged because that one should go first.

jgabry:
#1266 is merged. I just fixed the merge conflicts and will merge this if CI passes.

jgabry:
The one failing ci run is unrelated, merging now. Thanks again Aki!

PR #1274, opened 2026-09-16
improve model methods performance

avehtari:
This PR and PR text were Claude assisted. I have checked the code and understand the logic of the change

## Summary

`$log_prob()`, `$grad_log_prob()`, `$hessian()`, `$constrain_variables()`, `$unconstrain_variables()`, `$unconstrain_draws()` and `$variable_skeleton()` all begin with `self$init_model_methods()`. When the methods are already initialised that call still cost ~54 us, which is 40-50x the Stan evaluation it guards for a small model. This adds an O(1) fast path so the initialised case costs 2.8 us, and removes two costs from it:

- the #1157 staleness predicate no longer round-trips through `capture.output(print(symbol))` (39.7 us);
- the initialised check no longer calls `ls()` on the model-methods environment (7.4 us — `ls()` lists and sorts every binding).

No API change, no change to when methods are compiled or rebuilt.

## Measurement

R 4.6.1, CmdStan 2.40.0, Linux, `parameters { vector[5] mu; } model { target += normal_lpdf(mu | 0, 1); }`. Minimum of 7 replicates of a 2000-iteration loop, `gc(FALSE)` before each, harness floor measured and subtracted. Both columns are the same tree with only these files reverted, so version, CmdStan and model are held fixed.

| call                                          |  before |   after | factor |
|-----------------------------------------------|--------:|--------:|-------:|
| `source_cpp_native_symbol_is_null(model_ptr)` | 39.70 us|  5.65 us|   7.0x |
| `drop_stale_model_methods(env)`               | 40.61 us|  6.37 us|   6.4x |
| `init_model_methods()` (already initialised)  | 54.14 us|  2.77 us|  19.5x |
| `fit$log_prob(u)`                             | 58.33 us|  5.70 us|  10.2x |
| `fit$grad_log_prob(u)`                        | 59.51 us|  5.94 us|  10.0x |
| raw `.Call` `grad_log_prob(ptr, u, TRUE)`     |  1.22 us|  1.23 us|     —  |

Wrapper-to-`.Call` ratio: 50.9x before, 5.0x after. The remainder is R call overhead, not guard work: an R6 method call is 0.51 us here and the wrapper makes two of them (the method itself, and `self$init_model_methods()`).

End to end, `bridgesampling::bridge_sampler()` on a `CmdStanMCMC` fit — conjugate normal model, so `logml` has a closed form:

| S     | before  | after   | factor |
|-------|--------:|--------:|-------:|
| 4000  | 0.492 s | 0.228 s |  2.16x |
| 32000 | 4.010 s | 1.709 s |  2.35x |

`logml` is bit-identical before and after (−57.886600 at S = 32000, against an analytic −57.886550). `loo(moment_match = TRUE)` pays the same constant through `$constrain_variables()`, once per draw per flagged observation.
## Where the time went

**`capture.output()`, and specifically the connection rather than the print.** `source_cpp_native_symbol_is_null()` decided whether a serialized `sourceCpp` wrapper had lost its native symbol by printing the symbol and grepping the captured text. `capture.output()` opens a text connection, sinks, prints, unsinks and reads back, per call:

| component                               | us/call |
|-----------------------------------------|--------:|
| `capture.output(invisible(NULL))`       |    24.7 |
| `capture.output(print(1L))`             |    28.9 |
| `capture.output(print(symbol))`         |    30.7 |
| `print(symbol)` into `sink(nullfile())` |    19.4 |

**`ls()`.** The initialised check was `!("model_ptr_" %in% ls(private$model_methods_env_))`. That environment holds ~16 bindings once the methods are exposed, and `ls()` lists and sorts all of them: 7.4 us, against 0.11 us for a direct lookup.

## The logic of the fast path

The guard exists for #1157: an object loaded from a previous session holds compiled bindings whose native symbol has degraded to a null pointer, and calling one of them aborts R. The question is how to recognise that state without paying for it on every call.

The fast path is `model_methods_are_live()` in `R/utils.R`, and the argument for
it is:

1. `model_ptr_` is an `Rcpp::XPtr` handed out by the compiled bindings — `initialize_model_pointer()` obtains it by calling into them.
2. R's serializer writes an external pointer's tag and protected field but not the address it contains; `unserialize()` reconstructs it with a null address. This is the same mechanism that creates the #1157 state in the first place.
3. Therefore a `model_ptr_` that still holds an address cannot have come through serialization, so it was produced by a successful call into bindings compiled in the running process, so those bindings' native symbols are valid.

That makes a non-null `model_ptr_` a sufficient condition for "nothing to do", decidable in constant time. It is also a slightly stronger check than the one it short-circuits: it tests the pointer that is about to be dereferenced rather than the wrapper's symbol.

Note this is not a memoised flag. Anything cached in the model-methods environment would be serialized along with it and come back asserting freshness; the fast path instead reads a value that serialization is guaranteed to destroy. That matters because the state is not session-scoped — the existing tests in `test-model-methods.R` and `test-fit-shared.R` both reload a serialized object inside a single session.

## Why `model_methods_are_live()` has three clauses

```r
typeof(model_ptr) == "externalptr" &&
  is.null(attributes(model_ptr)) &&
  !identical(model_ptr, .cmdstanr$NULL_EXTERNAL_POINTER)
```

`identical()` is the right primitive because its behaviour here is documented: per `?identical`, argument `extptr.as.ref`, by default "external pointers are considered identical if the addresses they contain are identical".

It also compares the objects' attributes, so a decorated pointer would not match a bare null one — and the failure would be in the dangerous direction: "not equal to null" reads as "live". The first two clauses exist so that only the exact shape this test can decide is ever reported as live, and anything else falls through to the full check instead. An earlier version of this patch omitted them, reported a reloaded fit as live, and reached `.Call(<pointer: (nil)>, ...)`.

## Why the null pointer is built with a `serialize()` round trip

```r
null_external_pointer <- function() {
  unserialize(serialize(methods::new("externalptr"), NULL))
}
```

`methods::new("externalptr")` does not allocate — it hands back the S4 prototype object for the class. External pointers are also not duplicated on assignment, so `structure(methods::new("externalptr"), class = "NativeSymbol")` does not build a classed copy: it attaches the class to the prototype itself, and every subsequent `new("externalptr")` in the session, in any package, comes back classed `"NativeSymbol"`. The round trip yields a private copy, and since serialization discards the address, the copy is null by construction. There is a test asserting cmdstanr leaves the prototype alone.

The result is cached in `.cmdstanr` because the round trip costs ~10 us and this comparison is on the per-call path; an external pointer's address never changes, so one instance serves the session.

`methods` moves to `Imports` for `methods::new()`.

## The symbol predicate

`source_cpp_native_symbol_is_null()` now compares the symbol against `unserialize(serialize(symbol, NULL))` — a null-valued twin of itself, carrying the same attributes by construction, so `identical()` reduces to the address comparison and is `TRUE` exactly when the symbol has already lost its address.

Besides being 7x cheaper, this drops a dependency on how a null pointer is formatted: the old regexp had to cover both `<pointer: 0x0>` and `<pointer: (nil)>`, which differ by platform.

This predicate is no longer on the per-call path — the fast path short-circuits before it — so its remaining cost is paid once per fit, in `CmdStanFit$initialize()`, and when a reloaded object rebuilds.

## Why `require_suggested_package("Rcpp")` moved

It is now inside the branch that compiles the methods. Rcpp is needed to build them; calling the already-compiled bindings only needs `.Call`. The error is still raised whenever it is relevant: a machine without Rcpp cannot have compiled bindings, and a fit carrying bindings from a machine that did has its stale ones dropped first, so `model_ptr` is `NULL` and that branch is taken.

## Tests

Existing #1157 coverage is unchanged and passing. Added:

- the reloaded-fit test now asserts the reloaded `model_ptr_` is present but not live, which is precisely the state the fast path must reject;
- `model_methods_are_live()` rejects a missing, an address-less, and a decorated pointer;
- `source_cpp_native_symbol_is_null()` flags a synthetic null `.Call` symbol and nothing else — no compilation needed;
- cmdstanr does not modify the shared `externalptr` prototype.

`devtools::test()` filtered to
`model-methods|fit-shared|utils|expose-functions|loo` passes, with one WSL-only
skip.

Verified outside the suite that a fit saved in one R process and reloaded in another rebuilds its model methods and returns the same `log_prob()`. Worth noting for reviewers that this cross-process case is what caught the bad version of the liveness test described above; the in-suite same-session reloads passed it.

## Not included

A batched entry point taking a matrix of unconstrained draws and looping in C++. That would amortise the R-side overhead that remains, which is what callers looping over draws actually want, but it is an API addition and belongs in its own issue.

## Copyright and Licensing

Aki Vehtari


jgabry:
Thanks Aki. Before reviewing I'll check how this intersects with the v1.0 branch I'm working on. I think it's mostly orthogonal so I can review it separately, but I want to double check in case it makes more sense to fold it into that work.

jgabry (commented):
This looks great. I'm going to make a few small edits and then merge if everything passes again.

jgabry (approved):

PR #1278, opened 2026-09-18
v1.0 stage 5: add `stan_build_info()` and tidy up the code around the build

jgabry:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR and the summary below were assisted by AI. I have reviewed all code and doc.

------

Part of #1258. This is Stage 5, the last of the feature stages: the public function that reads a build record back, plus three cleanup commits over the code the earlier stages wrote.

**`stan_build_info(exe_file)`** reads the record beside an executable and returns a `stan_build_info` object. Two fields are always there: `record`, with `status` and, when the record cannot be used, a `reason` (`missing`, `unreadable`, `executable_mismatch`, or `unsupported_format` with the `format_version` found); and `reported_features`, what the executable says about its own build, with `NA` kept apart from `FALSE` for a feature the executable did not report. With a usable record there are four more: `configuration` (the `cpp_options`, `stanc_options` and `include_paths` the build was asked for), `dependencies` (each file the build read, with the path it had then and whether it exists now), `cmdstan` (the installation's path and version, and whether the path exists), and `untracked_dependencies` (files the build depended on that cmdstanr cannot follow). A field that is absent means there was no record to read it from; an empty list is a recorded empty value. The result is narrower than the record on purpose, so the record's format can change underneath it. There is a print method, and `$build_info()` on a `CmdStanModel` calls the function on the model's executable. It reads the executable and record as they are on disk now, which is the point of it, so it does not go through the same check as the methods that run the executable.

The `?CmdStanModel` methods tables are regrouped around the program and the executable now that `$compile()` is gone, and gain `$variables()`, which was missing. The `cmdstan_model()` docs for `cpp_options` and `stanc_options` now say how an entry is written and point at `stan_build_info()` for an example.

**The second commit** goes over every internal roxygen block and comment the v1.0 stages wrote, so each has a title, its fields, and wording that stands without the design doc. A few names changed with it: the record checkers are `assert_record_*()`, `assess_build()` takes `expected` rather than `wanted`, `constructor_message()` is `build_message()`, and `test-model-guard.R` is `test-model-stale-executable.R`. `install_executable()` moved from `utils.R` to `build.R` beside its only caller, with its tests.

**The third commit** gives two groups of functions files of their own. Everything that talks to stanc, from `model.R` and `utils.R`, is now `R/stanc.R`; the Rcpp code behind `$init_model_methods()` and `$expose_functions()`, from `utils.R`, is now `R/expose.R`. Two exe-info helpers join their only caller in `build_record.R`, `tbb_path()` moves from `run.R` to `path.R`, and `cpp_opts.R` is renamed `cpp_options.R` after the argument it handles. Every function's body is unchanged; the one non-move edit is an error in `args.R` that still told the user to run `$compile()`. **The fourth commit** moves the unit tests of those functions into test-stanc.R and test-build-record.R to match.

Also removes `tests/testthat/testthat-problems.rds`, committed by accident in #1277, and ignores testthat's problem files from now on.

Not in this PR: the development-version NEWS entries that still describe `$compile()`, which the pre-RC NEWS pass rewrites.

Targeted test files run locally: test-build-info, test-build-record, test-build-assess, test-build-record-compile, test-build-install, test-model-stale-executable, test-model-rebuild-rules, test-model-compile, test-model-expose-functions, test-model-methods, test-cpp-options, test-utils, test-stanc, all passing.

#### Copyright and Licensing

Jonah Gabry


jgabry:
@andrjohns just wanted to give you a heads up that this also does some file reorganization that includes moving your code to expose Stan functions and model methods into a dedicated `expose.R`. I think that and some other file reorganization makes the code in the repo easier to find and follow.

